PR #81, opened 2015-03-19
Updates to Allow EF Design-Time Code to re-use Scaffolding

lajones:
@PradeepKadubandi 
Here's the PR we were talking about. I'll annotate with some comments.
Changes are all about providing net45 support.

### DIFF
--- a/src/Microsoft.Framework.CodeGeneration.Common/CommonUtilities.cs
+++ b/src/Microsoft.Framework.CodeGeneration.Common/CommonUtilities.cs
@@ -6,14 +6,12 @@ using System.Linq;
 using System.Reflection;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Emit;
-using Microsoft.Framework.Runtime;
 
 namespace Microsoft.Framework.CodeGeneration
 {
     internal static class CommonUtilities
     {
         public static CompilationResult GetAssemblyFromCompilation(
-            IAssemblyLoadContext loader,
             Compilation compilation)
         {
             EmitResult result;
@@ -45,12 +43,29 @@ namespace Microsoft.Framework.CodeGeneration
                     Assembly assembly;
                     if (PlatformHelper.IsMono)
                     {
-                        assembly = loader.LoadStream(ms, assemblySymbols: null);
+                        var assemblyLoadMethod = typeof(Assembly).GetTypeInfo().GetDeclaredMethods("Load")
+                            .First(
+                                m =>
+                                {
+                                    var parameters = m.GetParameters();
+                                    return parameters.Length == 1 && parameters[0].ParameterType == typeof(byte[]);
+                                });
+                        assembly = (Assembly)assemblyLoadMethod.Invoke(null, new[] { ms.ToArray() });
                     }
                     else
                     {
                         pdb.Seek(0, SeekOrigin.Begin);
-                        assembly = loader.LoadStream(ms, pdb);
+
+                        var assemblyLoadMethod = typeof(Assembly).GetTypeInfo().GetDeclaredMethods("Load")
+                            .First(
+                                m =>
+                                {
+                                    var parameters = m.GetParameters();
+                                    return parameters.Length == 2
+                                        && parameters[0].ParameterType == typeof(byte[])
+                                        && parameters[1].ParameterType == typeof(byte[]);
+                                });
+                        assembly = (Assembly)assemblyLoadMethod.Invoke(null, new[] { ms.ToArray(), pdb.ToArray() });
                     }
 
                     return CompilationResult.FromAssembly(assembly);
--- a/src/Microsoft.Framework.CodeGeneration.Common/TemplateFoldersUtilities.cs
+++ b/src/Microsoft.Framework.CodeGeneration.Common/TemplateFoldersUtilities.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+#if DNX451 || DNXCORE50
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -63,3 +64,4 @@ namespace Microsoft.Framework.CodeGeneration
         }
     }
 }
+#endif
--- a/src/Microsoft.Framework.CodeGeneration.EntityFramework/project.json
+++ b/src/Microsoft.Framework.CodeGeneration.EntityFramework/project.json
@@ -1,7 +1,7 @@
 {
     "version": "1.0.0-*",
     "dependencies": {
-        "EntityFramework": "7.0.0-*",
+        "EntityFramework.Core": "7.0.0-*",
         "Microsoft.Framework.CodeGeneration.Common": { "version": "1.0.0-*", "type": "build" },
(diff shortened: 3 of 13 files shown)


msotclas:
Hi __@lajones__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Lawrence Jones). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

src/Microsoft.Framework.CodeGeneration.Common/TemplateFoldersUtilities.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+#if DNX451 || DNXCORE50
```
lajones: This is used by Microsoft.Framework.CodeGeneration.EntityFramework and .Mvc. But not by .Templating. So updated so it only applies there.

src/Microsoft.Framework.CodeGeneration.EntityFramework/project.json
```diff
@@ -1,7 +1,7 @@
 {
     "version": "1.0.0-*",
     "dependencies": {
-        "EntityFramework": "7.0.0-*",
+        "EntityFramework.Core": "7.0.0-*",
```
lajones: Prevents circular dependency.

src/Microsoft.Framework.CodeGeneration.Templating/Compilation/MetadataReferencesProvider.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Reflection;
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.Framework.CodeGeneration.Templating.Compilation
+{
+    public class MetadataReferencesProvider
```
lajones: Originally I created this class from the dnx451/dnxcore50-dependent parts of RoslynCompilationService. The idea was that I would have `#if DNX451 || DNXCORE50` that would surround the ILibraryManager and IApplicationEnvironment parts and do something different for net45. But when I did the net45 stuff I found that I could apply that to dnx451 & dnxcore50 as well and have a class that did not need the `#if'. If you like I could subsume this class back into RoslynCompilationService. However I thought it might be useful if someone needed to override just this service to provide their own references.

davidfowl: I'm confused here. Did you just change how getting app references works on all platforms?

davidfowl:
I think you need to ifdef. This isn't something I would merge in.

src/Microsoft.Framework.CodeGeneration.Common/CommonUtilities.cs
```diff
@@ -6,14 +6,12 @@
...
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Emit;
-using Microsoft.Framework.Runtime;
 
 namespace Microsoft.Framework.CodeGeneration
 {
     internal static class CommonUtilities
     {
         public static CompilationResult GetAssemblyFromCompilation(
-            IAssemblyLoadContext loader,
```
davidfowl: Why don't you just implement one on net45 and pass it in instead of using reflection?

src/Microsoft.Framework.CodeGeneration.Templating/Compilation/MetadataReferencesProvider.cs
lajones: Yes. But those references are used solely for compiling the .cs file which the Razor Engine creates from the template. The other references weren't necessary. Have tried this out with all the tests and templates I could find and checked with @PradeepKadubandi and it was fine.

src/Microsoft.Framework.CodeGeneration.Common/CommonUtilities.cs
lajones: @bricelam  and I were discussing this and he had used this approach for his work on T4 templating.

lajones:
I started off with lots of ifdef's but when I came to try to load the references for net45 I found that I didn't need all the references that had been used - in fact I only needed the ones I've put in here. Then I found that they worked on dnx451 and dnxcore50 as well - so this approach was cleaner.

src/Microsoft.Framework.CodeGeneration.Templating/Compilation/MetadataReferencesProvider.cs
PradeepKadubandi: I am confused too... taking a closer look

src/Microsoft.Framework.CodeGeneration.Common/CommonUtilities.cs
PradeepKadubandi: +1 to Fowler's comment.

lajones: To what end? It seems much more complicated?

src/Microsoft.Framework.CodeGeneration.Templating/Compilation/RoslynCompilationService.cs
```diff
@@ -59,82 +36,5 @@ public CompilationResult Compile(string content)
...
-
-            // Todo: When the target app references scaffolders as nuget packages rather than project references,
-            // since the target app does not actully use code from scaffolding framework, our dlls
-            // are not in app dependencies closure. However we need them for compiling the
-            // generated razor template. So we are just using the known references for now to make things work.
-            // Note that this model breaks for custom scaffolders because they may be using
-            // some other references which are not available in any of these closures.
-            // As the above comment, the right thing to do here is to use the dependency closure of
-            // the assembly which has the template.
-            var baseProjects = new string[] { _environment.ApplicationName, "Microsoft.Framework.CodeGeneration" };
```
PradeepKadubandi: You seem to have removed this whole logic and I don't see an equivalent replacement somewhere (may be I am missing) ... this will break the scaffolding functionality... (it was on my to do list to add tests for templates but I was waiting till I finalize the content a little more)

lajones: Can you show me a template which would break? As we mentioned in our emails the ones in Mvc use a different mechanism and this does not break any of the existing templates.

The file that is produced just looks like this:

```public class __CompiledTemplate : RazorTemplateBase

    public void ExecuteAsync()
    {
        Write(xxxx);
        WriteLiteral(yyyy);
        ...
    }
```

The only thing that is referenced anywhere is Model which is a dynamic object defined on RazorTemplateBase.

As I say I found nothing that would depend on the project in which the template exists.

lajones:
Closing while I work with Pradeep on the issues. Will send out a new one once we've figured it out.

PR #93, opened 2015-04-07
Add serviceable attribute to projects.

NTaylorMullen:


NTaylorMullen:
Updated.

src/Microsoft.Framework.CodeGeneration.Common/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Reflection;
+
+[assembly: AssemblyMetadata("Serviceable", "True")]
```
Eilon: Looks like this is a shared source project.

Eilon:
@PradeepKadubandi in case you want to take a look.

src/Microsoft.Framework.CodeGeneration.Common/Properties/AssemblyInfo.cs
NTaylorMullen: Yup, updated script to account for *.cs shared projects.

Eilon:
:shipit:

PR #109, opened 2015-09-04
Pk/ef add

PradeepKadubandi:
This PR adds support for adding a DbSet<Model> property to DbContext when one does not exist.

To give some context, the above scenario can be achieved in 2 ways:
-  Use EF ModelBuilder API to dynamically add missing EntitySet to DbContext at design time, and if EF metadata can be fetched fine, then add it in code (this is the current approach WAP Mvc4 scaffolding uses but the current EF API did not yet offer this and after talking to EF folks I logged an issue in their repo)

- Other approach is to add the property first in code and use Roslyn to compile the entire app again in memory and reflect to get the EF metadata. (Older scaffolding used this approach at one point in time but was doing only partial compilation by taking a subset of project which was fragile , but given we are using the entire app this time, this should be more reliable but could be less performant).

To unblock the above scenario for usability study, I just coded it up 2nd approach and added a simple test for a simple scenario, I am planning to add more tests and fix the code as I find issues).

Once EF team fixes their issue we may revert to approach 1.

### DIFF
--- a/src/Microsoft.Framework.CodeGeneration.Core/ModelType.cs
+++ b/src/Microsoft.Framework.CodeGeneration.Core/ModelType.cs
@@ -13,6 +13,11 @@ namespace Microsoft.Framework.CodeGeneration
 
         public string FullName { get; set; }
 
+        // Violating the principle that ModelType should be decoupled from Roslyn's API.
+        // I had to do this for editing DbContext scenarios but I need to figure out if there
+        // is a better way.
+        public ITypeSymbol TypeSymbol { get; private set; }
+
         public static ModelType FromITypeSymbol([NotNull]ITypeSymbol typeSymbol)
         {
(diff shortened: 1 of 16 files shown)


src/Microsoft.Framework.CodeGeneration.Core/ModelType.cs
```diff
@@ -13,6 +13,11 @@ public class ModelType
 
         public string FullName { get; set; }
 
+        // Violating the principle that ModelType should be decoupled from Roslyn's API.
+        // I had to do this for editing DbContext scenarios but I need to figure out if there
+        // is a better way.
+        public ITypeSymbol TypeSymbol { get; private set; }
```
davidfowl: Digging the hole deeper :smile:

PR #116, opened 2015-10-14
Pk/web api

PradeepKadubandi:
Some more refactoring!


src/Microsoft.Extensions.CodeGenerators.Mvc/Controller/ControllerGen.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Extensions.CodeGeneration;
+using Microsoft.Extensions.CodeGeneration.CommandLine;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Extensions.CodeGenerators.Mvc.Controller
+{
+    [Alias("controller")]
+    public class ControllerGenerator : ICodeGenerator
```
abpiskunov: File name does not match the class name?

src/Microsoft.Extensions.CodeGenerators.Mvc/Controller/ControllerGeneratorModel.cs
```diff
@@ -17,7 +17,10 @@ public class ControllerGeneratorModel : CommonCommandLineModel
         [Option(Name = "controllerName", ShortName = "name", Description = "Name of the controller")]
         public string ControllerName { get; set; }
 
-        [Option(Name = "restWithNoViews", ShortName = "api", Description = "When specified, generates a Controller with REST style API, noViews is assumed and any view related options are ignored")]
-        public string IsRestController { get; set; }
+        [Option(Name = "restWithNoViews", ShortName = "api", Description = "Specify this switch to generate a Controller with REST style API, noViews is assumed and any view related options are ignored")]
```
abpiskunov: What is the story for localizing all console output messages?

abpiskunov:
:shipit:

src/Microsoft.Extensions.CodeGenerators.Mvc/Controller/ControllerGen.cs
PradeepKadubandi: :-), yeah, I wasn't very sure if git gets confused because another file with the name as this class was renamed in my local repo ... I corrected the file name in a later check in...

src/Microsoft.Extensions.CodeGenerators.Mvc/Controller/ControllerGeneratorModel.cs
PradeepKadubandi: Good question, when I was writing this first time in CTP3 or CTP4 time frame, I think there was no resource support in DNX and I never got to fix this part (I am not myself sure what's the support right now) , We probably need a bug to track this ...

PR #118, opened 2015-10-14
Support for specifying relative path at command line.

PradeepKadubandi:


src/Microsoft.Extensions.CodeGenerators.Mvc/View/ViewGenerator.cs
```diff
@@ -85,11 +85,11 @@ public virtual IEnumerable<string> TemplateFolders
...
-            var outputPath = Path.Combine(
-                appbasePath,
-                Constants.ViewsFolderName,
-                model.Name,
-                viewGeneratorModel.ViewName + Constants.ViewExtension);
+            var outputFolder = String.IsNullOrEmpty(viewGeneratorModel.RelativeFolderPath)
+                ? Path.Combine(appbasePath)
+                : Path.Combine(appbasePath, viewGeneratorModel.RelativeFolderPath);
+
+            var outputPath = Path.Combine(outputFolder, viewGeneratorModel.ViewName + Constants.ViewExtension);
```
abpiskunov: Didn't you forget Constants.ViewsFolderName here?

abpiskunov:
:watch:

abpiskunov:
Also - are there any tests for new change?

src/Microsoft.Extensions.CodeGenerators.Mvc/View/ViewGenerator.cs
PradeepKadubandi: Like I mentioned before, I intentionally removed the convention from command line ... for example, in VS, if user chose context menu on project and created a view using Add View dialog, the relative path would be empty and the expected behavior is to generate view in the project folder but not inside Views folder ... hence I needed to remove the convention from command line ... So if someone wants to generate a view inside 'Views' folder from command line, they need to pass '-outdir Views' explicitly...

PradeepKadubandi:
I missed adding tests specific to this change ... I was planning to add some end to end tests like I mentioned before either from VS or command line to test the full code generated in end to end scenarios ... not sure if I get a chance before this week though given the issues I have...

PR #121, opened 2015-10-15
Refactoring.

PradeepKadubandi:
More Refactoring.

### DIFF
--- a/src/Microsoft.Extensions.CodeGenerators.Mvc/View/ViewGenerator.cs
+++ b/src/Microsoft.Extensions.CodeGenerators.Mvc/View/ViewGenerator.cs
@@ -16,13 +16,17 @@ using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.Extensions.CodeGenerators.Mvc.View
 {
+    /// <summary>
+    /// Both the command line entry point for View generation (implements ICodeGenerator with Alias attribute)
+    /// and View generator itself (extends CommonGeneratorBase).
+    /// Consider separating if there are going to multiple view generators like Controllers.
+    /// </summary>
     [Alias("view")]
-    public class ViewGenerator : ICodeGenerator
+    public class ViewGenerator : CommonGeneratorBase, ICodeGenerator
     {
         private readonly IModelTypesLocator _modelTypesLocator;
         private readonly IEntityFrameworkService _entityFrameworkService;
         private readonly ILibraryManager _libraryManager;
-        private readonly IApplicationEnvironment _applicationEnvironment;
         private readonly ICodeGeneratorActionsService _codeGeneratorActionsService;
         private readonly IServiceProvider _serviceProvider;
         private readonly ILogger _logger;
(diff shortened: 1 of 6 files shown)


src/Microsoft.Extensions.CodeGenerators.Mvc/View/ViewGenerator.cs
```diff
@@ -62,77 +64,72 @@ public virtual IEnumerable<string> TemplateFolders
...
-            ModelType dataContext = ValidationUtil.ValidateType(viewGeneratorModel.DataContextClass, "dataContext", _modelTypesLocator, throwWhenNotFound: false);
-
-            // Validation successful
-            Contract.Assert(model != null, "Validation succeded but model type not set");
-
-            var appbasePath = _applicationEnvironment.ApplicationBasePath;
-            var outputFolder = String.IsNullOrEmpty(viewGeneratorModel.RelativeFolderPath)
-                ? Path.Combine(appbasePath)
-                : Path.Combine(appbasePath, viewGeneratorModel.RelativeFolderPath);
+            var outputPath = ValidateAndGetOutputPath(viewGeneratorModel, outputFileName: viewGeneratorModel.ViewName + Constants.ViewExtension);
```
abpiskunov: "Views" folder in the path is not added (in base class also)?

src/Microsoft.Extensions.CodeGenerators.Mvc/View/ViewGenerator.cs
```diff
@@ -62,77 +64,72 @@ public virtual IEnumerable<string> TemplateFolders
...
+                ViewDataTypeName = modelTypeAndContextModel.ModelType.FullName,
+                ViewDataTypeShortName = modelTypeAndContextModel.ModelType.Name,
                 ViewName = viewGeneratorModel.ViewName,
                 LayoutPageFile = viewGeneratorModel.LayoutPage,
                 IsLayoutPageSelected = isLayoutSelected,
                 IsPartialView = viewGeneratorModel.PartialView,
                 ReferenceScriptLibraries = viewGeneratorModel.ReferenceScriptLibraries,
-                ModelMetadata = modelMetadata,
+                ModelMetadata = modelTypeAndContextModel.ModelMetadata,
                 JQueryVersion = "1.10.2" //Todo
```
abpiskunov: how is this tracked?

abpiskunov:
:watch:

src/Microsoft.Extensions.CodeGenerators.Mvc/View/ViewGenerator.cs
PradeepKadubandi: The behavior in VS is that view is created in the folder where the context menu is invoked, so I removed any assumptions about conventions in command line code and only using RelativePath if provided as the single options for modifying the output path. VS will pass the right value for this parameter.

src/Microsoft.Extensions.CodeGenerators.Mvc/View/ViewGenerator.cs
PradeepKadubandi: You mean the JQueryVersion? Yeah, the scripts issues that I logged should be used to fix this...As of now, I think the parameter is not even used in templates ... we need to decide what scripts do we want to reference and what version and accordingly fix this...

PR #123, opened 2015-10-16
Web Api Controller With Context

PradeepKadubandi:
Web API Controller with Context functionality.


src/Microsoft.Extensions.CodeGenerators.Mvc/Controller/ControllerWithContextGenerator.cs
```diff
@@ -26,8 +40,87 @@ public abstract class ControllerWithContextGenerator : ControllerGeneratorBase
...
+            if (string.IsNullOrEmpty(controllerGeneratorModel.ControllerName))
+            {
+                //Todo: Pluralize model name
+                controllerGeneratorModel.ControllerName = modelTypeAndContextModel.ModelType.Name + Constants.ControllerSuffix;
+            }
+
+            var templateModel = new ControllerWithContextTemplateModel(modelTypeAndContextModel.ModelType, modelTypeAndContextModel.DbContextFullName)
+            {
+                ControllerName = controllerGeneratorModel.ControllerName,
+                AreaName = string.Empty, //ToDo
```
abpiskunov: What is Area?

abpiskunov: ah mvc area - never mind

src/Microsoft.Extensions.CodeGenerators.Mvc/Controller/ControllerWithContextGenerator.cs
```diff
@@ -26,8 +40,87 @@ public abstract class ControllerWithContextGenerator : ControllerGeneratorBase
...
+            string controllerRootName)
+        {
+            if (!controllerGeneratorModel.IsRestController && !controllerGeneratorModel.NoViews)
+            {
+                var layoutDependencyInstaller = ActivatorUtilities.CreateInstance<MvcLayoutDependencyInstaller>(ServiceProvider);
+                await layoutDependencyInstaller.Execute();
+
+                foreach (var viewTemplate in _views)
+                {
+                    var viewName = viewTemplate == "List" ? "Index" : viewTemplate;
```
abpiskunov: no one else I using _views, why we have List there? Why not have Index? Any sort of convention?

src/Microsoft.Extensions.CodeGenerators.Mvc/Controller/ControllerWithContextGenerator.cs
```diff
@@ -26,8 +40,87 @@ public abstract class ControllerWithContextGenerator : ControllerGeneratorBase
...
+
+                    // Todo: Need logic for areas
+                    var viewOutputPath = Path.Combine(
+                        ApplicationEnvironment.ApplicationBasePath,
+                        Constants.ViewsFolderName,
+                        controllerRootName,
+                        viewName + Constants.ViewExtension);
+
+                    await CodeGeneratorActionsService.AddFileFromTemplateAsync(viewOutputPath,
+                        viewTemplate + Constants.RazorTemplateExtension, TemplateFolders, viewTemplateModel);
```
abpiskunov: This actually generates a file for the view. So the flow is Controller generates views, but calls generation directly , not using ViewGenerator at all (which essentially does similar thing. I see that ViewGenerator does more and is called with Alias view (-view parameter?) , it just feels like naturally controller should use ViewGenerator and not know how views are generated .. ? Existing code also looks good, I am just trying to  see if I understand things right...

abpiskunov:
:shipit:

src/Microsoft.Extensions.CodeGenerators.Mvc/Controller/ControllerWithContextGenerator.cs
PradeepKadubandi: yep

src/Microsoft.Extensions.CodeGenerators.Mvc/Controller/ControllerWithContextGenerator.cs
PradeepKadubandi: The template name for 'Add View' dialog is 'List', I am not sure if there was some background, I kind of took the same list as VS 2013, perhaps the view name List makes more sense when generating individual view which shows all entities from database but when generating several views for controller we want the view name to be 'index' because that's the default.

If there is no push back , we could actually change the list of template names for view generator to have 'Index' as well instead of 'List'

src/Microsoft.Extensions.CodeGenerators.Mvc/Controller/ControllerWithContextGenerator.cs
PradeepKadubandi: yeah, you are right, in fact, Mvc5 Controller scaffoder creates an instance of ViewScaffolder just to do that ... And when I see it, there is a special method in ViewScaffolder specifically for creating views for a Controller ... 

I thought Controller generator decides some parameters for view generation and it's all controller generator specific ... so I chose this approach, all we are doing is invoking the template ... however, feel free to try your suggesting and change it if you like that more ...

PR #132, opened 2015-10-24
155994: Scaffolding should use short type names for known types.

PradeepKadubandi:
Used a string based building approach that could work for most known cases.
Added good number of test cases.

(Currently scaffolding is broken by breaking change from EF but I just took my new code and validated it in a console app)


abpiskunov:
:shipit:

PR #133, opened 2015-11-02
Fix for bug 155796 formatting of the Edit.cshtml is iincorrect

abpiskunov:
@BillHiebert


abpiskunov:
I optimistically merging - it is just a markup change. I tested it in IE and chrome and markup looks good (problem was that label did not show up at all and input size was very small in the Edit.cshtml)

PR #148, opened 2015-12-07
Fix CRUD templates in scaffolding to match the pattern in templates

prafullbhosale:
TFS bug 163912 
cc @abpiskunov


prafullbhosale:
Needs more updates.

PR #152, opened 2015-12-16
Handle reserved keywords while creating model variables in controller

prafullbhosale:
cc @abpiskunov @barrytang


src/Microsoft.Extensions.CodeGenerators.Mvc/Controller/ControllerWithContextTemplateModel.cs
```diff
@@ -73,8 +73,7 @@ public string ModelVariable
         {
```
abpiskunov: Tests are missing

abpiskunov:
:shipit , but add tests for this and next bug fixes and new features

prafullbhosale:
Added Unit tests.

PR #153, opened 2015-12-18
Moving all Exception strings to resource files.

prafullbhosale:
cc @abpiskunov


abpiskunov:
Did you sync up with someone on runtime side about how localized dlls would be generated out of resources you created?

abpiskunov:
:shipit:

prafullbhosale:
Plan to bring up during the next design meeting.

PR #157, opened 2015-12-31
DB context related changes

prafullbhosale:
Addressing below issues identified by @rowanmiller  

•	Null check after using Single is not needed as Single will throw if no entity is found. We probably want to swap to SingleOrDefault (rather than removing the null check) so that we return a Not Found status code.
•	We already discussed, but we should remove the EnsureCreated call to create the database. It’s not thread safe, and blocks you from using migrations in the future. We could look at scaffolding the corresponding migration for the user when we change the model for a given context.
•	Remove empty override of OnConfiguring, folks should use Startup.cs rather than this hook in ASP.NET 5 apps.
•	I would remove the empty override of OnModelCreating too – it doesn’t add much value.

cc @abpiskunov, @sayedihashimi, @barrytang

### DIFF
--- a/src/Microsoft.Extensions.CodeGeneration.EntityFramework/Templates/DbContext/NewLocalDbContext.cshtml
+++ b/src/Microsoft.Extensions.CodeGeneration.EntityFramework/Templates/DbContext/NewLocalDbContext.cshtml
@@ -23,23 +23,9 @@ using Microsoft.Data.Entity;
     } 
 }    public class @Model.DbContextTypeName : @baseClassName
     {
-        private static bool _created = false;
-
         public @(Model.DbContextTypeName)()
         {
-            if (!_created)
-            {
-                _created = true;
-                Database.EnsureCreated();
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.CodeGeneration.EntityFramework/Templates/DbContext/NewLocalDbContext.cshtml
```diff
@@ -23,23 +23,9 @@ using Microsoft.Data.Entity;
...
-            }
-        }
-
-        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
-        {
-        }
-
-        protected override void OnModelCreating(ModelBuilder builder)
-        {
+
```
rowanmiller: No real need to define this constructor either since it's the same as what you get by default.

rowanmiller:
:shipit: from my perspective

PR #178, opened 2016-02-27
Fix #177

### LINKED ISSUE #177: If the user project is not built yet, trying to scaffold throws an Exception that the project's dll was not found
This is happening due to the use of project's exports for razor template compilation. 
The intent is to get the Microsoft.Extensions.Codegeneration exports (or any custom scaffolder's exports) for the compilation. So we do not really need the types in the project itself.

prafullbhosale:
Try to get the metadata references and fail only if the compilation fails with the available references.
+ Minor cleanup and formatting

cc @abpiskunov

### DIFF
--- a/src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExportExtensions.cs
+++ b/src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExportExtensions.cs
@@ -30,7 +30,16 @@ namespace Microsoft.Extensions.CodeGeneration.DotNet
                     }
                 }
                 catch (Exception ex) 
+                    when (ex is FileNotFoundException 
+                            || ex is DirectoryNotFoundException
+                            || ex is NotSupportedException
+                            || ex is ArgumentException
+                            || ex is ArgumentOutOfRangeException
+                            || ex is BadImageFormatException
+                            || ex is IOException
+                            || ex is ArgumentNullException)
                 {
+                    // TODO: Log this
                     if(throwOnError) 
                     {
                         throw ex;
--- a/src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExportExtensions.cs
+++ b/src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExportExtensions.cs
@@ -14,17 +14,28 @@ namespace Microsoft.Extensions.CodeGeneration.DotNet
 {
     public static class LibraryExportExtensions
     {
-        public static IEnumerable<MetadataReference> GetMetadataReferences(this LibraryExport export)
+        public static IEnumerable<MetadataReference> GetMetadataReferences(this LibraryExport export, bool throwOnError = true)
         {
             var references = new List<MetadataReference>();
             AssemblyMetadata assemblyMetadata;
             foreach (var lib in export.CompilationAssemblies)
             {
-                using (var stream = File.OpenRead(lib.ResolvedPath))
+                try
                 {
-                    var moduleMetadata = ModuleMetadata.CreateFromStream(stream, PEStreamOptions.PrefetchMetadata);
-                    assemblyMetadata = AssemblyMetadata.Create(moduleMetadata);
-                    references.Add(assemblyMetadata.GetReference());
+                    using (var stream = File.OpenRead(lib.ResolvedPath))
+                    {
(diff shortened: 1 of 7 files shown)


prafullbhosale:
Ping.

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExportExtensions.cs
```diff
@@ -14,17 +14,28 @@ namespace Microsoft.Extensions.CodeGeneration.DotNet
 {
     public static class LibraryExportExtensions
     {
-        public static IEnumerable<MetadataReference> GetMetadataReferences(this LibraryExport export)
+        public static IEnumerable<MetadataReference> GetMetadataReferences(this LibraryExport export, bool throwOnError = true)
```
abpiskunov: Unit tests?

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExportExtensions.cs
```diff
@@ -14,17 +14,28 @@ namespace Microsoft.Extensions.CodeGeneration.DotNet
...
+                        references.Add(assemblyMetadata.GetReference());
+                    }
+                }
+                catch (Exception ex) 
+                {
+                    if(throwOnError) 
+                    {
+                        throw ex;
+                    }
+                    continue;
```
abpiskunov: What would happen if we continue? we would have missing reference and compilation would fail? How it would look for user? 
Also what kind of exceptions could be here? Can we look for specific ones we expect and let others go through? It is not good in general to catch all exceptions.

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExportExtensions.cs
```diff
@@ -14,17 +14,28 @@ namespace Microsoft.Extensions.CodeGeneration.DotNet
...
+                    using (var stream = File.OpenRead(lib.ResolvedPath))
+                    {
+                        var moduleMetadata = ModuleMetadata.CreateFromStream(stream, PEStreamOptions.PrefetchMetadata);
+                        assemblyMetadata = AssemblyMetadata.Create(moduleMetadata);
+                        references.Add(assemblyMetadata.GetReference());
+                    }
+                }
+                catch (Exception ex) 
+                {
+                    if(throwOnError)
```
abpiskunov: Does any code use throwOnError=true?

abpiskunov:
:watch:

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExportExtensions.cs
prafullbhosale: Nothing right now.

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExportExtensions.cs
prafullbhosale: Right now, the whole error message experience is pretty crude for the user. 
If compilation fails, it would show the compilation errors to the user. 
Types of exception expected here: 
FileNotFoundException, NotSupportedException, ArgumentException, ArgumentOutOfRangeException, BadImageFormatException, IOException, ArgumentNullException. 

Will change to catch these specific exceptions. 

Also I think currently the debug logging is not great. I will open a new issue for debug logging this and other exceptions/ errors.

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExportExtensions.cs
prafullbhosale: Added.

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExportExtensions.cs
abpiskunov: hm did you just list all exceptions from CreateFromStream and Create? I meant if there are some specific scenarios which cause a specific exception  that you aware of and expect then we should handle that . However if you want to protect from any exceptions to avoid throwing to user , then generic Exception should work as soon as we report it somehow differently.

abpiskunov:
:shipit:

prafullbhosale:
Merge blocked by dotnet/cli#1755

PR #179, opened 2016-03-01
Fix build and tests on *nix.

cesarbs:
cc @pranavkm

### DIFF
--- a/src/Microsoft.Extensions.CodeGeneration.Utils/CommonUtilities.cs
+++ b/src/Microsoft.Extensions.CodeGeneration.Utils/CommonUtilities.cs
@@ -20,55 +20,36 @@ namespace Microsoft.Extensions.CodeGeneration
             EmitResult result;
             using (var ms = new MemoryStream())
             {
-                using (var pdb = new MemoryStream())
+                result = compilation.Emit(ms, pdbStream: null);
+
+                if (!result.Success)
                 {
-                    if (PlatformHelper.IsMono)
-                    {
-                        result = compilation.Emit(ms, pdbStream: null);
-                    }
-                    else
-                    {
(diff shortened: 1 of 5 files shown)


src/Microsoft.Extensions.CodeGeneration.Utils/CommonUtilities.cs
```diff
@@ -22,7 +22,7 @@ public static class CommonUtilities
             {
                 using (var pdb = new MemoryStream())
                 {
-                    if (PlatformHelper.IsMono)
+                    if (PlatformHelper.IsMono || !PlatformHelper.IsWindows)
```
pranavkm: We probably don't care about the pdbs for scaffolded items in general. I think it should be ok to remove the else entirely.

pranavkm:
:shipit:

src/Microsoft.Extensions.CodeGeneration.Utils/CommonUtilities.cs
cesarbs: Wouldn't it be better to run this by other folks first?

pranavkm: Sure. cc @prafullbhosale, thoughts on this?

prafullbhosale: Yeah, we shouldn't care about the pdbs. These assemblies are generated either for retrieving EF related information or to compile the razor templates. 
Should be safe to always pass pdbstream as null

cesarbs:
Ping. Please take another look now that I've removed the pdb stream stuff.

pranavkm:
Nice. :shipit:

prafullbhosale:
:shipit:

PR #180, opened 2016-03-03
TypeActivate CodeGenCommand instead of new()

HaoK:
Removes GetRequiredService

Part of https://github.com/aspnet/Coherence-Signed/issues/186

@lodejard

### DIFF
--- a/src/Microsoft.Extensions.CodeGeneration/CodeGenCommand.cs
+++ b/src/Microsoft.Extensions.CodeGeneration/CodeGenCommand.cs
@@ -13,14 +13,16 @@ namespace Microsoft.Extensions.CodeGeneration
         private readonly ILogger _logger;
         private readonly ICodeGeneratorLocator _locator;
 
-        public IServiceProvider ServiceProvider { get; set; }
-        public CodeGenCommand(IServiceProvider serviceProvider, ILogger logger, ICodeGeneratorLocator locator)
+        public CodeGenCommand(ILogger logger, ICodeGeneratorLocator locator)
         {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.CodeGeneration/CodeGenCommand.cs
```diff
@@ -5,42 +5,42 @@
...
-using System.Threading.Tasks;
-using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.Extensions.CodeGeneration
 {
     public class CodeGenCommand
     {
+        private readonly ILogger _logger;
+        private readonly ICodeGeneratorLocator _locator;
+
```
HaoK: We probably could get rid of the IServiceProvider dependency now, but not sure if that's needed for consumer usage outside of this class

prafullbhosale: Its not needed for consumer usage. 
Should be safe to remove. The ServiceProvider didn't need to be public in the first place.

HaoK: Will do, does the change look good to you other than that?

prafullbhosale: :+1:

prafullbhosale:
:shipit:

HaoK:
a2eaf1e5ec1c3d76fd0e15029535dcd28e74b18f

PR #182, opened 2016-03-17
Temporary solution for DisplayPropertyName.

AlirezaP:
scaffolder has problem in controller generate.
in the other words "DisplayPropertyName" property in the "NavigationMetadata" class is not correct.() --> //Needs further implementation
so we can solve this with reflection Temporary.
we can change "@property.DisplayPropertyName" to "typeof(@property.EntitySetName).GetProperties().Where(p=>p.Name!= "@property.PrimaryKeyNames[0]").FirstOrDefault().Name" in the "MvcControllerWithContext" (templates) or set it for "DisplayPropertyName" in the "NavigationMetadata" class.

### DIFF
--- a/src/Microsoft.Extensions.CodeGenerators.Mvc/Templates/ControllerGenerator/MvcControllerWithContext.cshtml
+++ b/src/Microsoft.Extensions.CodeGenerators.Mvc/Templates/ControllerGenerator/MvcControllerWithContext.cshtml
@@ -132,7 +132,7 @@ namespace @Model.ControllerNamespace
 @{
     foreach (var property in relatedProperties.Values)
     {
-            @:ViewData["@(property.ForeignKeyPropertyNames[0])"] = new SelectList(_context.@property.EntitySetName, "@property.PrimaryKeyNames[0]", "@property.DisplayPropertyName");
+            @:ViewData["@(property.ForeignKeyPropertyNames[0])"] = new SelectList(_context.@property.EntitySetName, "@property.PrimaryKeyNames[0]", typeof(@property.EntitySetName).GetProperties().Where(p=>p.Name!= "@property.PrimaryKeyNames[0]").FirstOrDefault().Name);
     }
 }            return View();
         }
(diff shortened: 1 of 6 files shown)


src/Microsoft.Extensions.CodeGenerators.Mvc/Templates/ControllerGenerator/MvcControllerWithContext.cshtml
```diff
@@ -171,7 +171,7 @@ namespace @Model.ControllerNamespace
 @{
     foreach (var property in relatedProperties.Values)
     {
-            @:ViewData["@(property.ForeignKeyPropertyNames[0])"] = new SelectList(_context.@property.EntitySetName, "@property.PrimaryKeyNames[0]", "@property.DisplayPropertyName", @(Model.ModelVariable).@property.ForeignKeyPropertyNames[0]);
+            @:ViewData["@(property.ForeignKeyPropertyNames[0])"] = new SelectList(_context.@property.EntitySetName, "@property.PrimaryKeyNames[0]", typeof(@property.EntitySetName).GetProperties().Where(p=>p.Name!= "@property.PrimaryKeyNames[0]").FirstOrDefault().Name, @(Model.ModelVariable).@property.ForeignKeyPropertyNames[0]);
```
prafullbhosale: The problem with this is that it chooses the first property from the navigation that is not the primary key. 

Consider the model: 

```
public class Car 
{
        public int CarId { get; set; }
        public int ManufacturerId { get; set; }
        public virtual Manufacturer Manufacturer { get; set; }
}

public class Manufacturer
{
        public int ManufacturerId { get; set; }
        public bool IsCertified { get; set; }
        public virtual ICollection<Car> Cars { get; set; }
}

```
In this model if the Display property ends up being `IsCertified`, the drop down would consist of `true` and `false` options for each of the available `Manufacturer` entries.

barrytang:
TFS Bug#<a href="https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workItems#id=232587&fullScreen=false&_a=edit">232587</a>

prafullbhosale:
Closing this.

PR #184, opened 2016-03-27
Use `IFormCollection` not `.Internal` implementation

dougbu:
/cc @davidfowl @prafullbhosale


dougbu:
Found while investigating aspnet/HttpAbstractions#549 but not exactly part of that issue. Regardless of the namespace, `FormCollection` should rarely appear in product or user code.

davidfowl:
:shipit:

dougbu:
99fad1e

PR #188, opened 2016-04-01
Fix sources tests

JunTaoLuo:
This will resolve https://github.com/aspnet/Coherence-Signed/issues/230

### DIFF
--- a/test/Microsoft.Extensions.CodeGeneration.Sources.Test/LibraryExportExtensionsTest.cs
+++ b/test/Microsoft.Extensions.CodeGeneration.Sources.Test/LibraryExportExtensionsTest.cs
@@ -2,13 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.DotNet.ProjectModel.Compilation;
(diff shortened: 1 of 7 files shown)


test/Microsoft.Extensions.CodeGeneration.Sources.Test/LibraryExportExtensionsTest.cs
```diff
@@ -29,13 +27,12 @@ public void LibraryExportExtensions_GetMetadataReferences()
...
             Assert.True(refList.Any());
         }
-        
+
         [Fact]
         public void LibraryExportExtensions_GetMetadataReferences_throwException()
         {
-            //arrange
             var libraryExporter = GetInvalidLibraryExporter();
-            _export = libraryExporter.GetAllExports().Last();
```
JunTaoLuo: Maybe this used to work when the order of the library exports were reversed. Currently, it will get `System.Xml.XDocument` or some other library which may be valid. Since it can get an arbitrary export, the test can break when it expect the library to be non-existent.

JunTaoLuo:
cc @prafullbhosale @cesarbs

JunTaoLuo:
@Tratcher @BrennanConroy

BrennanConroy:
:shipit: assuming it passes

JunTaoLuo:
This will resolve some problems, There's another one that popped up on *nix systems which I will investigate separately.

PR #195, opened 2016-04-13
Address output review feedback

prafullbhosale:
Feedback based on aspnet/Tooling/issues/444
cc @sayedihashimi, @DamianEdwards, @Eilon  @barrytang


prafullbhosale:
Tagging aspnet/Release#4

src/Microsoft.Extensions.CodeGenerators.Mvc/Templates/ControllerGenerator/MvcControllerWithContext.cshtml
```diff
@@ -216,31 +226,52 @@ namespace @Model.ControllerNamespace
...
-        @:public async Task<IActionResult> Edit(@Model.ModelTypeName @Model.ModelVariable)
+        @:public async Task<IActionResult> Edit(@primaryKeyShortTypeName id, [Bind("@bindString")] @Model.ModelTypeName @Model.ModelVariable)
     }
     else
     {
-        @:public IActionResult Edit(@Model.ModelTypeName @Model.ModelVariable)
+        @:public IActionResult Edit(@primaryKeyShortTypeName id, [Bind("@bindString")] @Model.ModelTypeName @Model.ModelVariable)
     }
 }        {
+            if(id != @Model.ModelVariable.@primaryKeyName)
```
Eilon: Add space after `if`

src/Microsoft.Extensions.CodeGenerators.Mvc/Templates/ControllerGenerator/MvcControllerWithContext.cshtml
```diff
@@ -216,31 +226,52 @@ namespace @Model.ControllerNamespace
...
+            {
+                return NotFound();
+            }
             if (ModelState.IsValid)
```
Eilon: Insert empty line before this `if`

Eilon:
Overall looks good.

As a side question: could we perhaps get rid of the `use async` option and just always do async controller actions? @DamianEdwards @danroth27 @sayedihashimi @barrytang any thoughts on that?

danroth27:
Well, I wouldn't want to *always* do async controller actions. I don't think `Index`, `About` and `Contact` in our templates and item templates need to be async. But for actions that we know are accessing a database I think opting into async actions would be ok as long as the behavior is predictable and not to magical.

Eilon:
Talked to @danroth27 in person so I will clarify: yeah I just mean that for our database-oriented controller actions we would always do async. And we could thus remove the checkbox from any UI that we have.

prafullbhosale:
@Eilon That sounds good to me. 
I will make the changes for making the database oriented controller actions async only. 
And will remove the checkbox from the UI. 

The other controllers will remain as they are.

prafullbhosale:
:up::date:

cc @Rick-Anderson

prafullbhosale:
cc @phenning

phenning:
:shipit:

PR #197, opened 2016-04-18
Pb/net45support

### LINKED ISSUE #196: Scaffolding a controller using entity framework fails if the user project has services.AddMvc
If the user project has 
`services.AddMvc();`
or 
```
app.UseMvc( ...
);
```

scaffolding an ef based controller fails. 

The reason for the failure is that these methods try to load the project's assembly to find out the parts of the assembly to use for MVC. 

Scaffolding code loads the project in memory and then modifies the startup.cs and creates a new in-memory assembly with a different name. 
So when the code tries to invoke the startup type, the mvc code tries to load the 'newName' assembly, which cannot be found.

prafullbhosale:
This change is to support projects targeting only full .net (eg. net451)
Changes include:

1. Build all projects for net451
2. Refactor dotnet-aspnet-codegenerator tool to work as a portable tool as well as a ProjectDependencyCommand (based on dotnet/cli/issues/2020)
3. Changed behavior of AssemblyLoadContext. (For full .net since the command is invoked from build folder, AssemblyLoadContext behavior is same as Assembly.Load())
4. Fixes #196 

cc @barrytang  @sayedihashimi @abpiskunov @NTaylorMullen 

Currently blocked on dotnet/cli/issues/2505 for supporting projects that target only full .net

Tagging aspnet/Release/issues/91, aspnet/Release/issues/4

### DIFF
--- a/src/Microsoft.Extensions.CodeGeneration.Core/DefaultCodeGeneratorAssemblyProvider.cs
+++ b/src/Microsoft.Extensions.CodeGeneration.Core/DefaultCodeGeneratorAssemblyProvider.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Reflection;
 using Microsoft.DotNet.ProjectModel;
 using Microsoft.Extensions.CodeGeneration.DotNet;
+using System.Diagnostics;
 
 namespace Microsoft.Extensions.CodeGeneration
 {
@@ -16,6 +17,12 @@ namespace Microsoft.Extensions.CodeGeneration
             new HashSet<string>(StringComparer.Ordinal)
             {
                 "Microsoft.Extensions.CodeGeneration",
+            };
+        private static readonly HashSet<string> _exclusions =
+            new HashSet<string>(StringComparer.Ordinal)
+            {
+                "dotnet-aspnet-codegenerator",
+                "Microsoft.Extensions.CodeGeneration"
             };
 
         private readonly ILibraryManager _libraryManager;
@@ -51,14 +58,15 @@ namespace Microsoft.Extensions.CodeGeneration
                     .SelectMany(_libraryManager.GetReferencingLibraries)
                     .Distinct()
                     .Where(IsCandidateLibrary);
-                
-                return list.Select(lib => _assemblyLoadContext.LoadFromPath(_libraryExporter.GetResolvedPathForDependency(lib)));
+                return list.Select(lib => _assemblyLoadContext.LoadFromName(new AssemblyName(lib.Identity.Name)));
             }
         }
 
         private bool IsCandidateLibrary(LibraryDescription library)
         {
-            return !_codeGenerationFrameworkAssemblies.Contains(library.Identity.Name) && (!"Project".Equals(library.Identity.Type.Value, StringComparison.OrdinalIgnoreCase));
+            return !_exclusions.Contains(library.Identity.Name)
+                && (!"Project".Equals(library.Identity.Type.Value, StringComparison.OrdinalIgnoreCase));
+
         }
     }
 }
--- a/src/Microsoft.Extensions.CodeGeneration.Core/DefaultCodeGeneratorAssemblyProvider.cs
+++ b/src/Microsoft.Extensions.CodeGeneration.Core/DefaultCodeGeneratorAssemblyProvider.cs
@@ -3,12 +3,9 @@
 
 using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
 using System.Reflection;
-using System.Runtime.Loader;
 using Microsoft.DotNet.ProjectModel;
-using Microsoft.DotNet.ProjectModel.Graph;
 using Microsoft.Extensions.CodeGeneration.DotNet;
 
 namespace Microsoft.Extensions.CodeGeneration
--- a/src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
+++ b/src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
@@ -244,11 +244,18 @@ namespace Microsoft.Extensions.CodeGeneration.EntityFrameworkCore
             out Type modelType)
         {
             var projectCompilation = _workspace.CurrentSolution.Projects
-                .FirstOrDefault(project => project.AssemblyName == _environment.ApplicationName)
+                .First(project => project.AssemblyName == _applicationInfo.ApplicationName)
                 .GetCompilationAsync().Result;
+            // Need these #ifdefs as coreclr needs the assembly name to be different to be loaded from stream. 
+            // On NET451 if the assembly name is different, MVC fails to load the assembly as it is not found on disk. 
+#if NET451
             var newAssemblyName = projectCompilation.AssemblyName;
-
-            var newCompilation = compilationModificationFunc(projectCompilation).WithAssemblyName(newAssemblyName);
+#else
+            var newAssemblyName = Path.GetRandomFileName();
+#endif
+            var newCompilation = compilationModificationFunc(projectCompilation)
+                .WithAssemblyName(newAssemblyName)
+                .WithOptions(new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary));
 
             var result = CommonUtilities.GetAssemblyFromCompilation(_loader, newCompilation);
 
@@ -361,7 +368,7 @@ namespace Microsoft.Extensions.CodeGeneration.EntityFrameworkCore
                 {
                     ex = ex.InnerException;
                 }
-                _logger.LogMessage("There was an error creating the DbContext instance to get the model." + ex.Message, LogMessageLevel.Error);
+                _logger.LogMessage($"There was an error creating the DbContext instance to get the model. {ex.Message}", LogMessageLevel.Error);
                 throw ex;
             }
 
@@ -397,8 +404,7 @@ namespace Microsoft.Extensions.CodeGeneration.EntityFrameworkCore
             try {
                 var builder = new WebHostBuilder();
                 builder.UseKestrel()
-                        .UseContentRoot(Directory.GetCurrentDirectory())
-                        .UseIISIntegration();
+                        .UseContentRoot(Directory.GetCurrentDirectory());
 
                 if (startupType != null)
                 {
--- a/src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
+++ b/src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
@@ -31,7 +31,6 @@ namespace Microsoft.Extensions.CodeGeneration.EntityFrameworkCore
         private readonly IPackageInstaller _packageInstaller;
         private readonly IServiceProvider _serviceProvider;
         private readonly ILogger _logger;
-        private static int _counter = 1;
         private const string EFSqlServerPackageName = "Microsoft.EntityFrameworkCore.SqlServer";
         private const string EFSqlServerPackageVersion = "7.0.0-*";
         private readonly Workspace _workspace;
@@ -244,12 +243,10 @@ namespace Microsoft.Extensions.CodeGeneration.EntityFrameworkCore
             out Type dbContextType, 
             out Type modelType)
         {
-            //TODO: @prbhosal Figure out how to lookup the correct project here.             
             var projectCompilation = _workspace.CurrentSolution.Projects
-                //.Where(project => project.Name == _applicationInfo.ApplicationName)
-                .FirstOrDefault()
+                .FirstOrDefault(project => project.AssemblyName == _environment.ApplicationName)
                 .GetCompilationAsync().Result;
-            var newAssemblyName = projectCompilation.AssemblyName + _counter++;
+            var newAssemblyName = projectCompilation.AssemblyName;
 
             var newCompilation = compilationModificationFunc(projectCompilation).WithAssemblyName(newAssemblyName);
 
@@ -399,10 +396,10 @@ namespace Microsoft.Extensions.CodeGeneration.EntityFrameworkCore
         {
             try {
                 var builder = new WebHostBuilder();
-                //TODO: Review 
-                builder.UseServer("Microsoft.AspNetCore.Server.WebListener")
-                        .UseContentRoot(Directory.GetCurrentDirectory());
-                        
+                builder.UseKestrel()
+                        .UseContentRoot(Directory.GetCurrentDirectory())
+                        .UseIISIntegration();
+
                 if (startupType != null)
                 {
                     var reflectedStartupType = dbContextType.GetTypeInfo().Assembly.GetType(startupType.FullName);
@@ -416,7 +413,11 @@ namespace Microsoft.Extensions.CodeGeneration.EntityFrameworkCore
             }
             catch(Exception ex)
             {
-                _logger.LogMessage(ex.Message);
+                while (ex != null)
+                {
+                    _logger.LogMessage($"{ex.Message} StackTrace:{Environment.NewLine}{ex.StackTrace}{Environment.NewLine}");
+                    ex = ex.InnerException;
+                }
                 return null;
             }
         }
--- a/src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
+++ b/src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
@@ -8,7 +8,6 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Reflection;
-using System.Runtime.Loader;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Hosting;
--- a/src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/project.json
+++ b/src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/project.json
@@ -6,7 +6,6 @@
   },
   "dependencies": {
     "Microsoft.AspNetCore.Hosting": "1.0.0-*",
-    "Microsoft.AspNetCore.Server.IISIntegration": "1.0.0-*",
     "Microsoft.AspNetCore.Server.Kestrel": "1.0.0-*",
     "Microsoft.EntityFrameworkCore": "1.0.0-*",
     "Microsoft.EntityFrameworkCore.SqlServer": "1.0.0-*",
--- a/src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/project.json
+++ b/src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/project.json
@@ -6,17 +6,14 @@
   },
   "dependencies": {
     "Microsoft.AspNetCore.Hosting": "1.0.0-*",
-    "Microsoft.AspNetCore.Server.WebListener": "0.1.0-*",
+    "Microsoft.AspNetCore.Server.IISIntegration": "1.0.0-*",
+    "Microsoft.AspNetCore.Server.Kestrel": "1.0.0-*",
     "Microsoft.EntityFrameworkCore": "1.0.0-*",
     "Microsoft.EntityFrameworkCore.SqlServer": "1.0.0-*",
     "Microsoft.Extensions.CodeGeneration.Core": "1.0.0-*"
   },
   "frameworks": {
-    "net451": {
-        "dependencies": {
-            "System.Threading.Tasks": "4.0.0-*"
-        }
-    },
+    "net451": { },
     "netstandard1.5": {
       "imports": [
         "dnxcore50",
--- a/src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/project.json
+++ b/src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/project.json
@@ -9,17 +9,22 @@
     "Microsoft.AspNetCore.Server.WebListener": "0.1.0-*",
     "Microsoft.EntityFrameworkCore": "1.0.0-*",
     "Microsoft.EntityFrameworkCore.SqlServer": "1.0.0-*",
-    "Microsoft.Extensions.CodeGeneration.Core": "1.0.0-*",
-    "Microsoft.Extensions.CodeGeneration.Templating": "1.0.0-*",
-    "Microsoft.Extensions.CodeGeneration.Utils": "1.0.0-*",
-    "System.AppContext": "4.1.0-*"
+    "Microsoft.Extensions.CodeGeneration.Core": "1.0.0-*"
   },
   "frameworks": {
+    "net451": {
+        "dependencies": {
+            "System.Threading.Tasks": "4.0.0-*"
+        }
+    },
     "netstandard1.5": {
       "imports": [
         "dnxcore50",
         "portable-net45+win8"
-      ]
+      ], 
+      "dependencies": {
+          "System.AppContext": "4.1.0-*"
+      }
     }
   },
   "packInclude": {
--- a/src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/DefaultAssemblyLoadContext.cs
+++ b/src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/DefaultAssemblyLoadContext.cs
@@ -16,241 +16,27 @@ namespace Microsoft.Extensions.CodeGeneration.DotNet
 {
     public partial class DefaultAssemblyLoadContext : ICodeGenAssemblyLoadContext
     {
-        private readonly IDictionary<AssemblyName, string> _assemblyPaths;
-        private readonly IDictionary<string, string> _nativeLibraries;
-        private readonly IEnumerable<string> _searchPaths;
-
-        private static readonly string[] NativeLibraryExtensions;
-        private static readonly string[] ManagedAssemblyExtensions = new[]
-        {
-            ".dll",
-            ".ni.dll",
-            ".exe",
-            ".ni.exe"
-        };
-        
-        public DefaultAssemblyLoadContext(IDictionary<AssemblyName, string> assemblyPaths,
-                                   IDictionary<string, string> nativeLibraries,
-                                   IEnumerable<string> searchPaths)
-        {
-            _assemblyPaths = assemblyPaths ?? new Dictionary<AssemblyName, string>();
-            _nativeLibraries = nativeLibraries ?? new Dictionary<string, string>();
-            _searchPaths = searchPaths ?? new List<string>();
-        }
-        
-        private bool SearchForLibrary(string[] extensions, string name, out string path)
-        {
-            foreach (var searchPath in _searchPaths)
-            {
-                foreach (var extension in extensions)
-                {
-                    var candidate = Path.Combine(searchPath, name + extension);
-                    if (File.Exists(candidate))
-                    {
-                        path = candidate;
-                        return true;
-                    }
-                }
-            }
-            path = null;
-            return false;
-        }
-        
-        public static ICodeGenAssemblyLoadContext CreateAssemblyLoadContext(string nugetPackageDir)
-        {
-            List<string> searchPaths = new List<string>();
-            if (Directory.Exists(nugetPackageDir))
-            {
-                Queue<string> queue = new Queue<string>();
-                queue.Enqueue(nugetPackageDir);
-                while (queue.Count > 0)
-                {
-                    var current = queue.Dequeue();
-                    searchPaths.Add(current);
-                    try
-                    {
-                        var subdirs = Directory.EnumerateDirectories(current);
-                        if (subdirs != null)
-                        {
-                            foreach (var sd in subdirs)
-                            {
-                                queue.Enqueue(sd);
-                            }
-                        }
-                    }
-                    catch
-                    {
-                        // Do not want to fail if we cannot access certain directories.
-                        continue;
-                    }
-                }
-            }
-
-            return new DefaultAssemblyLoadContext(null, null, searchPaths);
-        }
-    }
-    
-#if !NET451 
-    public partial class DefaultAssemblyLoadContext : AssemblyLoadContext
-    {
-        static DefaultAssemblyLoadContext()
-        {
-            {
-                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-                {
-                    NativeLibraryExtensions = new[] { ".dll" };
-                }
-                else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
-                {
-                    NativeLibraryExtensions = new[] { ".dylib" };
-                }
-                else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
-                {
-                    NativeLibraryExtensions = new[] { ".so" };
-                }
-                else
-                {
-                    NativeLibraryExtensions = new string[0];
-                }
-            }
-        }
-
-        protected override Assembly Load(AssemblyName assemblyName)
-        {
-            string path;
-            if (_assemblyPaths.TryGetValue(assemblyName, out path) || SearchForLibrary(ManagedAssemblyExtensions, assemblyName.Name, out path))
-            {
-                return LoadFromAssemblyPath(path);
-            }
-            return null;
-        }
-
-        protected override IntPtr LoadUnmanagedDll(string unmanagedDllName)
-        {
-            string path;
-            if (_nativeLibraries.TryGetValue(unmanagedDllName, out path) || SearchForLibrary(NativeLibraryExtensions, unmanagedDllName, out path))
-            {
-                return LoadUnmanagedDllFromPath(path);
-            }
-
-            return base.LoadUnmanagedDll(unmanagedDllName);
-        }
-
-        public Assembly LoadFromPath(AssemblyName assemblyName, string path)
-        {
-            if (path == null)
-            {
-                throw new ArgumentNullException(nameof(path));
-            }
-            if(assemblyName == null)
-            {
-                throw new ArgumentNullException(nameof(assemblyName));
-            }
-            foreach(var extension in ManagedAssemblyExtensions)
-            {
-                var resolvedPath = Path.Combine(path, assemblyName.Name + extension);
-                if(File.Exists(resolvedPath))
-                {
-                    return LoadFromAssemblyPath(resolvedPath);
-                }
-            }
-            throw new FileNotFoundException(string.Format("Could not find assembly {0} in path {1}",assemblyName.Name, path));
-        }
-
-        public Assembly LoadFromPath(string path)
-        {
-            if (path == null)
-            {
-                throw new ArgumentNullException(nameof(path));
-            }
-            if (File.Exists(path))
-            {
-                return LoadFromAssemblyPath(path);
-            }
-            throw new FileNotFoundException(string.Format("Could not find assembly {0}", path));
-        }
-
-        public Assembly LoadStream(Stream assembly, Stream symbols)
-        {
-            return base.LoadFromStream(assembly, symbols);
-        }
-
         public Assembly LoadFromName(AssemblyName AssemblyName)
         {
-            if(AssemblyName == null)
-            {
-                throw new ArgumentNullException(nameof(AssemblyName));
-            }
-            return Load(AssemblyName);
-        }
-    }
-#else
-    public partial class DefaultAssemblyLoadContext : ICodeGenAssemblyLoadContext
-    {
-        static DefaultAssemblyLoadContext()
-        {
-            NativeLibraryExtensions = new[] { ".dll" };
-        }
-
-        public Assembly LoadFromPath(AssemblyName assemblyName, string path)
-        {
-            if (path == null)
-            {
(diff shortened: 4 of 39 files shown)


prafullbhosale:
cc @Eilon and @DamianEdwards for Rc2 approval

src/Microsoft.Extensions.CodeGeneration.Core/DefaultCodeGeneratorAssemblyProvider.cs
```diff
@@ -54,8 +51,12 @@ public IEnumerable<Assembly> CandidateAssemblies
                     .SelectMany(_libraryManager.GetReferencingLibraries)
                     .Distinct()
                     .Where(IsCandidateLibrary);
-                
+#if NET451
+                // For Net 451 the exe is being invoked from the build folder so all dependencies are next to the exe.
+                return list.Select(lib => _assemblyLoadContext.LoadFromName(new AssemblyName(lib.Identity.Name)));
```
NTaylorMullen: Doesn't `Assembly.Load` suffice for both of these? I thought that it was hooked up to point to the correct location.

prafullbhosale: No, for scaffolding the tool dispatches to the projectdependency command only if the user project is targeting only full .net 
The reason for this was to avoid having to build the project if the user project is targetting core.

natemcmaster: We tried this too. Do you ever instantiate and execute code from this assembly? If so, you may run into this issue: https://github.com/dotnet/cli/issues/2514

src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
```diff
@@ -245,12 +243,11 @@ public class EntityFrameworkServices : IEntityFrameworkService
             out Type dbContextType, 
             out Type modelType)
         {
-            //TODO: @prbhosal Figure out how to lookup the correct project here.             
             var projectCompilation = _workspace.CurrentSolution.Projects
-                //.Where(project => project.Name == _environment.ApplicationName)
+                .Where(project => project.AssemblyName == _environment.ApplicationName)
                 .FirstOrDefault()
```
NTaylorMullen: `FirstOrDefault` takes a predicate, no need to do the `where` above. Also, is "project" a struct? If it can't find one I'd imagine it exploding with null ref.

src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
```diff
@@ -245,12 +243,11 @@ public class EntityFrameworkServices : IEntityFrameworkService
...
                 .GetCompilationAsync().Result;
-            var newAssemblyName = projectCompilation.AssemblyName + _counter++;
+            var newAssemblyName = projectCompilation.AssemblyName;
```
NTaylorMullen: Why'd this change?

src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/project.json
```diff
@@ -6,20 +6,26 @@
...
-    "Microsoft.Extensions.CodeGeneration.Core": "1.0.0-*",
-    "Microsoft.Extensions.CodeGeneration.Templating": "1.0.0-*",
-    "Microsoft.Extensions.CodeGeneration.Utils": "1.0.0-*",
-    "System.AppContext": "4.1.0-*"
+    "Microsoft.Extensions.CodeGeneration.Core": "1.0.0-*"
   },
   "frameworks": {
+    "net451": {
+        "dependencies": {
+            "System.Threading.Tasks": "4.0.0-*"
```
NTaylorMullen: Shouldn't need to do this for net451

src/Microsoft.Extensions.CodeGeneration.Core/DefaultCodeGeneratorAssemblyProvider.cs
NTaylorMullen: For scaffolding would it really be cumbersom to dispatch for both? It'd cleanup your code paths significantly in this PR.

src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
prafullbhosale: This was to fix #196

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExporterExtensions.cs
```diff
@@ -21,29 +21,31 @@ public static string GetResolvedPathForDependency(this ILibraryExporter _library
             {
                 throw new ArgumentNullException(nameof(library));
             }
-
             var exports = _libraryExporter.GetAllExports();
+            if(exports == null)
```
NTaylorMullen: O_o how are they ever null?

src/Microsoft.Extensions.CodeGeneration.Utils/project.json
```diff
@@ -6,17 +6,24 @@
...
     "Microsoft.CodeAnalysis.CSharp.Workspaces": "1.3.0-*",
     "Microsoft.DotNet.ProjectModel": "1.0.0-*",
-    "System.AppContext": "4.1.0-*",
-    "System.Security.Cryptography.Algorithms": "4.1.0-*",
     "System.Threading.Tasks.Parallel": "4.0.1-*"
   },
   "frameworks": {
+    "net451": {
+        "dependencies": {
+            "System.Threading.Tasks":"4.0.0-*"
```
NTaylorMullen: don't need

src/dotnet-aspnet-codegenerator/CodeGenCommandExecutor.cs
```diff
@@ -0,0 +1,103 @@
...
+            AddCodeGenerationServices(serviceProvider);
+            var codeGenCommand = serviceProvider.GetService<CodeGenCommand>();
+            codeGenCommand.Execute(_codeGenArguments);
+            return 0;
+        }
+        
+        private void AddFrameworkServices(ServiceProvider serviceProvider, ProjectContext context, string nugetPackageDir)
+        {
+            var applicationEnvironment = new ApplicationEnvironment(context.RootProject.Identity.Name, context.ProjectDirectory);
+            serviceProvider.Add(typeof(IServiceProvider), serviceProvider);
```
NTaylorMullen: There's a nicer Add<T> extension method for these guys.

src/Microsoft.Extensions.CodeGeneration.Core/DefaultCodeGeneratorAssemblyProvider.cs
prafullbhosale: @natemcmaster Scaffolding does try to host the user's Startup.cs to retrieve model metadata information. 
I just tried with a project that referenced Newtonsoft.Json 8.0.3, and it seems to work fine. 
It may be because scaffolding doesn't have a dependency on ProjectModel.Loader

natemcmaster: The real reason you don't get an error is that the tool is pulling in Newtonsoft.Json 8. https://github.com/aspnet/Scaffolding/blob/81d4b51c6282df5ea2d185a2936dcfb4dba284d9/src/Microsoft.Extensions.CodeGeneration.Core/project.json#L12 Use caution with `AssemblyLoadContext`. We are ditching it completely for dotnet-ef because it doesn't provide the isolation we actually need to execute user code.

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExporterExtensions.cs
prafullbhosale: Well, ideally they shouldn't be. But I did run into Null reference exceptions on net451 process while converting the DefaultAssemblyLoadContext, so kept it here as a check.

src/dotnet-aspnet-codegenerator/Program.cs
```diff
@@ -6,127 +6,242 @@
...
+        private static bool isProjectDependencyCommand;
+
         public static void Main(string[] args)
         {
             Stopwatch stopWatch = new Stopwatch();
             stopWatch.Start();
-
             _logger = new ConsoleLogger();
+            // We don't want to expose this flag, as it only needs to be used for identifying self invocation. 
+            isProjectDependencyCommand = ToolCommandLineHelper.CheckProjectDependencyFlag(args);
```
NTaylorMullen: This can be `args.Contains`.

Also side note, we may want to consider unifying the naming for this type of flag for our assembly loading tools. I ended up using `--no-dispatch` [here](https://github.com/aspnet/RazorTooling/blob/0103843cf9a817561ccf0ea359b72f4c19355585/src/dotnet-razor-tooling/Internal/ResolveTagHelpersCommand.cs#L53).

src/dotnet-aspnet-codegenerator/Program.cs
```diff
@@ -6,127 +6,242 @@
...
+            try
+            {
+                Execute(args);
+            }
+            catch
```
NTaylorMullen: Remove

src/dotnet-aspnet-codegenerator/CodeGenCommandExecutor.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        private void AddCodeGenerationServices(ServiceProvider serviceProvider)
+        {
+            if (serviceProvider == null)
+            {
+                throw new ArgumentNullException(nameof(serviceProvider));
+            }
+
+            //Ordering of services is important here
+            serviceProvider.Add(typeof(ILogger), _logger);
```
NTaylorMullen: Should probably make this a custom logger to utilize the Microsoft.DotNet.Cli.Utils reporter. [Example](https://github.com/aspnet/RazorTooling/blob/0103843cf9a817561ccf0ea359b72f4c19355585/src/dotnet-razor-tooling/Internal/ResolveTagHelpersCommand.cs#L230)

NTaylorMullen: Or could just use the reporter directly.

src/dotnet-aspnet-codegenerator/Program.cs
prafullbhosale: :+1:  `--no-dispatch` sounds good.

src/dotnet-aspnet-codegenerator/Program.cs
```diff
@@ -6,127 +6,242 @@
...
+
+
+
+            // Define app Options;
             app.HelpOption("-h|--help");
             var projectPath = app.Option("-p|--project", "Path to project.json", CommandOptionType.SingleValue);
             var packagesPath = app.Option("-n|--nugetPackageDir", "Path to check for Nuget packages", CommandOptionType.SingleValue);
             var appConfiguration = app.Option("-c|--configuration", "Configuration for the project (Possible values: Debug/ Release)", CommandOptionType.SingleValue);
-            
+            var framework = app.Option("-tfm|--targetFramework", "Target Framework to use. (Short folder name of the tfm. eg. net451)", CommandOptionType.SingleValue);
```
NTaylorMullen: -f|--framework to be consistent.

src/dotnet-aspnet-codegenerator/CodeGenCommandExecutor.cs
prafullbhosale: I have been meaning to change to use Reporter. Will change the logger to use Reporter for now. 
Will remove the custom logger altogether post RC2.

src/dotnet-aspnet-codegenerator/Program.cs
```diff
@@ -6,127 +6,242 @@
...
+            var buildBasePath = app.Option("-b|--buildBasePath", "", CommandOptionType.SingleValue);
+            var outputPath = app.Option("-o|--buildOutputPath", "", CommandOptionType.SingleValue);
```
NTaylorMullen: VisualStudio only ever passes in buildBasePath right? I'd nuke this one since build base path and build output path are mutually exclusive in regards to how you're using them. FYI we've been trying to avoid pascal cased command line params. I heard @davidfowl felt strongly about that.

src/dotnet-aspnet-codegenerator/Program.cs
prafullbhosale: -f is used by the codegenerators to indicate force overwrite.

src/dotnet-aspnet-codegenerator/Program.cs
```diff
@@ -6,127 +6,242 @@
...
+                NuGetFramework nugetFramework = FrameworkConstants.CommonFrameworks.NetCoreApp10;
+
+                if (isProjectDependencyCommand && !TryGetNugetFramework(framework.Value(), out nugetFramework))
+                {
+                    throw new ArgumentException($"Could not understand the NuGetFramework information. Framework short folder name passed in was: {framework.Value()}");
+                }
+
+                var projectContextInfo = GetProjectContextInformation(projectPath.Value(), nugetFramework);
+
+                if (isProjectDependencyCommand && projectContextInfo.ProjectContext == null)
```
NTaylorMullen: What does `ProjectContextInformation` buy you that `ProjectContext` by itself doesn't?

src/dotnet-aspnet-codegenerator/Program.cs
```diff
@@ -6,127 +6,242 @@
...
+
+                    if (buildExitCode != 0)
+                    {
+                        //Build failed. 
+                        // Stop the process here. 
+                        return buildExitCode;
+                    }
+
+                    // Invoke the dependency command
+                    var projectDirectory = Directory.GetCurrentDirectory();
```
NTaylorMullen: the projectcontext has a projectdirectory on it. See [here](https://github.com/aspnet/RazorTooling/blob/0103843cf9a817561ccf0ea359b72f4c19355585/src/dotnet-razor-tooling/Internal/ResolveTagHelpersCommand.cs#L120)

src/dotnet-aspnet-codegenerator/Program.cs
```diff
@@ -6,127 +6,242 @@
...
-            serviceProvider.AddServiceWithDependencies<ICodeGeneratorAssemblyProvider, DefaultCodeGeneratorAssemblyProvider>();
-            serviceProvider.AddServiceWithDependencies<ICodeGeneratorLocator, CodeGeneratorsLocator>();
-            serviceProvider.AddServiceWithDependencies<CodeGenCommand, CodeGenCommand>();
-
-            serviceProvider.AddServiceWithDependencies<ICompilationService, RoslynCompilationService>();
-            serviceProvider.AddServiceWithDependencies<ITemplating, RazorTemplating>();
+            ProjectContext context = new ProjectContextBuilder()
+                    .WithProject(projectFile)
+                    .WithTargetFramework(nearestNugetFramework)
+                    //.WithRuntimeIdentifiers(PlatformServices.Default.Runtime.GetAllCandidateRuntimeIdentifiers())
```
NTaylorMullen: nuke

src/dotnet-aspnet-codegenerator/Program.cs
prafullbhosale: At this point project context is null. So cannot use it here.

src/dotnet-aspnet-codegenerator/project.json
```diff
@@ -4,21 +4,20 @@
...
+    "net451": { },
     "netcoreapp1.0": {
       "imports": [
+        "dotnet5.6",
         "dnxcore50",
         "portable-net45+win8"
-      ]
+      ],
+      "dependencies": {
+          "Microsoft.NetCore.App": { "version": "1.0.0-rc2-*", "type": "platform"}
```
NTaylorMullen: `-*`

NTaylorMullen:
:watch:

src/dotnet-aspnet-codegenerator/project.json
prafullbhosale: :+1:

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/DefaultAssemblyLoadContext.cs
```diff
@@ -144,38 +183,74 @@ public Assembly LoadFromName(AssemblyName AssemblyName)
...
-                        continue;
-                    }
+                    return Assembly.LoadFile(resolvedPath);
                 }
             }
+            throw new FileNotFoundException(string.Format("Could not find assembly {0} in path {1}",assemblyName.Name, path));
+        }
 
-            return new DefaultAssemblyLoadContext(null, null, searchPaths);
+        public Assembly LoadFromPath(string path)
```
natemcmaster: Doesn't .net provide implementations for this?

src/dotnet-aspnet-codegenerator/Program.cs
prafullbhosale: The `ProjectContext` is built for a specific TargetFramework, `ProjectContextInformation` gives all available target frameworks. 

Also the intent of having the class was to avoid having to read the project twice (Although I can avoid that by in-lining the code for `GetProjectContextInformation()` method as well)

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExporterExtensions.cs
NTaylorMullen: Hmm, sounds like a bug.

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/DefaultAssemblyLoadContext.cs
prafullbhosale: Yup, I will remove all the checks around `Assembly.LoadFile()` or just use `Assembly.LoadFrom()` here

src/dotnet-aspnet-codegenerator/Program.cs
NTaylorMullen: Ah, fair enough

src/dotnet-aspnet-codegenerator/Program.cs
NTaylorMullen: You only ever do `.TargetFrameworks.FirstOrDefault()` though. You could resolve the framework separately couldn't you?

prafullbhosale: yup, that's what I meant by in-lining the `GetProjectContextInformation()`

src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
prafullbhosale: `Microsoft.CodeAnalysis.Project` is a class

src/Microsoft.Extensions.CodeGeneration.Utils/project.json
prafullbhosale: Its needed for the compilation.Emit call [here](https://github.com/aspnet/Scaffolding/blob/pb/net45support/src/Microsoft.Extensions.CodeGeneration.Utils/CommonUtilities.cs#L22 ). The Emit method has a default parameter of type `CancellationToken` which is defined in the System.Threading.Tasks assembly. 
The code doesn't build without this.

src/Microsoft.Extensions.CodeGeneration.Core/DefaultCodeGeneratorAssemblyProvider.cs
prafullbhosale: @NTaylorMullen having to build the project for net45 is necessary, but for core it is just going to add an overhead. 

Also, if there are build errors in the user's project, scaffolding can still work for adding non ef based controllers, which will also be affected in case the build is required.

Although it might be reasonable to fail to scaffold if there are build errors, scaffolding internally compiles the project (while scaffolding an ef based controller) in memory after modifying the project's startup and adding a DbContext (modifications also in memory) 
Requiring it to do build before invocation seems wasteful.

src/dotnet-aspnet-codegenerator/Program.cs
prafullbhosale: Yeah, will remove the outputPath. 
For the pascal casing though the code generator arguments are wired to use pascal case as well. 
Will revisit it to make them all consistent.

src/Microsoft.Extensions.CodeGeneration.Utils/project.json
NTaylorMullen: Shouldn't this be a framework assembly though?

src/Microsoft.Extensions.CodeGeneration.Core/DefaultCodeGeneratorAssemblyProvider.cs
NTaylorMullen: If you don't build before scaffolding will it actually work as expected though? I can imagine a user working on their project for an extended period of time without building and then finally trying to scaffold something. Next thing they know the wrong data has been scaffolded because they didn't build and the tool didn't force it either.

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExporterExtensions.cs
```diff
@@ -21,29 +21,31 @@ public static string GetResolvedPathForDependency(this ILibraryExporter _library
...
+            {
+                return string.Empty;
+            }
             var assets = exports
-                .SelectMany(export => export.RuntimeAssemblyGroups.GetDefaultGroup().Assets)
+                .SelectMany(export => export.RuntimeAssemblyGroups?.GetDefaultGroup()?.Assets)
```
NTaylorMullen: Adding `?` makes the next line explode if any item ends up being `null`.

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExporterExtensions.cs
```diff
@@ -21,29 +21,31 @@ public static string GetResolvedPathForDependency(this ILibraryExporter _library
...
                 .Where(asset => asset.Name == library.Identity.Name);
-
-            if (assets.Any())
+            if (assets != null && assets.Any())
```
NTaylorMullen: assets can't be null, you just did a `SelectMany` on it, would have null ref'd, same below.

NTaylorMullen: Also, `.Any` takes a predicate.

src/Microsoft.Extensions.CodeGeneration.Utils/Workspaces/ProjectJsonWorkspace.cs
```diff
@@ -214,7 +212,11 @@ private static CSharpCompilationOptions AddSigningOptions(CSharpCompilationOptio
             {
                 keyFile = Path.GetFullPath(Path.Combine(projectDirectory, compilerOptions.KeyFile));
 
-                if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows) || useOssSigning)
+                if (
+#if !NET451
```
NTaylorMullen: mono?

src/dotnet-aspnet-codegenerator/CodeGenCommandExecutor.cs
```diff
@@ -0,0 +1,102 @@
...
+            _codeGenArguments = codeGenArguments;
+            _configuration = configuration;
+            _logger = logger;
+            _nugetPackageDir = nugetPackageDir;
+        }
+        
+        public int Execute()
+        {
+            var serviceProvider = new ServiceProvider();
+            Directory.SetCurrentDirectory(_projectContext.ProjectDirectory);
```
NTaylorMullen: Curious: Why do you need to do this?

src/dotnet-aspnet-codegenerator/Program.cs
NTaylorMullen: Curious then, why do we ever need to expect `ProjectContext` to be null? `ProjectContext` construction only relies on a project being restored (a lock file). And that's the same expectation as the rest of the code here. 

Just asking because always creating a `ProjectContext` could simplify a lot of what you have here. Unless I'm missing something of course :smile:

src/Microsoft.Extensions.CodeGeneration.Core/DefaultCodeGeneratorAssemblyProvider.cs
prafullbhosale: Well scaffolding, does internally build the user's project in memory using Roslyn.

src/dotnet-aspnet-codegenerator/Program.cs
prafullbhosale: Ok, lets take a step back. 
Here (when its not the dependency command) the portable app (tool) was trying with tfm as netcoreapp1.0. If the user's project doesn't target netcoreapp1.0, the tool doesn't actually build the project context, hence its null.

src/dotnet-aspnet-codegenerator/CodeGenCommandExecutor.cs
prafullbhosale: This is a remnant of some old issue, that needed the directory to be set (I am not sure, but I think it had to do something with the project exports.) should not be needed anymore. 
Will remove.

src/Microsoft.Extensions.CodeGeneration.Utils/Workspaces/ProjectJsonWorkspace.cs
prafullbhosale: Currently scaffolding is supported only on windows through VS. 
Also this is code copied from ProjectModel.Workspace. 

Once that targets net451, this code will be removed altogether.

src/dotnet-aspnet-codegenerator/CodeGenCommandExecutor.cs
prafullbhosale: The logger is now using Reporter.

prafullbhosale:
@NTaylorMullen This is now tested on both net451 and netcoreapp1.0
The tests are also ported to target both net451 and netcoreapp1.0 
(Although the net451 tests are failing due the assembly redirect issue, I believe that is fixed by dotnet/cli/pull/2586)

src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
```diff
@@ -245,14 +243,19 @@ public class EntityFrameworkServices : IEntityFrameworkService
             out Type dbContextType, 
             out Type modelType)
         {
-            //TODO: @prbhosal Figure out how to lookup the correct project here.             
             var projectCompilation = _workspace.CurrentSolution.Projects
-                //.Where(project => project.Name == _environment.ApplicationName)
-                .FirstOrDefault()
+                .FirstOrDefault(project => project.AssemblyName == _environment.ApplicationName)
```
NTaylorMullen: `.First(...)`

src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
```diff
@@ -400,10 +403,10 @@ private DbContext TryCreateContextUsingAppCode(Type dbContextType, ModelType sta
         {
             try {
                 var builder = new WebHostBuilder();
-                //TODO: Review 
-                builder.UseServer("Microsoft.AspNetCore.Server.WebListener")
-                        .UseContentRoot(Directory.GetCurrentDirectory());
-                        
+                builder.UseKestrel()
+                        .UseContentRoot(Directory.GetCurrentDirectory())
+                        .UseIISIntegration();
```
NTaylorMullen: Why add IISIntegration?

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/DefaultAssemblyLoadContext.cs
```diff
@@ -7,175 +7,33 @@
...
+                return Assembly.Load(ms.ToArray());
             }
-            return Load(AssemblyName);
+#else
+            return System.Runtime.Loader.AssemblyLoadContext.Default.LoadFromStream(assembly);
+#endif
         }
 
-        public static ICodeGenAssemblyLoadContext CreateAssemblyLoadContext(string nugetPackageDir)
+        public static DefaultAssemblyLoadContext CreateAssemblyLoadContext()
```
NTaylorMullen: Just do `new DefaultAssemblyLoadContext()` wherever you call this.

src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
prafullbhosale: This was just to keep it consistent with the Program.cs from templates. I will remove the IIS integration and test if it works.

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExporterExtensions.cs
```diff
@@ -50,5 +48,14 @@ public static string GetResolvedPathForDependency(this ILibraryExporter _library
 
             return string.Empty;
         }
+
+        private static IEnumerable<LibraryAsset> GetAssets(IEnumerable<LibraryAssetGroup> group)
+        {
+            if (group == null || group.GetDefaultGroup() == null)
```
NTaylorMullen: `if (group?.GetDefaultGroup == null)`

src/Microsoft.Extensions.CodeGeneration.Utils/DotNet/LibraryExporterExtensions.cs
```diff
@@ -50,5 +48,14 @@ public static string GetResolvedPathForDependency(this ILibraryExporter _library
...
+            {
+                return new List<LibraryAsset>();
```
NTaylorMullen: `Enumerable.Empty<LibraryAsset>()`

src/dotnet-aspnet-codegenerator/DotNetBuildCommandHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.DotNet.Cli.Utils;
+using NuGet.Frameworks;
+
+namespace Microsoft.Extensions.CodeGeneration
+{
+    internal class DotNetBuildCommandHelper
+    {
+        internal static int Build(
```
NTaylorMullen: You and EF both do this. Could be worth moving to common.

/cc @natemcmaster

src/dotnet-aspnet-codegenerator/DotNetBuildCommandHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            // TODO: Specify --runtime?
+            var args = new List<string>()
+            {
+                project,
+                "--configuration", configuration,
+                "--framework", framework.GetShortFolderName(),
+            };
+
+            if (buildBasePath != null)
```
NTaylorMullen: Should do what's done [here](https://github.com/aspnet/Common/blob/cbcc732a7e9bc51debdf4c5db5ad19de4ba64d75/src/Microsoft.Extensions.DotnetToolDispatcher.Sources/DotnetToolDispatcher.cs#L47-L51).

src/dotnet-aspnet-codegenerator/Program.cs
```diff
@@ -6,127 +6,218 @@
...
+
         public static void Main(string[] args)
         {
+
             Stopwatch stopWatch = new Stopwatch();
             stopWatch.Start();
-
             _logger = new ConsoleLogger();
+            // We don't want to expose this flag, as it only needs to be used for identifying self invocation. 
+            isProjectDependencyCommand = args.Contains("--no-dispatch");
```
NTaylorMullen: If you transition to using the `DotnetToolDispatcher` (will be getting it in today) you wont have to do this work.

src/dotnet-aspnet-codegenerator/Program.cs
```diff
@@ -6,127 +6,218 @@
...
+            _logger.LogMessage($"Is Dependency COmmand: {isProjectDependencyCommand}");
```
NTaylorMullen: `Command`

src/dotnet-aspnet-codegenerator/Program.cs
```diff
@@ -6,127 +6,218 @@
...
             };
+
+            // Define app Options;
             app.HelpOption("-h|--help");
             var projectPath = app.Option("-p|--project", "Path to project.json", CommandOptionType.SingleValue);
             var packagesPath = app.Option("-n|--nugetPackageDir", "Path to check for Nuget packages", CommandOptionType.SingleValue);
             var appConfiguration = app.Option("-c|--configuration", "Configuration for the project (Possible values: Debug/ Release)", CommandOptionType.SingleValue);
-            
+            var framework = app.Option("-tfm|--targetFramework", "Target Framework to use. (Short folder name of the tfm. eg. net451)", CommandOptionType.SingleValue);
+            var buildBasePath = app.Option("-b|--buildBasePath", "", CommandOptionType.SingleValue);
```
NTaylorMullen: Shouldn't be pascal casing these params.

`nuget-package-dir`, `target-framework`, `build-base-path`

src/dotnet-aspnet-codegenerator/Program.cs
prafullbhosale: Will also change it to log only in verbose mode.

src/dotnet-aspnet-codegenerator/Program.cs
```diff
@@ -6,127 +6,218 @@
...
             {
-                var serviceProvider = new ServiceProvider();
-                var context = CreateProjectContext(projectPath.Value());
-                
+                string project = projectPath.Value();
+                if (string.IsNullOrEmpty(project))
+                {
+                    project = Directory.GetCurrentDirectory();
+                }
                 var configuration = appConfiguration.Value() ?? "Debug";
```
NTaylorMullen: https://github.com/aspnet/Common/blob/cbcc732a7e9bc51debdf4c5db5ad19de4ba64d75/src/Microsoft.Extensions.DotnetToolDispatcher.Sources/DotnetToolDispatcher.cs#L53

src/dotnet-aspnet-codegenerator/Program.cs
```diff
@@ -6,127 +6,218 @@
...
+                    if(nearestNugetFramework == null)
+                    {
+                        // This should never happen as long as we dispatch correctly.
+                        var msg = "Could not find a compatible framework to execute."
+                            + Environment.NewLine
+                            +$"Available frameworks in project:{string.Join($"{Environment.NewLine} -", frameworksInProject.Select(f => f.GetShortFolderName()))}";
+                        throw new InvalidOperationException(msg);
+                    }
+
+                    ProjectContext context = new ProjectContextBuilder()
```
NTaylorMullen: You no longer use this. The CodeGenCommand doesn't seem to either. I missing something?

NTaylorMullen:
:watch:

src/dotnet-aspnet-codegenerator/Program.cs
prafullbhosale: The context is being passed to the CodeGenCommandExecutor

NTaylorMullen: It's not used there though

prafullbhosale: Not directly, but its used to create the project's workspace and also to define the applicationEnvironment Name and directory
https://github.com/aspnet/Scaffolding/pull/197/files#diff-9ada89cdefcec186eb76ee394fbf2060R65

src/dotnet-aspnet-codegenerator/DotNetBuildCommandHelper.cs
natemcmaster: Yes. As long as we are trying to share implementation, we may as well move this too.

src/dotnet-aspnet-codegenerator/Program.cs
NTaylorMullen: Ah I see, that was the piece I was missing :+1:

prafullbhosale:
:up: :date:

src/Microsoft.Extensions.CodeGeneration.EntityFrameworkCore/project.json
```diff
@@ -6,20 +6,22 @@
   },
   "dependencies": {
     "Microsoft.AspNetCore.Hosting": "1.0.0-*",
-    "Microsoft.AspNetCore.Server.WebListener": "0.1.0-*",
+    "Microsoft.AspNetCore.Server.IISIntegration": "1.0.0-*",
```
NTaylorMullen: Nuke

NTaylorMullen:
:shipit:

PR #198, opened 2016-04-20
Add [ValidateAntiForgeryToken] for all post methods

prafullbhosale:
cc @phenning @barrytang 
@Eilon and @DamianEdwards  for RC2 approval.


Eilon:
@rynowak @blowdart for the code review.

This shouldn't need RC2 approval because this is tooling, and tooling is wide open right now.

blowdart:
Oh, there are API Controllers. Those should be removed, as it's very dependent on how you call them

phenning:
Same comment as @blowdart, otherwise looks good.

prafullbhosale:
Removed the token for API controllers. 

cc @vkichline

prafullbhosale:
Spoke to @phenning offline. 
This is ok to merge.

PR #202, opened 2016-04-22
Renaming the package to Microsoft.Extensions.CodeGeneration.Tools

moozzyk:


moozzyk:
/cc @prafullbhosale 

I ran tests and ran the tool manually after the rename and got to this: point: 

```
dotnet.exe aspnet-codegenerator view -partial -m "Program"
Finding the generator 'view'...
Running the generator 'view'...
The ViewName cannot be empty"
   at Microsoft.Extensions.CodeGeneration.ActionInvoker.<BuildCommandLine>b__6_0()
RunTime 00:00:02.11
   at Microsoft.Extensions.CommandLineUtils.CommandLineApplication.Execute(String[] args)
   at Microsoft.Extensions.CodeGeneration.ActionInvoker.Execute(String[] args)
   at Microsoft.Extensions.CodeGeneration.CodeGenCommand.Execute(String[] args)
```

which I think it's fine since it is after the tool called into `Microsoft.Extensions.CodeGenerators.Mvc` which is specified in project dependencies.

prafullbhosale:
@moozzyk I am assuming that this change will still be valid for the tools that are specified in both `'tools'` and `'dependencies'` sections of project.json. 
Is that correct?

prafullbhosale:
cc @phenning @rustd as FYI. This will require templates to be changed.

moozzyk:
@prafullbhosale - how do I test the other scenario? I did fix https://github.com/dotnet/cli/issues/2592 yesterday but would like to test it end to end.

phenning:
Let me know when the package rename is in and I'll monitor the builds for the change.

prafullbhosale:
@moozzyk the changes for that are still in PR. #197

prafullbhosale:
@moozzyk #197 is now merged.

moozzyk:
@prafullbhosale - Thanks - I was waiting for the build system to pick up a new version of CLI. But since you merged I can try the latest available even if it is not in the build system...

PR #207, opened 2016-04-26
Change netstandardapp to netcoreapp

BrennanConroy:
@pakrym @Eilon


prafullbhosale:
There is a PR out for supporting net451 projects, that includes this change as well. 
https://github.com/aspnet/Scaffolding/pull/197

Issue #210, opened 2016-04-27
Build base path not passed correctly to dotnet build during build and dispatch phase


prafullbhosale:
If the path is already rooted, it doesn't get passed to the build, and leads to scaffolding to fail 
https://github.com/aspnet/Scaffolding/blob/dev/src/dotnet-aspnet-codegenerator/DotNetBuildCommandHelper.cs#L30 

Also, the warnings from build need to be filtered out,  as VS looks at stdErr of the process and shows error dialog if there is anything on the stdErr.

Issue #206, opened 2016-04-26
Use DotnetToolDispatcher for dispatching project dependency command in scaffolding


prafullbhosale:
aspnet/Coherence-Signed/issues/276

PR #213, opened 2016-04-27
Pb/rc2bugfix

prafullbhosale:
Addresses #210 and #206 
cc @NTaylorMullen 

@Eilon, @barrytang

### DIFF
--- a/src/dotnet-aspnet-codegenerator/ToolCommandLineHelper.cs
+++ b/src/dotnet-aspnet-codegenerator/ToolCommandLineHelper.cs
@@ -17,6 +17,7 @@ namespace Microsoft.Extensions.CodeGeneration
         private const string PROJECT = "--project";
         private const string NUGET_PACKAGE_DIR = "--nuget-package-dir";
         private const string BUILD_BASE_PATH = "--build-base-path";
+        private const string DISPATCHER_VERSION = "--dispatcher-version";
 
         private const string NO_DISPATCH_FLAG_SHORT = "-nd";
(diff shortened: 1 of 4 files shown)


src/dotnet-aspnet-codegenerator/ToolCommandLineHelper.cs
```diff
@@ -17,6 +17,7 @@ internal class ToolCommandLineHelper
         private const string PROJECT = "--project";
         private const string NUGET_PACKAGE_DIR = "--nuget-package-dir";
         private const string BUILD_BASE_PATH = "--build-base-path";
+        private const string DISPATCHER_VERSION = "--dispatcher-version";
```
NTaylorMullen: Hmm, if you need this might be worth making a PR to common to expose it so it's not hardcoded here.

NTaylorMullen:
:shipit:

src/dotnet-aspnet-codegenerator/project.json
```diff
@@ -4,7 +4,8 @@
     "emitEntryPoint": true
   },
   "dependencies": {
-    "Microsoft.Extensions.CodeGeneration": "1.0.0-*"
+    "Microsoft.Extensions.CodeGeneration": "1.0.0-*",
+    "Microsoft.Extensions.DotnetToolDispatcher.Sources": "1.0.0-*"
```
NTaylorMullen: Shoot, I missed this. It needs to be a build time dependency only.

javiercn: I'll fix this

PR #236, opened 2016-05-19
Changes to write connection string to appsettings.json

prafullbhosale:
Porting changes from https://github.com/aspnet/Scaffolding/tree/pk/155791

The original issue of app reload when appsettings is modified is resolved already.

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/DbContextEditorServices.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/DbContextEditorServices.cs
@@ -13,6 +13,7 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Text;
 using Microsoft.VisualStudio.Web.CodeGeneration.DotNet;
 using Microsoft.VisualStudio.Web.CodeGeneration.Templating;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore
 {
@@ -22,17 +23,29 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore
         private readonly ILibraryManager _libraryManager;
         private readonly ITemplating _templatingService;
         private readonly IFilesLocator _filesLocator;
+        private readonly IFileSystem _fileSystem;
 
         public DbContextEditorServices(
             ILibraryManager libraryManager,
             IApplicationInfo applicationInfo,
             IFilesLocator filesLocator,
             ITemplating templatingService)
+            : this (libraryManager, applicationInfo, filesLocator, templatingService, DefaultFileSystem.Instance)
+        {
+        }
+
+        internal DbContextEditorServices(
+            ILibraryManager libraryManager,
+            IApplicationInfo applicationInfo,
+            IFilesLocator filesLocator,
+            ITemplating templatingService,
+            IFileSystem fileSystem)
         {
             _libraryManager = libraryManager;
             _applicationInfo = applicationInfo;
             _filesLocator = filesLocator;
             _templatingService = templatingService;
+            _fileSystem = fileSystem;
         }
 
         public async Task<SyntaxTree> AddNewContext(NewDbContextTemplateModel dbContextTemplateModel)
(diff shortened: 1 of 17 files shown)


prafullbhosale:
cc @abpiskunov

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/DbContextEditorServices.cs
```diff
@@ -171,6 +187,76 @@ public EditSyntaxTreeResult EditStartupForNewContext(ModelType startUp, string d
...
+                .Select(m => m as IPropertySymbol)
+                .Where(s => s != null);
+
+            foreach (var pSymbol in propertySymbols)
+            {
+                var namedType = pSymbol.Type as INamedTypeSymbol; //When can this go wrong?
+                if (namedType != null &&
+                    namedType.ContainingAssembly.Name == "Microsoft.Extensions.Configuration.Abstractions" &&
+                    namedType.ContainingNamespace.ToDisplayString() == "Microsoft.Extensions.Configuration" &&
+                    namedType.Name == "IConfigurationRoot") // What happens if the type is referenced in full in code??
```
abpiskunov: Did you test that already?

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/DbContextEditorServices.cs
```diff
@@ -171,6 +187,76 @@ public EditSyntaxTreeResult EditStartupForNewContext(ModelType startUp, string d
...
+                content[dataNodeName] = new JObject();
+            }
+
+            if (content[dataNodeName][connectionStringName] == null)
+            {
+                writeContent = true;
+                content[dataNodeName][connectionStringName] = new JObject();
+            }
+
+            if (content[dataNodeName][connectionStringName][connectionStringNodeName] == null)
```
abpiskunov: Instead doing this you might want to have a model/class which then would be automatically converted to json (see how runtime read/write to project.lock.json)

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/DbContextEditorServices.cs
```diff
@@ -122,18 +135,21 @@ public EditSyntaxTreeResult EditStartupForNewContext(ModelType startUp, string d
...
+                var configRootProperty = TryGetIConfigurationRootProperty(startUp.TypeSymbol);
+
+                if (configServicesMethod != null && configRootProperty != null)
                 {
                     var servicesParam = configServicesMethod.ParameterList.Parameters
                         .FirstOrDefault(p => p.Type.ToString() == "IServiceCollection") as ParameterSyntax;
 
                     if (servicesParam != null)
                     {
+                        AddConnectionString(dbContextTypeName, dataBaseName);
```
abpiskunov: it is not called if configRootProperty = null , it is different form original code flow - is that intentional?

abpiskunov:
:shipit:

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/DbContextEditorServices.cs
prafullbhosale: Yes, in case the configRootProperty is null, it throws an error saying that it needs that property. 
If the property isn't defined, we could potentially fall back to old way of adding the connection string in the Startup.cs, but that means the user will manually have to do that later on anyways.  

---
In reply to: [66547326](https://github.com/aspnet/Scaffolding/pull/236#discussion_r66547326) [](ancestors = 66547326)

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/DbContextEditorServices.cs
prafullbhosale: Yes, I did test this manually and it does work irrespective if the property is defined by using IConfigurationRoot or Microsoft.Extensions.Configuration.IConfigurationRoot

---
In reply to: [66546977](https://github.com/aspnet/Scaffolding/pull/236#discussion_r66546977) [](ancestors = 66546977)

PR #238, opened 2016-05-24
Add controllerNamespace parameter to Controller generator

### LINKED ISSUE #229: Scaffolding a controller does not generate the correct namespace
Example:

In a project named `MyProject`, Scaffold a controller from within `ApiControllers` folder

Expected Namespace:
`namespace MyProject.ApiControllers`

Actual Namespace:
`namespace MyProject.Controllers`

prafullbhosale:
Fixes #229 by accepting a namespace parameter from user instead of always using `<appName>.Controllers` as the namespace. 

If the param is not provided, it falls back to old behavior, but checks if the `<appName>.Controllers` is valid or not. If not just uses `Controllers` as the namespace. 

cc @abpiskunov, @sayedihashimi

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Controller/ControllerGeneratorBase.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Controller/ControllerGeneratorBase.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.Web.CodeGeneration.DotNet;
 using Microsoft.VisualStudio.Web.CodeGeneration;
@@ -83,13 +84,21 @@ namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Controller
             }
         }
 
-        protected string GetControllerNamespace()
+        protected string GetDefaultControllerNamespace(string relativeFolderPath)
         {
-            // Review: MVC scaffolding used ActiveProject's MSBuild RootNamespace property
-            // That's not possible in command line scaffolding - the closest we can get is
-            // the name of assembly??
-            var appName = ApplicationInfo.ApplicationName;
-            return appName + "." + Constants.ControllersFolderName;
+            return NameSpaceUtilities.GetSafeNameSpaceFromPath(relativeFolderPath, ApplicationInfo.ApplicationName);
+        }
+
+        protected void ValidateNameSpaceName(CommandLineGeneratorModel generatorModel)
+        {
+            if (!string.IsNullOrEmpty(generatorModel.ControllerNamespace) &&
+                !RoslynUtilities.IsValidNamespace(generatorModel.ControllerNamespace))
+            {
+                throw new InvalidOperationException(string.Format(
+                    CultureInfo.CurrentCulture,
+                    MessageStrings.InvalidNamespaceName,
+                    generatorModel.ControllerNamespace));
+            }
         }
 
         protected string ValidateAndGetOutputPath(CommandLineGeneratorModel generatorModel)
--- a/src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Controller/ControllerWithContextGenerator.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Controller/ControllerWithContextGenerator.cs
@@ -78,7 +78,7 @@ namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Controller
         public override async Task Generate(CommandLineGeneratorModel controllerGeneratorModel)
         {
             Contract.Assert(!String.IsNullOrEmpty(controllerGeneratorModel.ModelClass));
-
+            ValidateNameSpaceName(controllerGeneratorModel);
             string outputPath = ValidateAndGetOutputPath(controllerGeneratorModel);
             var modelTypeAndContextModel = await ValidateModelAndGetMetadata(controllerGeneratorModel);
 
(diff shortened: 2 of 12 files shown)


src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Controller/ControllerGeneratorBase.cs
```diff
@@ -83,13 +84,31 @@ protected virtual IEnumerable<string> TemplateFolders
...
+            // If the app name is not a valid namespace, then just use Controllers as default. 
+            return RoslynUtilities.IsValidNamespace(defaultControllerNamespace)
+                ? defaultControllerNamespace
+                : Constants.ControllersFolderName;
+        }
+
+        protected void ValidateNameSpaceName(CommandLineGeneratorModel generatorModel)
+        {
+            if (!string.IsNullOrEmpty(generatorModel.ControllerNamespace) &&
+                RoslynUtilities.IsValidNamespace(generatorModel.ControllerNamespace))
```
abpiskunov: Throws if IsValidNamespace returns true?

src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Controller/ControllerGeneratorBase.cs
```diff
@@ -83,13 +84,31 @@ protected virtual IEnumerable<string> TemplateFolders
...
-        protected string GetControllerNamespace()
+        protected string GetDefaultControllerNamespace()
         {
             // Review: MVC scaffolding used ActiveProject's MSBuild RootNamespace property
             // That's not possible in command line scaffolding - the closest we can get is
             // the name of assembly??
             var appName = ApplicationInfo.ApplicationName;
-            return appName + "." + Constants.ControllersFolderName;
+
+            var defaultControllerNamespace = appName + "." + Constants.ControllersFolderName;
```
abpiskunov: Issue 229 complained about Controllers instead of ApiControllers. Why don't we look at current folder name and take it first as default namespace?

src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Controller/ControllerGeneratorBase.cs
```diff
@@ -83,13 +84,31 @@ protected virtual IEnumerable<string> TemplateFolders
...
+
```
abpiskunov: Issue 155 also brought a valid point : you know location of project.json don't you? xproj if it exist should be located in the same folder. If it is the only xproj in the folder you could get its <AssemblyName>, if there are several xproj's in the folder , try to find the one with name ="containing folder name" (?). In all other cases use your default existing logic..

src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Controller/ControllerGeneratorBase.cs
abpiskunov: Or allowing user specify namespace is a workaround for both issues 229 and 155?

src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Controller/ControllerWithContextGenerator.cs
```diff
@@ -87,13 +87,15 @@ public class ControllerWithContextGenerator : ControllerGeneratorBase
                 //Todo: Pluralize model name
                 controllerGeneratorModel.ControllerName = modelTypeAndContextModel.ModelType.Name + Constants.ControllerSuffix;
             }
-
+            var namespaceName = string.IsNullOrEmpty(controllerGeneratorModel.ControllerNamespace)
```
abpiskunov: Tests?

src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Controller/MvcController.cs
```diff
@@ -38,11 +38,13 @@ public abstract class MvcController : ControllerGeneratorBase
             {
                 throw new ArgumentException(GetRequiredNameError);
             }
-
+            ValidateNameSpaceName(controllerGeneratorModel);
             var layoutDependencyInstaller = ActivatorUtilities.CreateInstance<MvcLayoutDependencyInstaller>(ServiceProvider);
             await layoutDependencyInstaller.Execute();
-
-            var templateModel = new ClassNameModel(className: controllerGeneratorModel.ControllerName, namespaceName: GetControllerNamespace());
+            var namespaceName = string.IsNullOrEmpty(controllerGeneratorModel.ControllerNamespace)
```
abpiskunov: Tests?

src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Controller/ControllerGeneratorBase.cs
```diff
@@ -83,13 +84,31 @@ protected virtual IEnumerable<string> TemplateFolders
...
+                ? defaultControllerNamespace
+                : Constants.ControllersFolderName;
+        }
+
+        protected void ValidateNameSpaceName(CommandLineGeneratorModel generatorModel)
+        {
+            if (!string.IsNullOrEmpty(generatorModel.ControllerNamespace) &&
+                RoslynUtilities.IsValidNamespace(generatorModel.ControllerNamespace))
+            {
+                throw new InvalidOperationException(string.Format(
```
abpiskunov: Tests?

abpiskunov:
:watch:

src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Controller/ControllerGeneratorBase.cs
prafullbhosale: I don't think looking for a .xproj file is the right approach. Agreed that by convention, it would be in the same folder as project.json and we could do the above heuristic approach to find the xproj when there are multiple. But unless we have a foolproof way of telling that a xproj is the correct one and the current intended one, using it is as good as guessing that it *might* be what user intended. 

The whole idea behind exposing the controllerNamespace is two fold: 
1. Scaffolding currently is still intended to be a Visual Studio feature and Visual studio already has mechanisms of finding the right namespace.
2. It also gives command line users the flexibility to specify the namespace if they do not want to follow the default pattern. 

---
In reply to: [64578045](https://github.com/aspnet/Scaffolding/pull/238#discussion_r64578045) [](ancestors = 64578045)

src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Controller/ControllerGeneratorBase.cs
prafullbhosale: Its a workaround for both, but since we have access to the relative path, we can try to use it to determine the default namespace. 

---
In reply to: [64578230](https://github.com/aspnet/Scaffolding/pull/238#discussion_r64578230) [](ancestors = 64578230)

src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Controller/ControllerWithContextGenerator.cs
prafullbhosale: Need to spend some time to come up with end to end tests here. 
Currently end to end tests are in the Visual Studio code base.

prafullbhosale:
:up: :date:

abpiskunov:
:shipit:

PR #239, opened 2016-05-24
Compute the display property for the navigation metadata.

### LINKED ISSUE #232: Model with navigation properties does not work
Repro:

Model `Blog.cs`:
```c#
    public class Blog
    {
        public int Id { get; set; }
        public string Title { get; set; }
        public string BloggerName { get; set; }
        public virtual ICollection<Post> Posts { get; set; }
    }
```

Model `Post.cs`:
```c#
    public class Post
    {
        public int Id { get; set; }
        public string Title { get; set; }
        public DateTime DateCreated { get; set; }
        public string Content { get; set; }
        public int BlogId { get; set; }

        [ForeignKey("BlogId")]
        public Blog Blog { get; set; }
    }
```

Now scaffold a controller for `Post.cs` and look at the `Create()` method in `PostsController`

Expectation: (Same as MVC5 behavior)
```c#
        public IActionResult Create()
        {
            ViewData["BlogId"] = new SelectList(_context.Blog, "Id", "Title");
            return View();
        }
```

Actual: 
```c#
        public IActionResult Create()
        {
            ViewData["BlogId"] = new SelectList(_context.Blog, "Id", "Blog");
            return View();
        }
```

This leads to `NullReferenceException` when `localhost:5000/Posts/Create` is hit

prafullbhosale:
Fixes #232 

The `Displayname` property is used in the [MvcControllerWithContext.cshtml](https://github.com/aspnet/Scaffolding/blob/dev/src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Templates/ControllerGenerator/MvcControllerWithContext.cshtml#L106) to create a `SelectList`

In MVC5 the behavior was if the `Navigation.TargetType` has a string property it would use it as the display property. 
If there is no string property it would use the primary key from the navigation. 
Here we do the same thing, except we also check if the string column is nullable or not. 
If it is nullable, we fall back to using the navigation's primary key.  

cc @natemcmaster

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/NavigationMetadata.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/NavigationMetadata.cs
@@ -21,7 +21,6 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore
             Contract.Assert(navigation.IsDependentToPrincipal());
 
             AssociationPropertyName = navigation.Name;
-            DisplayPropertyName = AssociationPropertyName; //Needs further implementation
 
             var otherEntityType = navigation.ForeignKey.ResolveOtherEntityType(navigation.DeclaringEntityType);
             EntitySetName = ModelMetadata.GetEntitySetName(dbContextType, otherEntityType.ClrType);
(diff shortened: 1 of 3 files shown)


natemcmaster:
cc @AndriySvyryd @smitpatel does this use of metadata API look okay? I'm not familiar enough with this part of EF to sign off

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/NavigationMetadata.cs
```diff
@@ -32,6 +31,20 @@ public NavigationMetadata(INavigation navigation, Type dbContextType)
...
                 .ToArray();
+
+            // The default for the display property is the primary key of the navigation. 
+            DisplayPropertyName = PrimaryKeyNames[0];
+
+            // If there is a non nullable string property in the navigation's target type, we use that instead. 
+            var displayPropertyCandidate = navigation
+                .GetTargetType()
+                .GetDeclaredProperties()
+                .FirstOrDefault(p => !p.IsColumnNullable() && p.ClrType == typeof(string));
```
AndriySvyryd: As the name implies `IsColumnNullable()` returns the corresponding column nullability, wouldn't calling `IsNullable()` be more appropriate here?

Also `GetDeclaredProperties()` does not return any inherited properties, shouldn't they be considered as well?

prafullbhosale:
:up::date:

AndriySvyryd:
:shipit:

PR #255, opened 2016-06-17
Look for model type in dependencies if its not found in the project directly.

### LINKED ISSUE #251: Scaffolding fails if model class is in a dependency (project/ library) of the project on which scaffolding is being run.
Project A has a dependeny on Project B. 
Project B has model class 
If you try to run scaffolding on Project A by using model class from Project B, it fails with the below error: 
`No model type returned for type: <class name>`

image

Related #249 #140

prafullbhosale:
Fixes #251

cc @abpiskunov

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
@@ -254,7 +254,7 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore
                 {
                     throw new InvalidOperationException(MessageStrings.DbContextCreationError_noTypeReturned);
                 }
-                modelType = result.Assembly.GetType(modelTypeName);
+                modelType = GetModelTypeFromAssembly(modelTypeName, result.Assembly);
                 if (modelType == null)
                 {
                     throw new InvalidOperationException("No Model Type returned for type: " + modelTypeName);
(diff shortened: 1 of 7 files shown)


barrytang:
TFS Bug#<a href="https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workItems#id=233431&fullScreen=false&_a=edit">233431</a>

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
```diff
@@ -289,6 +290,36 @@ public class EntityFrameworkServices : IEntityFrameworkService
             return true;
         }
 
+        // Look for the model type in the current project. 
+        // If its not found in the current project, look in the dependencies.
+        private Type GetModelTypeFromAssembly(string modelTypeName, Assembly assembly)
+        {
+            Type modelType = assembly.GetType(modelTypeName);
```
abpiskunov: GetType can throw several kinds of exceptions, do we protect from it?

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
```diff
@@ -289,6 +290,36 @@ public class EntityFrameworkServices : IEntityFrameworkService
...
+            if (modelType == null)
+            {
+                // Need to look in the dependencies of this project now.
+                var dependencies = assembly.GetReferencedAssemblies().GetEnumerator();
```
abpiskunov: I don't remember , but does build happens before we do that? If not assume user added a new model in the project B and did not build yet, then tries to scaffold in project A, which would find Project B assembly, but no new type yet... What is the plan for this case?

prafullbhosale: At this point a build has already happened. 
Also, before reaching here, we already make sure that the modelType actually exists using the ModelTypesLocator. 

In this particular part, we are trying to get the modelType from the in-memory compilation of the project (and its dependencies) after the DbContext has been added/ modified.

abpiskunov:
:shipit:

PR #264, opened 2016-06-28
Get distinct model types from projects and dependencies

### LINKED ISSUE #249: Error multiple types matching the name while Scaffolding ASP.NET Core MVC RC2
I'm developing a web app in Asp.Net Core RC2 according to n-layered architecture.

f

When I try to use scaffolding to generate CRUD Controller and Views, I get the following error.

e

I have separated my Models/ViewModels and Context/Repository Classes into Business Logic Layer and Data Access Layer respectively. I'm referencing both these layers in OpenBiz project as nuget packages. But cannot perform scaffolding as seen above. Is there an issue with my architecture or am I missing something here?

prafullbhosale:
Fixes #249 

cc @abpiskunov

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.Core/ModelTypesLocator.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.Core/ModelTypesLocator.cs
@@ -34,13 +34,12 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration
 
         public IEnumerable<ModelType> GetAllTypes()
         {
-
             return _projectWorkspace.CurrentSolution.Projects
                 .Select(project => project.GetCompilationAsync().Result)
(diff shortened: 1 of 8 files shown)


src/Microsoft.VisualStudio.Web.CodeGeneration.Core/ModelTypesLocator.cs
```diff
@@ -66,5 +65,37 @@ public IEnumerable<ModelType> GetType(string typeName)
...
+                if (Object.ReferenceEquals(x, y))
+                {
+                    return true;
+                }
+                if (Object.ReferenceEquals(x, null) || Object.ReferenceEquals(y, null))
+                {
+                    return false;
+                }
+                return x.Name == y.Name
+                    && x.ContainingNamespace.Name == y.ContainingNamespace.Name
```
abpiskunov: ContainingNamespace is never null?

abpiskunov:
:shipit:

PR #266, opened 2016-06-29
Add --no-build flag to dotnet-aspnet-codegenerator. Fixes #261

### LINKED ISSUE #261: Add a --no-build flag to dotnet-aspnet-codegenerator
Currently for every invocation, scaffolding always calls `dotnet build`. 
There should be a way to make this optional. 
One example of scenario where this might be useful is the user is running the project in IIS express, running `dotnet build` would cause failures if the project is `not safe for incremental compilation`
This will cause scaffolding to fail.

prafullbhosale:
cc @abpiskunov

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/ToolCommandLineHelper.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/ToolCommandLineHelper.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using Microsoft.DotNet.Cli.Utils;
 using Microsoft.DotNet.ProjectModel;
 using NuGet.Frameworks;
+using System.Linq;
 
 namespace Microsoft.VisualStudio.Web.CodeGeneration.Tools
 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/ToolCommandLineHelper.cs
```diff
@@ -6,6 +6,7 @@
 using Microsoft.DotNet.Cli.Utils;
 using Microsoft.DotNet.ProjectModel;
 using NuGet.Frameworks;
+using System.Linq;
```
abpiskunov: needed?

src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/ToolCommandLineHelper.cs
```diff
@@ -76,5 +79,7 @@ internal static string[] GetProjectDependencyCommandArgs(string[] args, string f
             cmdArgs.AddRange(args);
             return cmdArgs.ToArray();
         }
+
+        internal static bool IsNoBuild(string[] args) => args.Contains(NO_BUILD, StringComparer.OrdinalIgnoreCase);
```
abpiskunov: Unit tests?

abpiskunov:
:shipit:

PR #278, opened 2016-07-26
Fix #262 and Add E2E tests

### LINKED ISSUE #262: Navigation properties are not displayed in List (Index), Details, or Delete
For generated views, navigation properties are displayed in Create and Edit, but not List (Index), Details, or Delete.

Sample models:
```
    public class Director
    {
        public int DirectorID { get; set; }
    }
    public class Movie
    {
        public int ID { get; set; }
        public string Title { get; set; }
        public DateTime ReleaseDate { get; set; }
        public string Genre { get; set; }
        public decimal Price { get; set; }

        public int DirectorID { get; set; }
        public virtual Director Director { get; set; }
    }
```

Also, is there any way to attribute either model so that the scaffold knew which field(s) from Director to display? Right now Create and Edit default to:
```
ViewData["DirectorID"] = new SelectList(_context.Set<Director>(), "DirectorID", "DirectorID");
```
When I'd rather have it generate:
```
ViewData["DirectorID"] = new SelectList(_context.Set<Director>(), "DirectorID", "Name");
```

prafullbhosale:
Changes: 
- Add the navigations properties to the List, Details and Delete Views. Fixes #262 
- Modify the controller template to add Includes for Foreign Key relationships. 
- Add E2E tests for MvcControllers. 

ToDo: 
- Add E2E tests for API Controllers and Views. 

cc @abpiskunov

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Templates/ViewGenerator/List.cshtml
+++ b/src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Templates/ViewGenerator/List.cshtml
@@ -41,16 +41,22 @@
 @:<table class="table">
     @:<thead>
         @:<tr>
-     
+
         IEnumerable<PropertyMetadata> properties = Model.ModelMetadata.Properties;
         foreach (var property in properties)
         {
             if (property.Scaffold && !property.IsPrimaryKey && !property.IsForeignKey)
             {
-            <th>
-                @@Html.DisplayNameFor(model => model.@GetValueExpression(property))
-            </th>
+                <th>
+                    @@Html.DisplayNameFor(model => model.@GetValueExpression(property))
+                </th>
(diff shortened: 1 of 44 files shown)


src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/Templates/ViewGenerator/List.cshtml
```diff
@@ -112,6 +123,12 @@
         return property.PropertyName;
     }
 
+    string GetValueExpression(NavigationMetadata navigation)
+    {
+        //Todo: Get the association for the property and use that.
```
abpiskunov: >//Todo: Get the association for the property and use that [](start = 8, length = 57)

Is it still todo?

abpiskunov:
:shipit:

Issue #283, opened 2016-07-28
Keep model order, not alphabetical 


Rick-Anderson:
Scaffolder puts fields in alpha order in the views – you would never want that, you’d want the same order you defined your properties in.  You get things like First Name, Hire Date, Last Name – you’d never write a model class that way.

cc @tdykstra

prafullbhosale:
Scaffolding retrieves the property information from `IEntityType` which returns the properties in alphabetical order. 

MVC5 scaffolding used `System.Web.DynamicData.MetaModel` and `System.Web.DynamicData.MetaTable` to get the properties in declared order. 

@divega do you know if there is any equivalent API that I can use that will keep the declared order of the properties?

divega:
My recommendation would be to merge the EF metadata with the original reflection metadata so that you can obtain the position of members from the latter.

TL;DR: In EF Core metadata we made the intentional choice to adopt alphabetical order because it is deterministic and stable across different platforms. In EF6 and before we had had many tricky issues because we tried to preserve the reflection order, which has never been deemed deterministic and can vary depending on the platform, version of the JIT compiler, especially when there is inheritance involved. 

Of course, the fact that properties are sorted alphabetically in EF metadata doesn't mean that users don't want the declared order (or its approximation, reflection order) to be preserved. In fact, 
we have https://github.com/aspnet/EntityFramework/issues/2272 in the EF Core backlog to preserve the declared order of properties when we produce a database table schema from the entity class (e.g. in Migrations) and the same could be applied to reverse engineering.

PR #290, opened 2016-08-22
Bug fixes

prafullbhosale:
- Newly generated DbContext will now be generated in Data folder instead
of Models Folder.
- The order of properties is now reconciled and sorted based on the
reflection order, instead of assuming the order from EF.
This may still end up being non deterministic, but this seems to be a good
solution till aspnet/EntityFramework#2272 is resolved. Addresses #283 

cc @abpiskunov

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/ModelMetadata.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/ModelMetadata.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Reflection;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Metadata;
+using System.Collections.Generic;
 
 namespace Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore
 {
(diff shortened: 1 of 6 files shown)


src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/ModelMetadata.cs
```diff
@@ -47,15 +48,50 @@ public PropertyMetadata[] Properties
...
+            {
+                throw new ArgumentNullException(nameof(entityType));
+            }
+
+            var properties = new Dictionary<string, PropertyMetadata>();
+            var entityProperties = entityType.GetProperties()
+                .Where(p => !p.IsShadowProperty)
+                .Select(p => new PropertyMetadata(p, DbContexType));
+
+            var sortedProperties = new PropertyMetadata[entityProperties.Count()];
```
abpiskunov: >var sort [](start = 12, length = 8)

move this line below next foreach , around line 80

abpiskunov:
:shipit:

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/ModelMetadata.cs
```diff
@@ -6,6 +6,7 @@
 using System.Reflection;
 using Microsoft.EntityFrameworkCore;
```
abpiskunov: >osoft [](start = 10, length = 5)

sort usings

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/ModelMetadata.cs
```diff
@@ -47,15 +48,50 @@ public PropertyMetadata[] Properties
...
+                if(properties.ContainsKey(r.Name))
+                {
+                    sortedProperties[i++] = properties[r.Name];
+                }
+            }
+
+            return sortedProperties;
+        }
+
+
```
abpiskunov: Extra line

PR #292, opened 2016-08-23
Surface proper exception when DbContext instance creation fails rather than attempting to create an instance using DI. 

### LINKED ISSUE #291: Show better error than 'No parameterless constructor defined for the object' when DbContext instance creation fails
Currently during scaffolding, if DbContext instance creation using the application startup code fails, scaffolding falls back to creating an instance using ActivatorUtilities.
This will always fail since the DbContext doesn't have a default constructor. 

So it seems like the best way to handle it would be to fail when we cannot create an instance using the app startup code rather than throwing an error that is not very useful.

prafullbhosale:
Fixes #291

cc @abpiskunov

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
@@ -439,25 +439,7 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore
                 throw new ArgumentNullException(nameof(modelType));
             }
 
-            DbContext dbContextInstance;
-            try
-            {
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
```diff
@@ -508,10 +490,11 @@ private DbContext TryCreateContextUsingAppCode(Type dbContextType, ModelType sta
             {
                 while (ex != null)
```
abpiskunov: I would create an extension method for Exception to do that, so it could be reused

abpiskunov: i mean unwrapping

abpiskunov:
:shipit

PR #296, opened 2016-08-31
Make ValidationUtil public. 

### LINKED ISSUE #94: Make ValidationUtil class public
Right now `Microsoft.Framework.CodeGenerators.Mvc.ValidationUtil` is internal, but the work it is doing when validating model types is useful outside the `Microsoft.Framework.CodeGenerators.Mvc` assembly. It should be public.

prafullbhosale:
Fixes #94

cc @abpiskunov


abpiskunov:
:shipit:

PR #297, opened 2016-09-05
Increase .travis.yml consistency between repos

dougbu:
- aspnet/Universe#349
- minimize `dotnet` setup time; no need for caching
- install openssl in OSX builds


dougbu:
a8ef07e

PR #307, opened 2016-10-13
Create Roslyn workspace with project references.

prafullbhosale:
### An overview of execution steps: 
- `dotnet-aspnet-codegenerator` tool evaluates the project information
  - It creates an `IProjectContext` for the current project. 
  - It also then creates an `IProjectContext` for each of the projectReferences for the project.
  - All of this is packaged into `ProjectInformation`.

- `dotnet-aspnet-codegenerator` invokes the _InsideMan_ (dotnet-aspnet-codegenerator-design) with a port number to callback.
- The _InsideMan_ connects to the specified port and gets the `ProjectInformation` from the caller.
  - With the information regarding the root project and project references, the _InsideMan_ builds a Roslyn Workspace. 

The approach seems very expensive right now, but don't see a good alternative to avoid creation of so many projectContexts.
One optimization would be to expose an API in the `Extensions.ProjectModel.MsBuild.Sources` that provides all available TFMs in a project without actually building the `ProjectContext`

### Additional Changes:
- Removed `imports` from the `Microsoft.VisualStudio.CodeGeneration.Tools` package.
- Added new packages to `NuGetPackageVerifier.json`

cc @mlorbetske @natemcmaster @NTaylorMullen


NuGetPackageVerifier.json
```diff
@@ -6,13 +6,20 @@
         "packages": {
             "Microsoft.VisualStudio.Web.CodeGeneration": { },
             "Microsoft.VisualStudio.Web.CodeGeneration.Core": { },
+            "Microsoft.VisualStudio.Web.CodeGeneration.Design": { },
             "Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore": { },
             "Microsoft.VisualStudio.Web.CodeGeneration.Templating": { },
             "Microsoft.VisualStudio.Web.CodeGeneration.Tools": { },
```
natemcmaster: Are planning to add 'packageType = DotnetCliTool' soon? cref https://github.com/aspnet/Coherence-Signed/issues/367 and https://github.com/aspnet/DotNetTools/blob/f8a1a66ab7962c2030528d68aabb94a41c5ca3e8/NuGetPackageVerifier.json#L7-L14

src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/Program.cs
```diff
@@ -157,70 +156,21 @@ private static void Execute(string[] args, bool isNoBuild, ILogger logger)
...
-                var frameworksInProject = TargetFrameworkFinder.GetAvailableTargetFrameworks(projectFileFinder.ProjectFilePath);
-
-                var nearestFramework = GetSuitableFrameworkFromProject(frameworksInProject);
-
-                return new MsBuildProjectContextBuilder()
-                    .AsDesignTimeBuild()
-                    .WithConfiguration("Debug")
-                    .WithProjectFile(projectFileFinder.ProjectFilePath)
-                    .WithTargetFramework(nearestFramework)
+                return new Internal.MsBuildProjectInformationBuilder(projectFileFinder.ProjectFilePath)
```
natemcmaster: Nit: add `using Microsoft.VisualStudio.Web.CodeGeneration.Tools.Internal` and remove this unnecessary qualifier.

natemcmaster:
Fundamental design question:

> the InsideMan builds a Roslyn Workspace.

Could this be done instead by the "outside man"? My concern is that putting roslyn in the inside man may introduce a dependency on Roslyn into the user's project. The design goal of the "inside man" should be to minimize its depedencies, as really it's only purpose is to load/read the user's assembly.

src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/ProjectInformation/DotNetProjectInformationBuilder.cs
```diff
@@ -0,0 +1,58 @@
...
+                throw new Exception("Failed to get Project Context Information.");
+            }
+
+            var rootContext = new DotNetProjectContext(context, "Debug", "");
+            return rootContext;
+        }
+
+    }
+
+
```
natemcmaster: Nit: remove empty lines here

src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/ProjectInformation/MsBuildProjectInformationBuilder.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        // TODO: Need a way to figure out the TargetFrameworks available in the project.
+        // Add APIs to Extensions.ProjectModel.MsBuild.Sources?
+        private MsBuildProjectContext BuildProjectContext(string projectPath)
+        {
+            var contexts =  new MsBuildProjectContextBuilder()
+                .AsDesignTimeBuild()
+                .WithConfiguration("Debug")
+                .WithProjectFile(projectPath)
+                .BuildAllTargetFrameworks();
```
natemcmaster: Call `.ToList()` to avoid multiple enumeration here.

test/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore.Test/EntityFrameworkServicesTests.cs
```diff
@@ -70,54 +70,61 @@ private EntityFrameworkServices GetEfServices(string path, string applicationNam
...
         }
 
-        private IProjectContext GetProjectContext(string path, bool isMsBuild)
+        private ProjectInformation GetProjectInformation(string path, bool isMsBuild)
         {
             if (isMsBuild)
             {
-                return new MsBuildProjectContextBuilder()
+                // TODO needs to be fixed to get all dependent Projects as well.
+                return new ProjectInformation(new MsBuildProjectContextBuilder()
```
natemcmaster: Optional nit: code style a little hard to read. Suggest making a local variable.

NuGetPackageVerifier.json
```diff
@@ -6,13 +6,20 @@
...
             "Microsoft.VisualStudio.Web.CodeGeneration.Utils": { },
             "Microsoft.VisualStudio.Web.CodeGenerators.Mvc": { }
         }
     },
+    "adx-nonshipping": {
+        "rules": [],
+        "packages": {
+            "Microsoft.VisualStudio.Web.CodeGeneration.Sources": { }
```
prafullbhosale: Nothing outside of scaffolding.

natemcmaster: In that case, I recommend not publishing as a package. It's easier to use a shared source folder, something like...
```json
"buildOptions": {
     "compile": {
         "include": "../Shared/*.cs"
    }
}
```

prafullbhosale:
> Could this be done instead by the "outside man"?

Inside man is where all the code generation/ manipulation happens currently. And the code generation relies heavily on Roslyn right now.

natemcmaster:
But is there some reason it **has to** be in the inside man? IIUC the only reason for it is that the tool needs to Assembly.Load to find existing types.

Current approach:

outside man --> push project info to --> inside man

This restricts scaffolding to runnable projects (e.g. you can't scaffold in class library projects)

What I'm suggesting:

outside man <--  push list of user types <-- inside man

This has the benefit of
(1) minimal dependencies required to be added to user project
(2)  you could use AssemblyLoadContext to scaffold class library projects
(3) the tool can do simple scaffolding operations without invoking the inside man (scenario: projects not using EF, operations that don't need much information, etc.)

prafullbhosale:
>This restricts scaffolding to runnable projects (e.g. you can't scaffold in class library projects)

This is currently intentional. 
> IIUC the only reason for it is that the tool needs to Assembly.Load to find existing types.

Another reason to make the inside man be the one to do the code generation is the way the current extensibility works for scaffolding. 
The inside man looks at the user's dependencies to figure out if any custom scaffolder is installed and needs to be invoked. 
Although this can be changed. We'd need to come up with a new contract for these custom scaffolders.

natemcmaster:
Can code generators return something to the roslyn workspace? IIUC "GenerateCode" is always a 'void' return type. I don't think my suggestion would break this model.

prafullbhosale:
Yes, that might work. I will look into prototyping the new approach. 
However for now, the workspace needs to live in the Inside man, because of another reason. 
The Roslyn packages need `imports` and we cannot add that to tools.

prafullbhosale:
:up: :date:

natemcmaster:
Code looks find, although I still think we should move roslyn into the outer guy.  :shipit: on this as the bigger design question is kinda beyond the scope of this PR.

PR #312, opened 2016-10-20
Redesign Creating Project Context for MsBuild projects

prafullbhosale:
Created a custom MsBuild `Task` and target which are carried by the `Microsoft.VisualStudio.CodeGeneration.Design` package. 
The `Task` evaluates the project and writes info to a json file. 
The _Tool_ invokes the task in a separate process, and then reads the json file to retrieve the project context. 

Ideally, the `Task` and the Targets file should be carried by the `Microsoft.VisualStudio.CodeGeneration.Tools` package, but a `DotNetCliToolReference` cannot add targets to the project, so have to use the _InsideMan_ package as the delivery mechanism for the task. 

cc @mlorbetske @natemcmaster @NTaylorMullen


src/Microsoft.VisualStudio.Web.CodeGeneration.Msbuild/ProjectContextWriter.cs
```diff
@@ -0,0 +1,198 @@
...
+using System.Linq;
+using Microsoft.Build.Framework;
+using Microsoft.Extensions.ProjectModel;
+using Microsoft.Extensions.ProjectModel.Resolution;
+using Microsoft.VisualStudio.Web.CodeGeneration.Utils;
+using Newtonsoft.Json;
+
+namespace Microsoft.VisualStudio.Web.CodeGeneration.Msbuild
+{
+    public class ProjectContextWriter : Build.Utilities.Task
```
natemcmaster: I totally get why we have to do this -- dropping project.json has made this necessary-- but it seems like the complexity of this is grown so much that I think the question "Does this feature make sense as a CLI tool?" is worth considering. If scaffolding were a VS extension instead, it could lean on VS to provide this information instead of needing to add a custom task inside the user's project.

prafullbhosale: I agree. It is a very round about way of getting the information. We are thinking of moving this in the VS extension direction as you suggested, but that will have to wait for the time being at least.

prafullbhosale:
:up: :Date:

PR #316, opened 2016-10-31
Remove usage of build3 for msbuild projects.

prafullbhosale:
- Now that CLI has merged the *3 verbs, no longer need to use `build3` for building msbuild projects.
- Remove MsBuild and NuGet dependency overrides from .Tools package.

Note: The same changes need to flow to feature/msbuild-1.1.0 branch as well. 

@natemcmaster @NTaylorMullen

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/DotNetBuildCommandHelper.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/DotNetBuildCommandHelper.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.IO;
 using Microsoft.DotNet.Cli.Utils;
@@ -16,7 +17,8 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration.Tools
             NuGetFramework framework,
             string buildBasePath)
         {
-            const string buildCommandName = "build";
+            var buildCommandName = "build";
+
             var args = new List<string>()
             {
                 project,
@@ -37,6 +39,13 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration.Tools
 
             var stdOutMsgs = new List<string>();
             var stdErrorMsgs = new List<string>();
+
+            // HACK :: Till we upgrade the repo to use msbuild based CLI, use *3 verbs for the tests.
+            if (string.Equals(System.Environment.GetEnvironmentVariable("USE_3_VERBS"), "1", StringComparison.Ordinal)
+                && string.Equals(Path.GetExtension(project), ".csproj", StringComparison.OrdinalIgnoreCase))
+            {
+                buildCommandName = "build3";
+            }
 
             var command = Command.CreateDotNet(
                     buildCommandName,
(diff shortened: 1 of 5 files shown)


natemcmaster:
Do you have a build of cli? I haven't seen successful build of CLI with the change yet.

prafullbhosale:
v3972 of CLI is the latest successful I could find for win-x64 and it has these changes. 
The CLI github page isn't updated yet though.

NTaylorMullen:
Does the new CLI that removes the 3 verbage still understand project.json? Aka does it switch based on the project file its handed? I ask because we're still relying on the hybrid build behavior of it.

prafullbhosale:
Nope. It does not.

NTaylorMullen:
> Nope. It does not.

I'd recommend not swapping over to it then. We can't run tests or accurately migrate all of our projects over to csproj quite yet.

natemcmaster:
Mixing PJ and msbuild is messy.  I'm working on porting aspnet/DotNetTools to build/test with MSBuild so I can use just one version of CLI instead of trying to mix a PJ version with an MSBuild version.

NTaylorMullen:
> Mixing PJ and msbuild is messy. I'm working on porting aspnet/DotNetTools to build/test with MSBuild so I can use just one version of CLI instead of trying to mix a PJ version with an MSBuild version.

Spent several days doing this with RazorTooling and failed with cli 3892. Lots of things were broken. Some examples: source packages/resources. If you're able to do it with the latest CLI I might give it another try. As of right now though I'm reluctant to try again 😄

prafullbhosale:
I tried with scaffolding and it took me a couple of days to work around issues to migrate the source code to msbuild, and still couldn't get everything working yet. 
Still WIP for now.

natemcmaster:
I have build/pack/test mostly working for the dotnettools repo, but I probably don't have as many requirements as your projects.

prafullbhosale:
@NTaylorMullen and I spoke with @Eilon and decided to add a environment variable for the CI to be able to build and test using an older CLI.

src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/DotNetBuildCommandHelper.cs
```diff
@@ -37,10 +39,12 @@ internal class DotNetBuildCommandHelper
             var stdOutMsgs = new List<string>();
             var stdErrorMsgs = new List<string>();
 
-            //TODO Remove this when CLI merges build3 to build.
-            var buildCommandName = Path.GetExtension(project).Equals(".csproj")
-                ? "build3"
-                : "build";
+            // HACK :: Till we upgrade the repo to use msbuild based CLI, use *3 verbs for the tests.
+            if (System.Environment.GetEnvironmentVariable("USE_3_VERBS").Equals("1")
```
NTaylorMullen: This'll explode when you publish and use outside of our CI 😄 .

`string.Equals(Environment.GetEnvironmentVariable("USE_3_VERBS"), "1", StringComparer.Ordinal)`.

I prefer to always use `string.Equals` instead of `.Equals` because it works well with `null`.

src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/DotNetBuildCommandHelper.cs
```diff
@@ -37,10 +39,12 @@ internal class DotNetBuildCommandHelper
...
+                && Path.GetExtension(project).Equals(".csproj"))
```
NTaylorMullen: string.Equals

src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/DotNetBuildCommandHelper.cs
```diff
@@ -16,6 +16,8 @@ internal class DotNetBuildCommandHelper
             NuGetFramework framework,
             string buildBasePath)
         {
+            string buildCommandName = "build";
```
NTaylorMullen: var

prafullbhosale:
:up: :date:

NTaylorMullen:
:shipit:

Eilon:
Nice 😄

PR #318, opened 2016-11-02
Load project dependencies by name instead of path in EFServices.

prafullbhosale:
Trying to fix the test issue on CI, where E2E tests fail while trying to load project dependencies. 

cc @natemcmaster


natemcmaster:
:shipit:

I believe you probably hit of the breaking changes introduce in .NET Core's assembly loader. cc @davidfowl

PR #327, opened 2016-11-04
feature/msbuild: pin to stable versions

natemcmaster:
cc @prafullbhosale 

:watch: on a final build version of Microsoft.DotNet.Cli.Utils


prafullbhosale:
Looks good so far. As you said, lets wait till we get the cli version.

PR #328, opened 2016-11-04
Bump version to preview4

prafullbhosale:
This is LTS version with Project.json support. 

cc @natemcmaster @pranavkm @NTaylorMullen

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.Utils/project.json
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.Utils/project.json
@@ -6,21 +6,16 @@
     ],
     "xmlDoc": true
   },
-  "version": "1.0.0-preview2-update1",
+  "version": "1.0.0-preview4-final",
   "description": "Contains utilities used by ASP.NET Core Code Generation packages.",
   "dependencies": {
     "Microsoft.CodeAnalysis.CSharp.Workspaces": "1.3.0",
     "Microsoft.DotNet.ProjectModel": "1.0.0-rc3-003121",
-    "Microsoft.Extensions.PlatformAbstractions": "1.0.0"
+    "Microsoft.Extensions.PlatformAbstractions": "1.0.0",
+    "NETStandard.Library": "1.6.0"
   },
   "frameworks": {
-    "net451": {
-      "frameworkAssemblies": {
-        "System.Threading.Tasks": {
-          "type": "build"
-        }
-      }
-    },
+    "net451": { },
     "netstandard1.6": {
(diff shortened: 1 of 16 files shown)


NTaylorMullen:
Why the code changes?

prafullbhosale:
The ProjectJsonWorkspaces code was duplicated in this repo due to a restore issue long long time back. 
The code is since outdated and has some bug fixes in that. 
So removed the duplicated code and took a dependency on the ProjectModel.Workspaces package.

NTaylorMullen:
Will it work without the code changes? 

There were some bug fixes done by RazorTooling that Damian suggest we not take to ensure no bugs. Slightly scary to have code changes when the goal is to just re-ship old bits with new versions.

NTaylorMullen:
Also note that the bug fixes done by Razor were super minor which is why it never resulted in an update.

prafullbhosale:
Yes, it would work without the code changes. Will revert the code changes and update this.

prafullbhosale:
:up: :date:

src/Microsoft.VisualStudio.Web.CodeGeneration.Utils/project.json
```diff
@@ -6,21 +6,16 @@
...
     "Microsoft.CodeAnalysis.CSharp.Workspaces": "1.3.0",
     "Microsoft.DotNet.ProjectModel": "1.0.0-rc3-003121",
-    "Microsoft.Extensions.PlatformAbstractions": "1.0.0"
+    "Microsoft.Extensions.PlatformAbstractions": "1.0.0",
+    "NETStandard.Library": "1.6.0"
   },
   "frameworks": {
-    "net451": {
-      "frameworkAssemblies": {
-        "System.Threading.Tasks": {
```
NTaylorMullen: This is definitely weird that it was here before howeeevvver, same question as the code changes before but with the dependency changes here being needed.

NTaylorMullen:
Last question is this: https://github.com/aspnet/Scaffolding/pull/328#discussion_r86494754. If they're not needed / don't solve any significant issue I'd suggest undoing the change; otherwise, if it can't work properly without then keep em in.

Other then that, :shipit:

src/Microsoft.VisualStudio.Web.CodeGeneration.Utils/project.json
prafullbhosale: Keeping this here and not adding NETStandard.Library 1.6.0 was causing build failure on my local about a missing System.* assembly reference for net451

PR #340, opened 2016-11-16
Use EF Design APIs to get DbContext instance from user code.

### LINKED ISSUE #339: Scaffolding Controller with EF broken for projects targeting net4x
On a project targeting net4x run 
`dotnet aspnet-codegenerator -p . controller -name CarController -m WebApplication1.Models.Car -dc WebApplication1.Data.ApplicationDbContext`

It fails with error

```
'Could not load file or assembly 'WebApplication1.exe' or one of its dependencies. The system cannot find the file specified.
   at Microsoft.VisualStudio.Web.CodeGeneration.ActionInvoker.<BuildCommandLine>b__6_0()
   at Microsoft.Extensions.CommandLineUtils.CommandLineApplication.Execute(String[] args)
   at Microsoft.VisualStudio.Web.CodeGeneration.ActionInvoker.Execute(String[] args)
   at Microsoft.VisualStudio.Web.CodeGeneration.CodeGenCommand.Execute(String[] args)'

```

cc @vkichline

prafullbhosale:
Instead of trying to host user's app to get the DbContext, while scaffolding, let EF run its magic.

Fixes #339 
cc @mlorbetske, @bricelam

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
@@ -10,10 +10,11 @@ using System.Linq;
 using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
-using Microsoft.AspNetCore.Hosting;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Design;
+using Microsoft.EntityFrameworkCore.Design.Internal;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.Extensions.ProjectModel;
 using Microsoft.VisualStudio.Web.CodeGeneration.DotNet;
@@ -119,6 +120,8 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore
             var dbContextSymbols = _modelTypesLocator.GetType(dbContextFullTypeName).ToList();
             var startupType = _modelTypesLocator.GetType("Startup").FirstOrDefault();
             Type modelReflectionType = null;
+            ReflectedTypesProvider reflectedTypesProvider = null;
+            string dbContextError = string.Empty;
             if (dbContextSymbols.Count == 0)
             {
                 await ValidateEFSqlServerDependency();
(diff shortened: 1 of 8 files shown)


src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
```diff
@@ -468,23 +390,25 @@ private ModelMetadata GetModelMetadata(Type dbContextType, Type modelType, Model
...
-            try {
-                var builder = new WebHostBuilder();
-                builder.UseKestrel()
-                        .UseContentRoot(Directory.GetCurrentDirectory());
+            try
+            {
+                // Use EF design APIs to get the DBContext instance.
+                var operationHandler = new OperationReportHandler();
+                var operationReporter = new OperationReporter(operationHandler);
+                var startupInvoker = new StartupInvoker(
```
bricelam: Hmm, there is a higher-level method that might be better: `DbContextOperations.CreateContext()`. This would account for `DbContext` types who's design-time creation has been overridden using `IDbContextFactory`.

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
```diff
@@ -14,6 +14,8 @@
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Design;
+using Microsoft.EntityFrameworkCore.Design.Internal;
```
bricelam: @divega How do you feel about this calling internal API? The scenario is being able to load a model in-process at design time. When we create a model visualizer, we'll either need to do the exact same thing, or create a serialization format for models.

bricelam: If we want to avoid calling internals, we could add a public facade to do this.

prafullbhosale:
:up: :date: to use `DbContextOperations` instead of `StartupInvoker`

bricelam:
LGTM, but we should close on whether Scaffolding should use internal APIs or if we need to add a public API on our side.

mlorbetske:
:shipit:

prafullbhosale:
The API doesn't seem to be working on net45x projects. 
It throws an error
```
No parameterless constructor was found on 'CarContext'. Either add a parameterless constructor to 'CarContext' or add an implementation of 'IDbContextFactory<CarContext>' in the same assembly as 'CarContext'. StackTrace:
   at Microsoft.EntityFrameworkCore.Design.Internal.DbContextOperations.<>c__DisplayClass12_3.<FindContextTypes>b__13()
   at Microsoft.EntityFrameworkCore.Design.Internal.DbContextOperations.CreateContext(Func`1 factory)
   at Microsoft.EntityFrameworkCore.Design.Internal.DbContextOperations.CreateContext(String contextType)
```

@bricelam Any ideas?

prafullbhosale:
Figured out the problem. When the `DbContextOperations` api tries to configure services, it ends up calling 
`services.AddMvc()`, which tries to load the assembly. 

Matching the name of the assembly to the one on disk for `net451` fixes the issue.

PR #345, opened 2016-11-23
Throw error from EF as is if finding entityType fails.

### LINKED ISSUE #343: Improve error message for the case where the model passed in for scaffolding is not a valid EF model
trying to scaffold using the below model

```cs
    public class SimpleModel
    {
        public string Foo { get; set; }
        public bool Bar { get; set; }
    }
```

throws and error 

`There is no entity type SimpleModel on DbContext` which doesn't help the user correct the problem.

The problem with the model is that it is not a valid EF model, since it is missing a property that is denoted (or detected) as a `Key`

Changing the model as below makes scaffolding work: 

```cs
    public class SimpleModel
    {
        public int ID { get; set; }
        public string Foo { get; set; }
        public bool Bar { get; set; }
    }
```

Or

```cs
    public class SimpleModel
    {
        [System.ComponentModel.DataAnnotations.Key]
        public int SimpleModelKey { get; set; }
        public string Foo { get; set; }
        public bool Bar { get; set; }
    }
```

prafullbhosale:
Fixes #343 

cc @mlorbetske


mlorbetske:
:shipit:

PR #367, opened 2016-12-22
Use nameof expression in controller RedirectToAction calls

scottaddie:
Use the C# 6 `nameof` expression in the MVC controller action method `RedirectToAction` calls. This makes the scaffolded controllers consistent with the `RedirectToAction` calls used in the project templates (in the aspnet/Templates repo).


scottaddie:
Will open another PR for this on the `feature/msbuild` branch.

PR #374, opened 2017-01-04
Add UserSecretsAttribute to in memory assembly

### LINKED ISSUE #373: Scaffolding a controller with entityframework fails if AddUserSecrets<Startup> method is used
If the Startup.cs has 

```cs
 builder.AddUserSecrets<Startup>();
```

scaffolding a controller with entityframework fails with the error 
`No parameterless constructor was found for the DbContext`

prafullbhosale:
Fixes #373 

cc @natemcmaster @mlorbetske

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/AssemblyAttributeGenerator.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/AssemblyAttributeGenerator.cs
@@ -1,0 +1,54 @@
+using System;
+using System.Reflection;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore
+{
+    /// <summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/AssemblyAttributeGenerator.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// <returns></returns>
+        internal SyntaxTree GenerateAttributeSyntaxTree()
+        {
+            if (_originalAssembly.CustomAttributes.Any())
+            {
+                foreach (var attr in _originalAssembly.CustomAttributes)
+                {
+                    if (attr.AttributeType.FullName == "Microsoft.Extensions.Configuration.UserSecrets.UserSecretsIdAttribute")
+                    {
+                        var attributeText = string.Format(attributeTextTemplate, Guid.NewGuid());
```
natemcmaster: You'll probably need to use the actual value of the attribute. This may affect how EF finds connection strings and settings. `UserSecretsIdAttribute.UserSecretsId`.

prafullbhosale:
:up: :date:

PR #376, opened 2017-01-09
Fix dispatcher's check for netcoreapp to ignore version of netcore app

prafullbhosale:
Fix the DotNetDispatcher to check only the framework name and not the version for `netcoreapp`

cc @pranavkm 

_This fix needs to flow to 1.1.0-msbuild3 branch as well_


src/Shared/DotNetDispatcher/DotnetToolDispatcher.cs
```diff
@@ -79,6 +79,15 @@ internal static class DotnetToolDispatcher
             return command;
         }
 
+        private static bool IsNetCoreAppFramework(NuGetFramework framework)
+        {
+            //Only need to compare the framework name to be netcoreapp. Version doesn't matter.
```
pranavkm: `// Only` (space after the comment

Issue #282, opened 2016-07-28
[DatabaseGenerated(DatabaseGeneratedOption.None)] should scaffold


Rick-Anderson:
In ASP.NET 4 the scaffolder is intelligent enough to know that if you define a key as `[DatabaseGenerated(DatabaseGeneratedOption.None)]` it should scaffold a field in the Create view – now it doesn’t.

cc @tdykstra

mlorbetske:
TFS Bug#248116

prafullbhosale:
This is fixed in `dev`

PR #378, opened 2017-01-13
Fix logic to determine if a property is AutoGenerated.

prafullbhosale:
Addresses #282 

cc @mlorbetske


src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/PropertyMetadata.cs
```diff
@@ -62,7 +62,10 @@ public PropertyMetadata(IProperty property, Type dbContextType)
 
             IsEnum = property.ClrType.GetTypeInfo().IsEnum;
             IsReadOnly = property.IsReadOnlyBeforeSave;
-            IsAutoGenerated = property.IsStoreGeneratedAlways;
+
+            IsAutoGenerated = ((property.RequiresValueGenerator
+                || property.IsStoreGeneratedAlways)
+                && (property.ValueGenerated != ValueGenerated.Never));
```
ajcvickers: How is `IsAutoGenerated` used? Values can be generated when entities start being tracked, and values can be generated/changed when an entity is saved to the database. However, in neither case is a value necessarily generated--usually if a value has already been set by the application, then it will be used instead of generating a new value.

prafullbhosale: It is used to determine whether the field should be shown on the Create/ Edit pages.

prafullbhosale: > Values can be generated when entities start being tracked, and values can be generated/changed when an entity is saved to the database. However, in neither case is a value necessarily generated

@ajcvickers So based on the `IProperty` what's the best way to determine whether a property's value can be set by the application or not? More specifically while creating entities.

ajcvickers: @prafullbhosale It depends... In EF Core, the general idea is that if an application specifies a value (i.e. sets the property to something other than the CLR default for the type), then that value will be pushed to the database regardless of whether the property is configured to use store generated values. A common use for this is when people want to insert specific seed data into the database for testing, but use generated values on normal running of the application. So at the highest level, any property _can_ be set by the application. 

If we look at this from a different perspective and ask, "can EF generate a value for this property?" then the answer is yes if:
- The property is marked as ValueGenerated.OnAdd or ValueGenerated.OnAddOrUpdate
- Or the property has a ValueGenerator explicitly set

It seems to me that using this is a reasonable heuristic for scaffolding, but it could result in situations where people want to set the value (e.g. for seeding/testing) but can't.

Three wrinkles to keep in mind with this:
- Values for foreign key properties will be populated by EF if a graph of entities including the entity where the PK is defined is being inserted together. If a single entity is being inserted, then FK values will likely need to be set by the application.
- Sometimes database columns are set up in such a way that explicit values cannot be inserted. For example, SQL Server Identity columns are like this by default. EF does not try to detect this and so if an application does try to specify a value for such a column, then an exception will be thrown by the database.
- IsStoreGeneratedAlways switches off the mechanism whereby EF will try to insert/update values that have been set by the application. However, this flag is only commonly used for concurrency tokens where it is coupled with StoreGenerated.OnAddOrUpdate. In the context of this conversation it can be ignored.

prafullbhosale: Thanks @ajcvickers! 

@Andrew-MSFT, @sayedihashimi, @mlorbetske based on the above, looks like we would need to decide on an approach for scaffolding to determine when to not scaffold fields that may be auto-generated. 

In my opinion, having a default behavior to not scaffold fields that follow below pattern should be ok: 
- Require a value generator and 
- Do not have a `[DatabaseGenerated(DatabaseGeneratedOption.None)]`

The users can always go edit the views generated by scaffolding in order to add input to the fields that they want to be set by the application. 

Thoughts?

prafullbhosale: Spoke to @sayedihashimi offline and we decided to go ahead with the current solution. 
For the users who want to let the application set the specific values, they can always edit the generated content. 

@ajcvickers unless there are any other concerns, I would like to merge this.

prafullbhosale:
Merging this now as we want to create release branches. 
Will update if there is anymore feedback.

PR #400, opened 2017-02-10
Update msbuild version

prafullbhosale:
cc @natemcmaster 
This same change will need to happen in `rel/1.1.0` branch


natemcmaster:
And for the next few weeks, you have my blessing to push future version bumps without review as the MSBuild RTM build number revs. I don't want our release to get hung waiting on a PR review.

PR #401, opened 2017-02-11
1.0.4 patch train updates

JunTaoLuo:
Do we need to update these packages for 1.0.4? @Eilon


prafullbhosale:
@Eilon Should I assume we will need similar updates for the `MSBuild` based tools as well for the 1.0.4 patch?

JunTaoLuo:
No longer need to build this in 1.0.4

PR #405, opened 2017-02-16
Updating to 1.1.1 patch versions

prafullbhosale:


JunTaoLuo:
Have you updated all the projects? What about https://github.com/aspnet/Scaffolding/blob/rel/1.1.0/src/Microsoft.VisualStudio.Web.CodeGeneration.Core/Microsoft.VisualStudio.Web.CodeGeneration.Core.csproj?

prafullbhosale:
> https://github.com/aspnet/Scaffolding/blob/rel/1.1.0/src/Microsoft.VisualStudio.Web.CodeGeneration.Core/Microsoft.VisualStudio.Web.CodeGeneration.Core.csproj?

Don't see what needs to change here. 
https://github.com/aspnet/Coherence-Patch/blob/rel/1.1.2/patchConfig.json
says the version of CommandLineUtils and DependencyInjection stays at 1.1.0. 

Am I missing something?

PR #408, opened 2017-02-23
Stop producing Microsoft.VisualStudio.Web.CodeGeneration.Msbuild

natemcmaster:
These binaries are shipped in the .Tools package. We don't need the .MSbuild package.


natemcmaster:
:bell:

PR #435, opened 2017-03-29
Mark EFCore.Design as private packageref.

### LINKED ISSUE #434: Enable users to scaffold non-entityframework controllers and views without needing EFCore dependency in the project.
Currently the scaffolding package `Microsoft.VisualStudio.Web.CodeGeneration.Design` has a dependency on `Microsoft.EntityFrameworkCore.Design` package. 
This causes issues for users who do not want to have `Microsoft.EntityFrameworkCore.*` dependencies in their project, but would still like to scaffold views and controllers.

prafullbhosale:
- Refactor model based view scaffolding to split ef and non ef based
views generation.
- Show error message if EF dependencies are not present and ef based
scaffolding is invoked.

Fixes #434


src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/CodeModelService.cs
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        public Task<ContextProcessingResult> GetModelMetadata(ModelType modelType)
+        {
+            if (modelType == null)
+            {
+                throw new ArgumentNullException(nameof(modelType));
+            }
+
+            var projectCompilation = _workspace.CurrentSolution.Projects
```
mlorbetske: Is there a reason not to await here?

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/ICodeModelService.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    public interface ICodeModelService
+    {
+        /// <summary>
+        /// Gets the metadata for a given model without using the Datacontext.
+        /// Without the datacontext, the metadata will consist of no Navigations/ Primary keys.
+        /// </summary>
+        /// <param name="modelType">Model type for the EF metadata has to be returned.</param>
+        /// <returns>Returns <see cref="ContextProcessingResult"/>.</returns>
+        Task<ContextProcessingResult> GetModelMetadata(ModelType modelType);
```
mlorbetske: Async suffix?

src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/MessageStrings.resx
```diff
@@ -144,6 +144,9 @@
   <data name="ForceOptionDesc" xml:space="preserve">
     <value>Specify whether to overwrite existing files.</value>
   </data>
+  <data name="InstallEfPackages" xml:space="preserve">
+    <value>To scaffold controllers and views using models please install Entity Framework core packages and try again: {0}</value>
```
mlorbetske: Should we indicate the specific package that's required?

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/ICodeModelService.cs
prafullbhosale: It'd probably be a good idea to change all method names to follow the convention of using Async suffix, but currently the methods in the repo don't do that. Will create a separate PR for that.

src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/MessageStrings.resx
prafullbhosale: We need two packages to be installed, `Microsoft.EntityFrameworkCore.Design` and `Microsoft.EntityFrameworkCore.SqlServer` 

The message displayed to the user will have both these ids. 
Are you suggesting, we only show the one that is missing?

PR #462, opened 2017-04-27
Remove additional feeds from NuGet.config

### LINKED ISSUE #446: Trim feeds from NuGet.config
We shouldn't need all the additional feeds in the NuGet.config. Additional components should be mirrored so any consumer of the package doesn't have to additionally list feeds.

prafullbhosale:
Fixes #446

cc @Eilon

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -2,11 +2,10 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="dotnet-cli" value="https://dotnet.myget.org/F/dotnet-cli/api/v3/index.json" />
-    <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
     <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-release/api/v3/index.json" />
     <add key="AspNetCoreTools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" /> 
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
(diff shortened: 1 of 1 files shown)


NuGet.config
```diff
@@ -2,11 +2,10 @@
...
   <packageSources>
     <clear />
-    <add key="dotnet-cli" value="https://dotnet.myget.org/F/dotnet-cli/api/v3/index.json" />
-    <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
     <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-release/api/v3/index.json" />
     <add key="AspNetCoreTools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" /> 
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
-    <add key="AspNetCore-Patch" value="https://dotnet.myget.org/F/aspnet-1-0-4-patch/api/v3/index.json" />
+    <!-- Todo Remove this after https://github.com/aspnet/CoreCLR/issues/180 is closed. -->
+    <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
```
prafullbhosale: This is needed for `Microsoft.Build.Runtime` package used in 
`Microsoft.VisualStudio.Web.CodeGeneration.MsBuild`  and `Microsoft.VisualStudio.Web.CodeGeneration.Tools`

Eilon: Ok I think that's fine. At least, I think this is certainly an improvement over the current state.

PR #469, opened 2017-04-29
[2.0.0-preview1] Re-add a reference to System.Reflection.TypeExtensions

bricelam:
It was removed from Microsoft.EntityFrameworkCore.Design.


bricelam:
Verified locally. Merging to fix the build.

PR #470, opened 2017-05-01
Enable E2E tests on CI.

prafullbhosale:


test/Shared/MsBuildProjectSetupHelper.cs
```diff
@@ -37,12 +37,32 @@ public void SetupProjects(TemporaryFileProvider fileProvider, ITestOutputHelper
...
+            RestoreAndBuild(Path.Combine(fileProvider.Root, "Root"), output);
+        }
+
+        private void RestoreAndBuild(string path, ITestOutputHelper output)
+        {
 
             var result = Command.CreateDotNet("restore",
-                new[] { Path.Combine(fileProvider.Root, "Root", "Test.csproj") })
+                new string[] {})
+                .InWorkingDirectory(path)
```
natemcmaster: You might also want to set the env var`DOTNET_SKIP_FIRST_TIME_EXPERIENCE=true`. In dotnet-watch, functional tests that call restore ended up breaking because the package cache warmup is too slow or fails.

PR #483, opened 2017-05-10
Fix EF based scaffolding

### LINKED ISSUE #481: Scaffolding should set ASPNETCORE_ENVIRONMENT=Development
The `DbContextOperations` API now ignores the `environment` parameter passed to it and expects 
`ASPNETCORE_ENVIRONMENT` variable to be set. 

https://github.com/aspnet/Scaffolding/blob/rel/2.0.0-preview1/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs#L448

cc @mlorbetske

prafullbhosale:
Fixes #481

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
@@ -441,12 +441,12 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore
                 // Use EF design APIs to get the DBContext instance.
                 var operationHandler = new OperationReportHandler();
                 var operationReporter = new OperationReporter(operationHandler);
+                // EF infers the environment (Development/ Production) based on the environment variable
+                // ASPNETCORE_ENVIRONMENT. This should already be set up by the CodeGeneration.Design process.
                 var dbContextOperations = new DbContextOperations(
                     operationReporter,
                     dbContextType.GetTypeInfo().Assembly,
-                    startupType.GetTypeInfo().Assembly,
-                    "Development",
-                    Path.GetDirectoryName(_projectContext.ProjectFullPath));
+                    startupType.GetTypeInfo().Assembly);
 
                 var dbContextService = dbContextOperations.CreateContext(dbContextType.FullName);
 
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/Templates/DbContext/NewDbContextFactory.cshtml
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/Templates/DbContext/NewDbContextFactory.cshtml
@@ -21,6 +21,6 @@ namespace @Model.DbContextNamespace
     public class @factoryClassName : IDbContextFactory<@Model.DbContextTypeName>
     {
         public @Model.DbContextTypeName Create(string[] args) =>
-            @(Model.ProgramTypeName).BuildWebHost(args).Services.GetRequiredService<@Model.DbContextTypeName>();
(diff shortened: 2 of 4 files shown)


src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
```diff
@@ -441,12 +441,12 @@ private DbContext TryCreateContextUsingAppCode(Type dbContextType, Type startupT
...
                 var operationHandler = new OperationReportHandler();
                 var operationReporter = new OperationReporter(operationHandler);
+                // EF infers the environment (Development/ Production) based on the environment variable
+                // ASPNETCORE_ENVIRONMENT. This should already be set up by the CodeGeneration.Design process.
                 var dbContextOperations = new DbContextOperations(
                     operationReporter,
                     dbContextType.GetTypeInfo().Assembly,
-                    startupType.GetTypeInfo().Assembly,
-                    "Development",
-                    Path.GetDirectoryName(_projectContext.ProjectFullPath));
```
mlorbetske: The startup type and path to project are no longer required?

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/Templates/DbContext/NewDbContextFactory.cshtml
```diff
@@ -21,6 +21,6 @@ namespace @Model.DbContextNamespace
     public class @factoryClassName : IDbContextFactory<@Model.DbContextTypeName>
     {
         public @Model.DbContextTypeName Create(string[] args) =>
-            @(Model.ProgramTypeName).BuildWebHost(args).Services.GetRequiredService<@Model.DbContextTypeName>();
+            @(Model.ProgramTypeName).BuildWebHost(args).Services.CreateScope().ServiceProvider.GetRequiredService<@Model.DbContextTypeName>();
```
mlorbetske: I thought that these were not going to be required going forward

bricelam: We also need to sort out if we even need an IDbContextFactory implementation.

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
bricelam: We want to expose a public API for this in the 2.0.0 timeframe, so we isolate you from additional breaks.

prafullbhosale: @mlorbetske, the startupType's assembly is still needed. 
The `environment` and `contentRootPath` parameters are no longer needed. 
See https://github.com/aspnet/EntityFramework/issues/7050#issuecomment-297833793 for reference.

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/Templates/DbContext/NewDbContextFactory.cshtml
prafullbhosale: Created https://github.com/aspnet/Scaffolding/issues/485. 
Will need to keep this for now so the tests don't break.

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/EntityFrameworkServices.cs
mlorbetske: Ah, missed the add for startup type, only saw the delete, sorry

PR #488, opened 2017-05-13
Unbreak failing tests

rynowak:
The razor engine had a feature that had built in default `using`s for
System and `System.Threading.Tasks`. We're removing that API. You can
either define your own global `using`s and other directives like I did
in this workaround, or just add the necessary namespaces where they are
used.

I also fixed the console output for tests. I'm not sure if was
intentionally not working, but this makes the tests much easier to
diagnose.


rynowak:
FYI @prafullbhosale

PR #495, opened 2017-05-23
React to aspnet/EntityFramework/pull/8534

prafullbhosale:
cc @mlorbetske

### DIFF
--- a/test/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore.Test/TestModels/TestModel.cs
+++ b/test/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore.Test/TestModels/TestModel.cs
@@ -4,6 +4,8 @@
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal;
+using Microsoft.EntityFrameworkCore.Storage;
+using Microsoft.EntityFrameworkCore.Storage.Internal;
 
 namespace Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore.Test.TestModels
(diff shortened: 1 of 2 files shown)


NuGet.config
```diff
@@ -2,7 +2,7 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
+    <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-volatile-dev/api/v3/index.json" />
```
prafullbhosale: Will need to change this back to CI-Dev once a coherence build passes and new packages are published

ryanbrandenburg: No, it's fine to change this locally to make sure that your changes catch everything but we shouldn't check these changes in. Scaffolding will be broken until Coherence passes but that shouldn't take long.

test/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore.Test/TestModels/TestModel.cs
```diff
@@ -13,7 +15,8 @@ public static IModel CategoryProductModel
         {
             get
             {
-                var builder = new ModelBuilder(new CoreConventionSetBuilder().CreateConventionSet());
+                var sqlServerTypeMapper = new SqlServerTypeMapper(new RelationalTypeMapperDependencies());
```
ryanbrandenburg: I assume these couldn't be easily mocked and that's why we needed to add that new package?

anpete: Can probably avoid by using CoreTypeMapper here instead of SqlServerTypeMapper.

PR #500, opened 2017-05-24
Bump target framework version to net461

prafullbhosale:
Need to react to https://github.com/aspnet/Microsoft.Data.Sqlite/commit/69bbe2215963a6ec7d4885af9a88fcd341008c76

cc @mlorbetske, @Eilon


pranavkm:
Merging to unbreak the build. @prafullbhosale - I think we want to do another pass and update TFMs to `netstandard2.0` and stop cross-compiling when possible.

PR #502, opened 2017-05-24
React to replace obsolete EF apis

prafullbhosale:
cc @pranavkm, @mlorbetske 

Going to merge this immediately to unblock CI. 
Will address PR comments later.


src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore.csproj
```diff
@@ -22,6 +22,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.Design" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
     <PackageReference Include="System.Diagnostics.Contracts" Version="$(CoreFxVersion)" />
     <PackageReference Include="System.Reflection.TypeExtensions" Version="$(CoreFxVersion)" />
+    <PackageReference Include="NETStandard.Library.NETFramework" Version="2.0.0-preview2-*" />
```
pranavkm: 👍 Let's fix up the repo to remove the cross-compilation and change this to only be used in desktop.

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/IPropertyExtensions.cs
```diff
@@ -28,12 +28,12 @@ internal static IPropertyMetadata ToPropertyMetadata(this IProperty property, Ty
             propertyMetadata.IsForeignKey = property.IsForeignKey();
 
             propertyMetadata.IsEnum = property.ClrType.GetTypeInfo().IsEnum;
-            propertyMetadata.IsReadOnly = property.IsReadOnlyBeforeSave;
+            propertyMetadata.IsReadOnly = property.BeforeSaveBehavior != PropertyValueBehavior.UseValue;
```
ajcvickers: What is IsReadOnly used for? EF will never set this by default for any property and it would be very uncommon for app code to set it. (Same is true for IsReadOnlyBeforeSave.)

That being said, this is the correct change to preserve the behavior from before.

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/IPropertyExtensions.cs
```diff
@@ -28,12 +28,12 @@ internal static IPropertyMetadata ToPropertyMetadata(this IProperty property, Ty
...
 
             //TODO: RequiresValueGenerator() is an internal EF Core method.
             // Need to confirm with EF if this is ok to use. 
             propertyMetadata.IsAutoGenerated = ((property.RequiresValueGenerator()
-                    || property.IsStoreGeneratedAlways)
+                    || property.BeforeSaveBehavior != PropertyValueBehavior.UseValue)
                 && (property.ValueGenerated != ValueGenerated.Never));
```
ajcvickers: This should be:
```C#
propertyMetadata.IsAutoGenerated = 
    property.ValueGenerated != ValueGenerated.Never
    || property.GetValueGeneratorFactory() != null;
```

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/IPropertyExtensions.cs
prafullbhosale: `Readonly` is used to determine whether to create an editable field for the property or not

src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/IPropertyExtensions.cs
prafullbhosale: Fixed in https://github.com/aspnet/Scaffolding/commit/c2847addc9561213f3e59a43c77fc48bcad9da82

PR #509, opened 2017-06-15
Pin package versions during test project creation.

prafullbhosale:
Porting the change from dev branch. 

cc @mlorbetske

### DIFF
--- a/test/Shared/MsBuildProjectSetupHelper.cs
+++ b/test/Shared/MsBuildProjectSetupHelper.cs
@@ -3,7 +3,10 @@
 
 using System;
 using System.IO;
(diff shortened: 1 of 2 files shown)


Eilon:
This change might be correct for preview2, but it is not necessarily correct in the general case. Each repo has its own version and different repos might not have the same version. For example, in patch releases MVC, logging, BrowserLink, StaticFiles, etc. can all have different versions.

prafullbhosale:
Need to think more on how to handle this in the general case where AspNetCore packages can be a range of versions. 

I will close this PR and create a new one with the minimal change to remove the deprecated packages from the test project.

PR #512, opened 2017-06-16
Pb/fix razor templating

prafullbhosale:
React to changes in Razor template engine. 

Note: Had to add `<NoWarn>NU1605</NoWarn>` to the generated test projects. 
For some weird reason, the test project ends up with an implicit reference of `Microsoft.NetCore.App` with a lower version than what the `Microsoft.VisualStudio.Web.CodeGeneration.Design` gets. 

Filing an issue to remove the `NoWarn`

cc @mlorbetske @ryanbrandenburg

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.Templating/Microsoft.VisualStudio.Web.CodeGeneration.Templating.csproj
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.Templating/Microsoft.VisualStudio.Web.CodeGeneration.Templating.csproj
@@ -15,6 +15,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Razor.Language" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Razor.Extensions" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="$(CodeAnalysisVersion)" />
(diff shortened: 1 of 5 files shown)


test/Shared/MsBuildProjectSetupHelper.cs
```diff
@@ -73,7 +73,7 @@ public void SetupProjects(TemporaryFileProvider fileProvider, ITestOutputHelper
 
         private void RestoreAndBuild(string path, ITestOutputHelper output)
         {
-
+
```
NTaylorMullen: nuke

test/Shared/MsBuildProjectStrings.cs
```diff
@@ -7,13 +7,15 @@ internal class MsBuildProjectStrings
     {
         public static string GetNugetConfigTxt(string artifactsDir)
         {
+            // TODO: Remove dotnet-core feed.
```
NTaylorMullen: Should just file an issue instead of having todos

src/Microsoft.VisualStudio.Web.CodeGeneration.Templating/Microsoft.VisualStudio.Web.CodeGeneration.Templating.csproj
```diff
@@ -15,6 +15,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Razor.Language" Version="$(AspNetCoreVersion)" />
```
NTaylorMullen: Can remove this package reference.

PR #521, opened 2017-06-20
Update Roslyn workspaces to 2.3.0-beta3

prafullbhosale:
Removes the need for having AssetTargetFallback

cc @Eilon @mlorbetske

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -11,7 +11,7 @@
   </PropertyGroup>
   <!-- Partner Versions-->
   <PropertyGroup>
-    <CodeAnalysisVersion>2.3.0-beta3-61731-06</CodeAnalysisVersion>
+    <CodeAnalysisVersion>2.3.0-beta3-61814-09</CodeAnalysisVersion>
     <JsonNetVersion>10.0.1</JsonNetVersion>
     <MicrosoftBuildVersion>15.3.0-preview-000388-01</MicrosoftBuildVersion>
     <NuGetFrameworksVersion>4.0.0-*</NuGetFrameworksVersion>
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -11,7 +11,7 @@
   </PropertyGroup>
   <!-- Partner Versions-->
   <PropertyGroup>
-    <CodeAnalysisVersion>2.3.0-beta1</CodeAnalysisVersion>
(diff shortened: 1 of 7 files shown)


NuGet.config
```diff
@@ -5,5 +5,6 @@
     <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-release/api/v3/index.json" />
     <add key="AspNetCoreTools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <add key="Roslyn" value="https://dotnet.myget.org/F/roslyn/api/v3/index.json" />
```
natemcmaster: We should add this to our mirror build instead of adding a new feed. cc @kichalla

Eilon: Yes, we definitely need to do that because otherwise we risk the build breaking in the future.

prafullbhosale:
Removed the additional feed. Need to wait for the package to be mirrored before merging this PR.

Eilon:
cc @ryanbrandenburg @mikeharder

build/dependencies.props
```diff
@@ -11,7 +11,7 @@
   </PropertyGroup>
   <!-- Partner Versions-->
   <PropertyGroup>
-    <CodeAnalysisVersion>2.3.0-beta1</CodeAnalysisVersion>
+    <CodeAnalysisVersion>2.3.0-beta3-61731-06</CodeAnalysisVersion>
```
Eilon: BTW do we have confirmation from the Roslyn team that this is the exact correct build of their packages to use for the Preview 2 release?

prafullbhosale: I was going with the assumption that since Razor.Workspaces is using this build, it is the one we should use. But will confirm with the Roslyn folks.

NTaylorMullen: This version does not ship on NuGet. Us in Razor use this version for our VisualStudio only components because that's what's shipped in VS.

Lets talk more with @rynowak I know he had concerns about Scaffolding's use of this package 😄

prafullbhosale: Confirmed with Roslyn and CLI folks that the right version here is `2.3.0-beta3-61814-09`

prafullbhosale:
Looks like we are not taking this for 2.0.0-preview2. 
Also, the package version of Roslyn in this PR is not correct. Need to wait for a public package of Roslyn that has the fix.

PR #524, opened 2017-06-26
Fix issue with string Ids

### LINKED ISSUE #511: Scaffolding incorrectly scaffolds IdentityUser
Given the following class:

```C#
public class Jaszzzzs : IdentityUser
{
    public bool Awesome { get; set; }
}
```

If you try and scaffold controllers with views using EF for this model you get:
image

/cc @jbagga

prafullbhosale:
Fixes #511

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.Templating/TypeUtilities.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.Templating/TypeUtilities.cs
@@ -13,7 +13,6 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration.Templating
         private static bool IsNonIntegerPrimitive(Type type)
         {
             type = Nullable.GetUnderlyingType(type) ?? type;
-
             return (type == typeof(bool))
                    || (type == typeof(byte[]))
                    || (type == typeof(DateTime))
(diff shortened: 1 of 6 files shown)


src/Microsoft.VisualStudio.Web.CodeGeneration.Templating/TypeUtilities.cs
```diff
@@ -22,7 +21,6 @@ private static bool IsNonIntegerPrimitive(Type type)
                    || (type == typeof(double))
                    || (type == typeof(float))
                    || (type == typeof(Guid))
-                   || (type == typeof(string))
                    || (type == typeof(TimeSpan))
```
seancpeters: It's not clear to me why string would be removed from this check, I'm missing some context.

prafullbhosale: This check is used to determine if its a nullable type or not. 
Without this change it would generate a controller method like 
```cs
public async Task<IActionResult> Details(string? id)
```
which would fail to compile.

prafullbhosale: Looks like I can't really remove this from here as this is going to break `CodeModelMetadata`.

prafullbhosale:
Updated. 
Exposed a new method `IsNullable(Type type)`  for the check.

src/Microsoft.VisualStudio.Web.CodeGeneration.Templating/TypeUtilities.cs
```diff
@@ -37,7 +36,18 @@ private static bool IsInteger(Type type)
                    || (type == typeof(ulong))
                    || (type == typeof(ushort))
                    || (type == typeof(sbyte))
-                   || (type == typeof(char));    
+                   || (type == typeof(char));
+        }
+
+        public static bool IsNullable(Type t)
```
seancpeters: 👍

PR #529, opened 2017-06-29
Use Nuget.config from repo root for E2E tests.

### LINKED ISSUE #527: Use NuGet.config from solution folder for E2E tests
From @ryanbrandenburg:

> Prafull Bhosale do you happen to know how scaffolding functional tests work with Universe? Universe creates an artifacts directory, and edits all the top level NuGet.config's to contain it
> But because you're creating a NuGet.config in memory only https://github.com/aspnet/Scaffolding/blob/c65dfcc4be2e76a16c364de4e264b0fc0676f413/test/Shared/MsBuildProjectSetupHelper.cs#L60 here it's not working for me locally
>Obviously SOMETHING makes these pass on the CI

On the CI for universe builds, the test projects end up restoring an published version of the packages from aspnetcore-ci-dev.

prafullbhosale:
Fixes #527

### DIFF
--- a/test/Shared/MsBuildProjectStrings.cs
+++ b/test/Shared/MsBuildProjectStrings.cs
@@ -1,21 +1,30 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.IO;
+using System.Xml;
+
 namespace Microsoft.VisualStudio.Web.CodeGeneration
 {
     internal class MsBuildProjectStrings
     {
-        public static string GetNugetConfigTxt(string artifactsDir)
+        public static string GetNugetConfigTxt(string artifactsDir, string nugetConfigPath)
         {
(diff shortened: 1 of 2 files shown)


test/Shared/MsBuildProjectStrings.cs
```diff
@@ -1,21 +1,28 @@
...
+            var xmlDoc = new XmlDocument();
+            xmlDoc.Load(nugetConfigPath);
+
+            var root = xmlDoc.DocumentElement;
+            var packageSources = root.SelectSingleNode("packageSources");
+            var newSource = xmlDoc.CreateElement("add");
+            newSource.SetAttribute("key", "local");
+            newSource.SetAttribute("value", artifactsDir);
+            packageSources.AppendChild(newSource);
+            StringWriter sw = new StringWriter();
```
ryanbrandenburg: It might not make a difference, but I'm a bit obsessive about putting anything that implements IDisposable in a `using` block.

PR #531, opened 2017-06-29
Remove AssetTargetFallback

ryanbrandenburg:


prafullbhosale:
Is this no longer needed? 
I thought we'd need to update to a version of Roslyn that depends on System.Composition instead of Microsoft.Composition. (2.3.0-beta3)

PR #536, opened 2017-07-04
Use latest runtime version in test projects

mikeharder:
* Without setting `RuntimeFrameworkVersion` to `2.0.0-*`, projects default to using the runtime version included in the CLI, which may not match the runtime version required by the current build of ASP.NET.  This causes test failures when there are breaking changes between runtime versions.


mikeharder:
Addressed by https://github.com/aspnet/Scaffolding/pull/538

PR #538, opened 2017-07-04
Use the same runtime version as the one used to run test

pranavkm:


pranavkm:
Feedback?

mikeharder:
Should the property be added to the net461 project?

test/Shared/MsBuildProjectStrings.cs
```diff
@@ -115,6 +117,7 @@ public static string GetNugetConfigTxt(string artifactsDir, string nugetConfigPa
     <RuntimeIdentifier>win7-x64</RuntimeIdentifier>
     <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
     <NoWarn>NU1605</NoWarn>
+    <RuntimeFrameworkVersion>{2}</RuntimeFrameworkVersion>
```
mikeharder: Should this be added to the net461 project?

pranavkm: It shouldn't matter. The value is only used in apps targeting `.NETCoreApp` (https://github.com/dotnet/sdk/blob/feca75bd43f6374715621024d930164de8bde67c/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.Sdk.DefaultItems.props#L48). In fact, for a while that's how it was listed in our dependencies.props: https://github.com/aspnet/Mvc/blob/bd6520cc501b9f8de4e71f5b29fb542677ddfaea/build/dependencies.props#L12

PR #543, opened 2017-07-06
Add explicit dependency to prevent conflict

ryanbrandenburg:
This is to prevent the `warning MSB3277`'s we've been seeing in Universe.


mikeharder:
@ryanbrandenburg: Could this have been fixed by supressing the warning instead?  This seems cleaner and less impactful than adding an explicit dependency.

ryanbrandenburg:
I tried adding MSB3277 to NoWarn in these projects and it didn't seem to suppress, is there another way to go about that or are we out of luck?

mikeharder:
I believe the correct fix is https://github.com/aspnet/Scaffolding/pull/562, and this commit should be reverted once https://github.com/aspnet/Scaffolding/pull/562 is merged.

PR #574, opened 2017-07-21
Add required files for the views and razor pages if they don't exist

prafullbhosale:
For View scaffolding, if the `Views` folder doesn't exist in the project yet, scaffolding will now add a `_ViewImports.cshtml` (and optionally `_ViewStart.cshtml` and `_Layout.cshtml`) files to the project.

Similarly for Razor Page scaffolding as well additional files will be scaffolded.

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/RazorPage/RazorPageScaffolderBase.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/RazorPage/RazorPageScaffolderBase.cs
@@ -70,6 +70,8 @@ namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Razor
             }
         }
 
+        protected bool IsRazorPageWireUpNeeded { get; set; }
+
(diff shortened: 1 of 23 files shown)


prafullbhosale:
cc @Eilon, @DamianEdwards, @mlorbetske

src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/RazorPage/RazorPageScaffolderBase.cs
```diff
@@ -140,6 +176,33 @@ protected RazorPageGeneratorTemplateModel GetRazorPageViewGeneratorTemplateModel
             return templateModel;
         }
 
+        protected bool RazorPagesFolderExists(string relativeFolderPath, string applicationBasePath)
```
seancpeters: This method is identical to to ViewScaffolderBase.ViewsFolderExists except for the value of the `const string viewsFolderName`. I'd suggest moving the method to CommonGeneratorBase and passing in the viewsFolderName as another param.

PR #578, opened 2017-07-24
Fixed logic for which Razor scaffolder gets invoked.

seancpeters:
When generating all pages, the templateName is "All". Initial code had danticipated it being null.

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.Design/Program.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.Design/Program.cs
@@ -43,7 +43,13 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration.Design
             var noBuild = app.Option("--no-build", "", CommandOptionType.NoValue);
             var simMode = app.Option("--simulation-mode", Resources.SimulationModeOptionDesc, CommandOptionType.NoValue);
 
-
+#if DEBUG
+            if (args.Contains("--debug", StringComparer.OrdinalIgnoreCase))
+            {
+                Console.WriteLine($"Attach a debugger to processID: {System.Diagnostics.Process.GetCurrentProcess().Id} and hit enter.");
+                Console.ReadKey();
+            }
+#endif
             app.OnExecute(async () =>
             {
                 CodeGenerationEnvironmentHelper.SetupEnvironment();
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.Design/Program.cs
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.Design/Program.cs
@@ -50,6 +50,7 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration.Design
                 Console.ReadKey();
             }
 #endif
+
             app.OnExecute(async () =>
             {
                 CodeGenerationEnvironmentHelper.SetupEnvironment();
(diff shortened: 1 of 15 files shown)


src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/RazorPage/RazorPageGenerator.cs
```diff
@@ -51,16 +51,18 @@ public class RazorPageGenerator : CommonGeneratorBase, ICodeGenerator
 
             if (string.IsNullOrEmpty(razorPageGeneratorModel.ViewName))
             {
-                // when this happens, we want to generate all the views
+                throw new ArgumentException(MessageStrings.ViewNameRequired);
```
prafullbhosale: This argument will still be null for the `All` template.

prafullbhosale: Looks like this change is compatible with the `2.0.0` packages, so it should be fine. 
Just need to move the null check for `ViewName` to after the `TemplateName` check and then 
:shipit:

src/Microsoft.VisualStudio.Web.CodeGeneration.Design/Program.cs
```diff
@@ -43,13 +43,22 @@ private static void Execute(string[] args, ConsoleLogger logger)
             var noBuild = app.Option("--no-build", "", CommandOptionType.NoValue);
             var simMode = app.Option("--simulation-mode", Resources.SimulationModeOptionDesc, CommandOptionType.NoValue);
 
-#if DEBUG
-            if (args.Contains("--debug", StringComparer.OrdinalIgnoreCase))
+// the for reals version
+//
+//#if DEBUG
+//            if (args.Contains("--debug", StringComparer.OrdinalIgnoreCase))
```
prafullbhosale: Looks like we need to evolve this condition to support debugging when testing using VS. 
Maybe along with the parameter, look for an environment variable?

src/Microsoft.VisualStudio.Web.CodeGeneration.Design/Program.cs
```diff
@@ -43,13 +43,22 @@ private static void Execute(string[] args, ConsoleLogger logger)
...
+//            {
+//                Console.WriteLine($"Attach a debugger to processID: {System.Diagnostics.Process.GetCurrentProcess().Id} and hit enter.");
+//                Console.ReadKey();
+//            }
+//#endif
+//
+
+// for testing via VS
+            while (!System.Diagnostics.Debugger.IsAttached)
```
prafullbhosale: Remove debug code.

src/Microsoft.VisualStudio.Web.CodeGeneration.Design/Program.cs
seancpeters: Yeah, not sure what exactly is the right way... I can experiment.

src/Microsoft.VisualStudio.Web.CodeGeneration.Design/Program.cs
seancpeters: Whoops, yep, will do now.

src/Microsoft.VisualStudio.Web.CodeGenerators.Mvc/RazorPage/EmptyRazorPageScaffolder.cs
```diff
@@ -35,7 +35,6 @@ public class EmptyRazorPageScaffolder : RazorPageScaffolderBase
 
             if (string.IsNullOrEmpty(razorGeneratorModel.ViewName))
             {
-                // TODO: make a separate message resource string for this (currently setup using the one for VIEW)
                 throw new ArgumentException(MessageStrings.ViewNameRequired);
```
prafullbhosale: We should change these to `Razor Page Name is required.`

PR #581, opened 2017-07-27
Switch to using DbContextActivator

### LINKED ISSUE #549: Stop using DbContextOperations.CreateContext
In EF Core 2.0.0, we've added [DbContextActivator](https://github.com/aspnet/EntityFramework/blob/rel/2.0.0/src/EFCore.Design/Design/DbContextActivator.cs) which provides non-internal API to do this.

prafullbhosale:
Fixes #549 
cc @mlorbetske


bricelam:
:tada:

PR #583, opened 2017-07-28
Shorten path names

ryanbrandenburg:
This should unblock Coherence-Signed and follow the new naming conventions we're moving towards.


Scaffolding.sln
```diff
@@ -31,23 +31,23 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		README.md = README.md
 	EndProjectSection
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.VisualStudio.Web.CodeGenerators.Mvc", "src\Microsoft.VisualStudio.Web.CodeGenerators.Mvc\Microsoft.VisualStudio.Web.CodeGenerators.Mvc.csproj", "{3E99E653-6B31-45CF-9B3A-30EEE497BE13}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.VisualStudio.Web.CodeGenerators.Mvc", "src\VS.Web.CG.Mvc\VS.Web.CG.Mvc.csproj", "{3E99E653-6B31-45CF-9B3A-30EEE497BE13}"
```
pranavkm: Maybe drop the `VS.Web` prefix too?

Scaffolding.sln
```diff
@@ -77,7 +77,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ClassLibrary1", "test\TestA
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.VisualStudio.Web.CodeGeneration.MSBuild.Test", "test\Microsoft.VisualStudio.Web.CodeGeneration.MSBuild.Test\Microsoft.VisualStudio.Web.CodeGeneration.MSBuild.Test.csproj", "{6C1398FF-A8C1-45A9-9796-6A60B2B011E1}"
```
pranavkm: Missed this?

Scaffolding.sln
ryanbrandenburg: I figured it would make it easier to know what these were for to have VS in the folder name at least, but if we think it matters I can trim that too.

Scaffolding.sln
ryanbrandenburg: I did not shorten any of the tests because I didn't think they would affect the Coherence-Signed thing one way or another, but I can change them too for consistency.

pranavkm: Yeah, it would be nice to have a consistent format. We could do this in a follow up since it's not particularly necessary at this point.

PR #596, opened 2017-08-16
Show help when no args are passed

### LINKED ISSUE #588: Invoking dotnet aspnet-codegenerator with no args should show help

prafullbhosale:
Fixes #588 

cc @mlorbetske @Eilon

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/Resources.resx
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/Resources.resx
@@ -174,4 +174,10 @@
   <data name="TargetFrameworkOptionDesc" xml:space="preserve">
     <value>Target Framework to use. (Short folder name of the tfm. eg. net46)</value>
   </data>
+  <data name="EnableTracingMessage" xml:space="preserve">
+    <value>To see more information, enable tracing by setting environment variable 'codegen_trace' = 1.</value>
+  </data>
+  <data name="GenericErrorMessage" xml:space="preserve">
+    <value>Something went wrong.</value>
+  </data>
 </root>
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/Resources.resx
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/Resources.resx
@@ -178,7 +178,7 @@
     <value>To see more information, enable tracing by setting environment variable 'codegen_trace' = 1.</value>
   </data>
   <data name="GenericErrorMessage" xml:space="preserve">
-    <value>Something went wrong.</value>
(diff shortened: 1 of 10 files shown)


src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/Resources.resx
```diff
@@ -174,4 +174,13 @@
...
     <value>Target Framework to use. (Short folder name of the tfm. eg. net46)</value>
   </data>
+  <data name="EnableTracingMessage" xml:space="preserve">
+    <value>To see more information, enable tracing by setting environment variable 'codegen_trace' = 1.</value>
+  </data>
+  <data name="GenericErrorMessage" xml:space="preserve">
+    <value>Scaffolding failed.</value>
+  </data>
+  <data name="NoCodeGeneratorsFound" xml:space="preserve">
+    <value>No codegenerators are available in this project. Please add 'Microsoft.VisualStudio.Web.CodeGeneration.Design' NuGet package to the project.</value>
```
mlorbetske: Nit: "code generators" maybe?

Eilon: Also: `Please add **the** 'foobar' ...`

Eilon:
This seems like a pretty large change (200+ lines) for something that on the surface sounds quite straightforward. Is there no simpler way to handle this case?

prafullbhosale:
The large part of the change is wrapping the entry points in a try-catch block (which has caused indentation changes).

Eilon:
Is there no way to just detect that there are no args and show the help info?

Eilon:
Also, even without the white space changes, it's still 150 lines...

prafullbhosale:
There is a way to just detect that with smaller number of changes. 
However, that doesn't handle the case when the command line looks something like this: 

```
dotnet aspnet-codegenerator -p pathToProject
```

This currently causes dotnet.exe to crash as exceptions are thrown and not handled gracefully. 

So yes, this PR does more than advertised earlier:
1. Show help when no args are passed.
2. Handle exceptions gracefully without crashing dotnet.exe

Eilon:
I see, so this change is really about "do better handling of all cases of invalid/missing/whatever command line args"?

prafullbhosale:
yes, that is correct.

Eilon:
Ok, it's up to you and @mlorbetske to figure out the best change. In general for patches we try to address just the affected issue or else we risk regressions, so do whatever you feel is best.

prafullbhosale:
Discussed this with @mlorbetske. We decided to take this as its a relatively safe change and provides a better user experience by not letting dotnet.exe crash with exceptions due to invalid/ missing command line arguments.

PR #602, opened 2017-08-21
Build dotnet-aspnet-codegeneration.design for AnyCpu

### LINKED ISSUE #601: net461 version of dotnet-aspnet-codegenerator-design.exe should be built for multiple platforms instead of win7-x86
For a net461 project with RID as win7-x64, the dotnet-aspnet-codegenerator-design.exe causes exception as MVC tries to load the 32-bit assembly in a 64-bit process:

```
System.BadImageFormatException occurred
  HResult=0x8007000B
  Message=Could not load file or assembly 'dotnet-aspnet-codegenerator-design' or one of its dependencies. An attempt was made to load a program with an incorrect format.
  Source=mscorlib 
```

Sample project: 

```xml
<Project Sdk="Microsoft.NET.Sdk.Web">

  <PropertyGroup>
    <TargetFramework>net461</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <Folder Include="wwwroot\" />
  </ItemGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore" Version="2.0.0" />
    <PackageReference Include="Microsoft.AspNetCore.Mvc" Version="2.0.0" />
    <PackageReference Include="Microsoft.AspNetCore.Mvc.Formatters.Xml" Version="2.0.0" />
    <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="2.0.0" />
  </ItemGroup>

  <ItemGroup>
    <DotNetCliToolReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Tools" Version="2.0.0" />
  </ItemGroup>

</Project>
```

prafullbhosale:
Fixes #601 

cc @Eilon @mlorbetske

### DIFF
--- a/src/Microsoft.VisualStudio.Web.CodeGeneration.Design/Microsoft.VisualStudio.Web.CodeGeneration.Design.csproj
+++ b/src/Microsoft.VisualStudio.Web.CodeGeneration.Design/Microsoft.VisualStudio.Web.CodeGeneration.Design.csproj
@@ -9,9 +9,12 @@
     <OutputType>exe</OutputType>
     <PackageTags>aspnetcore;codegenerator;scaffolding;visualstudioweb</PackageTags>
     <PackageId>Microsoft.VisualStudio.Web.CodeGeneration.Design</PackageId>
-    <RuntimeIdentifier Condition=" '$(TargetFramework)' != 'netcoreapp2.0' ">win7-x86</RuntimeIdentifier>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(TargetFramework)' != 'netcoreapp2.0' ">
+    <PlatformTarget>AnyCPU</PlatformTarget>
+    <Prefer32Bit>true</Prefer32Bit>
+  </PropertyGroup>
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.VisualStudio.Web.CodeGenerators.Mvc\Microsoft.VisualStudio.Web.CodeGenerators.Mvc.csproj" />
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Web.CodeGeneration.Design/Microsoft.VisualStudio.Web.CodeGeneration.Design.csproj
```diff
@@ -9,9 +9,12 @@
     <OutputType>exe</OutputType>
     <PackageTags>aspnetcore;codegenerator;scaffolding;visualstudioweb</PackageTags>
     <PackageId>Microsoft.VisualStudio.Web.CodeGeneration.Design</PackageId>
-    <RuntimeIdentifier Condition=" '$(TargetFramework)' != 'netcoreapp2.0' ">win7-x86</RuntimeIdentifier>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(TargetFramework)' != 'netcoreapp2.0' ">
```
pranavkm: Does this work correctly when user's app references native binaries like Sqlite \ Libuv? We went the route of compiling the binary for x86 and x64 in precompilation but if this works, perhaps we could consider this as a solution.

prafullbhosale: Turns out this doesn't work. It solves the problem of MVC loading the assembly, but scaffolding itself fails when it tries to run the user code.

prafullbhosale:
Closing as this approach doesn't work.

PR #605, opened 2017-08-23
Add binaries built for multiple windows platforms

### LINKED ISSUE #601: net461 version of dotnet-aspnet-codegenerator-design.exe should be built for multiple platforms instead of win7-x86

prafullbhosale:
#### Changes include: 
- Add binaries for the following RIDs to the Microsoft.VisualStudio.Web.CodeGeneration.Design package
  - win7-x86
  - win7-x64
  - win-arm
  - win-arm64
- Also add a `AnyCPU` PlatformTarget binary in `lib\net461` path to handle the case when the user project specifies `<PlatformTarget>AnyCpu</PlatformTarget>` without a RID

- Change how we locate the _inside-man_ to dispatch for net461 case to handle the fact that the compilation assemblies will no longer have the _inside man_ for net461 case

#### Impact of packing 6 binaries instead of 2: 
On local machine the `2.0.0` package size is 55KB, the new package size is 127 KB

cc @Eilon, @mlorbetske 

Fixes #601


src/Shared/DotNetDispatcher/DotnetToolDispatcher.cs
```diff
@@ -71,7 +71,7 @@ internal static class DotnetToolDispatcher
             {
                 EnsureBindingRedirects(assemblyFullPath, Path.GetFileName(dependencyToolPath));
                 // For Full framework, we can directly invoke the <dependencyTool>.exe from the user's bin folder.
-                dependencyToolPath = Path.Combine(Path.GetDirectoryName(assemblyFullPath), Path.GetFileName(dependencyToolPath));
+                // dependencyToolPath = Path.Combine(Path.GetDirectoryName(assemblyFullPath), Path.GetFileName(dependencyToolPath));
```
pranavkm: Delete?

src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/Program.cs
```diff
@@ -193,6 +205,15 @@ private static void Execute(string[] args, bool isNoBuild, ILogger logger)
                     assemblyFullPath: context.AssemblyFullPath);
         }
 
+        private static bool IsNetCoreAppFramework(NuGetFramework framework)
+        {
+            //Only need to compare the framework name to be netcoreapp. Version doesn't matter.
```
pranavkm: `// Only` (spacing)

src/Microsoft.VisualStudio.Web.CodeGeneration.Tools/Program.cs
```diff
@@ -166,15 +166,27 @@ private static void Execute(string[] args, bool isNoBuild, ILogger logger)
...
-            if (string.IsNullOrEmpty(dotnetCodeGenInsideManPath))
+                if (string.IsNullOrEmpty(dotnetCodeGenInsideManPath))
+                {
+                    throw new InvalidOperationException(Resources.AddDesignPackage);
+                }
+            }
+            else
             {
-                throw new InvalidOperationException(Resources.AddDesignPackage);
+                dotnetCodeGenInsideManPath = Path.Combine(Path.GetDirectoryName(context.AssemblyFullPath), DESIGN_TOOL_NAME+".exe");
```
pranavkm: `DESIGN_TOOL_NAME + ".exe"` (spacing)

version.props
```diff
@@ -1,7 +1,7 @@
 <!-- This file may be overwritten by automation. Only values allowed here are VersionPrefix and VersionSuffix.  -->
 <Project>
     <PropertyGroup>
-        <VersionPrefix>2.0.0</VersionPrefix>
+        <VersionPrefix>2.0.1</VersionPrefix>
```
pranavkm: Is this for the patch release?

src/Microsoft.VisualStudio.Web.CodeGeneration.Design/Microsoft.VisualStudio.Web.CodeGeneration.Design.csproj
```diff
@@ -9,14 +9,35 @@
     <OutputType>exe</OutputType>
     <PackageTags>aspnetcore;codegenerator;scaffolding;visualstudioweb</PackageTags>
     <PackageId>Microsoft.VisualStudio.Web.CodeGeneration.Design</PackageId>
-    <RuntimeIdentifier Condition=" '$(TargetFramework)' != 'netcoreapp2.0' ">win7-x86</RuntimeIdentifier>
+    <RuntimeIdentifier Condition=" '$(TargetFramework)' != 'netcoreapp2.0' ">win7-x64</RuntimeIdentifier>
```
pranavkm: Looks good. As discussed, you could simplify by this by either using this project do generate rid specific exes or have a common property file so you don't repeat the same details multiple times over.

version.props
prafullbhosale: If @Eilon approves 😃

jamesabolton:
Any update on this being released?

Eilon:
@jamesabolton we don't have a precise ETA for the release, but we're setting up a temporary package feed where you'll be able to try out the pre-release builds of the patch. I'll update all the patch-related issues w/ info on the feed when it's available. We're working on getting the 1.0.x and 1.1.x patches ready first, and then the 2.0.x patch.

jamesabolton:
Thank you, @Eilon. As noted in that other bug on dotnet/cli the new projects for Core 2.0 fail with an exception due to this issue. I haven't tested all of them but the MVC and Angular ones do not work. If you remove reference to the DLL then the web UI doesn't function.

jamesabolton:
Is there anything we can do locally to get around this issue to work with the new templates and net standard 2.0 until a patch is released? I tried to downgrade the project that was created to 1.1/1.2 and that breaks various parts of the code in the template project.

ddimitrov90:
any new info about this issue? 
can't really do anything to workaround it and use scaffolding in aspnet core 2.0 in Service Fabric project :(

prafullbhosale:
> Is there anything we can do locally to get around this issue to work with the new templates and net standard 2.0 until a patch is released? I tried to downgrade the project that was created to 1.1/1.2 and that breaks various parts of the code in the template project.

The only possible workaround at the moment is to set the RuntimeIdentifier to `win7-x86` to use scaffolding. Then change it back to the x64 variant.

mshindal:
What about the other Windows RIDs, like `win10-x64`?

Eilon:
@mshindal in general the `win7-XYZ` RIDs will run on newer OSes, including the `win10-XYZ` RIDs. Do you have a specific concern?

PR #611, opened 2017-08-28
Prepare for 1.1.2 version

prafullbhosale:
cc @Eilon @mlorbetske 

Creating against the right branch this time.


prafullbhosale:
>Is this still waiting for something?

Will need to update the runtime framework version once we have the new runtime. 
Will create a separate PR for that.

PR #612, opened 2017-08-28
Prepare for 1.0.2 version

prafullbhosale:
cc @Eilon @mlorbetske 

Created against the right branch.


prafullbhosale:
> Note that you'll also need to update https://github.com/aspnet/Coherence-Patch/blob/rel/1.0.6/targets/Config.targets#L143

You mean once we create a `rel/` branch? The PR is created against the branch mentioned in the config.targets

JunTaoLuo:
since there is a version update of src packages, you'll need to create the rel/1.1.2 and rel/1.0.2 branches. After you create these branches, you'll need to update the config.targets file I linked in the PRs

PR #623, opened 2017-09-15
Replace occurrences of SingleOrDefaultAsync

### LINKED ISSUE #300: Switch to use Find[Async]() instead of First[OrDefault][Async]() or Single[OrDefault][Async]() as appropriate
`Find()` (added to EF Core in 1.1) can return objects already in memory without performing a database roundtrip. We are also unnecessarily using `SingleOrDefault()` in some cases in comparisons by key which would never return more than one row.

scottaddie:
Fixes #300 

Make use of `FindAsync` instead of `SingleOrDefaultAsync` in the scaffolding templates. Fixes the root cause of the problem identified by @ajcvickers in https://github.com/aspnet/Docs/pull/4253.

@divega Please review.

### DIFF
--- a/src/VS.Web.CG.Mvc/Templates/RazorPageGenerator/DeletePageModel.cshtml
+++ b/src/VS.Web.CG.Mvc/Templates/RazorPageGenerator/DeletePageModel.cshtml
@@ -53,7 +53,7 @@ namespace @Model.NamespaceName
                 return NotFound();
             }
 
-            @Model.ViewDataTypeShortName = await _context.@(Model.ModelTypeName)@(inlineIncludes).FindAsync(id);
+            @Model.ViewDataTypeShortName = await _context.@(Model.ModelTypeName)@(inlineIncludes).FirstOrDefaultAsync(m => m.@primaryKeyName == id);
 
             if (@Model.ViewDataTypeShortName == null)
             {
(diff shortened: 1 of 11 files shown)


src/VS.Web.CG.Mvc/Templates/RazorPageGenerator/DeletePageModel.cshtml
```diff
@@ -53,7 +53,7 @@ namespace @Model.NamespaceName
                 return NotFound();
             }
 
-            @Model.ViewDataTypeShortName = await _context.@(Model.ModelTypeName)@(inlineIncludes).SingleOrDefaultAsync(m => m.@primaryKeyName == id);
+            @Model.ViewDataTypeShortName = await _context.@(Model.ModelTypeName)@(inlineIncludes).FindAsync(id);
```
divega: Not sure what `@(inlineIncludes)` expands too. If it is calls to `Include()` then this would result in generating invalid code because `FindAsync()` is only defined on `DbSet<T>` but not on the type returned by `Include()`. For these cases in which you cannot use `FindAsync()` I would prefer `FirstOrDefaultAsync()` rather than `SingleOrDefaultAsync()` because it leads to simpler SQL.

scottaddie:
@divega Please take another look.

prafullbhosale:
Changes look good.

PR #625, opened 2017-09-19
Update versions for the patch

prafullbhosale:
cc @mlorbetske

### DIFF
--- a/test/Shared/MsBuildProjectStrings.cs
+++ b/test/Shared/MsBuildProjectStrings.cs
@@ -41,19 +41,19 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration
 
   <ItemGroup>
     <PackageReference Include=""Microsoft.ApplicationInsights.AspNetCore"" Version=""2.1.1"" />
-    <PackageReference Include=""Microsoft.AspNetCore"" Version=""{0}"" />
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore/Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore.csproj
```diff
@@ -17,7 +17,7 @@
        without adding EntityFrameworkCore dependency.
        https://github.com/aspnet/scaffolding/issues/434
        -->
-    <PackageReference Include="Microsoft.EntityFrameworkCore.Design" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.Design" Version="$(AspNetCore201PatchVersion)" PrivateAssets="All" />
```
natemcmaster: This should already be covered by https://github.com/aspnet/Scaffolding/blob/rel/2.0.1/build/dependencies.targets. It's a little hacky for now, but we want to eventually move away from having lots of variables in dependencies.props and using instead a list that has an exact `packageID => version` mapping

PR #637, opened 2017-10-05
Patch test fixes

JunTaoLuo:
Not the best fix but this will fix the tests.

### DIFF
--- a/test/Shared/MsBuildProjectSetupHelper.cs
+++ b/test/Shared/MsBuildProjectSetupHelper.cs
@@ -20,43 +20,10 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration
         #endif
 
         private static object _syncObj = new object();
-        private static string _nugetConfigText;
-        private static string NuGetConfigText
-        {
-            get
-            {
-                if (string.IsNullOrEmpty(_nugetConfigText))
-                {
-                    lock(_syncObj)
-                    {
-                        if (string.IsNullOrEmpty(_nugetConfigText))
-                        {
-                            string artifactsDir = null;
-                            string nugetConfigPath = null;
-                            var current = new DirectoryInfo(AppContext.BaseDirectory);
-                            while (current != null)
-                            {
-                                if (File.Exists(Path.Combine(current.FullName, "Scaffolding.sln")))
-                                {
-                                    artifactsDir = Path.Combine(current.FullName, "artifacts", "build");
-                                    nugetConfigPath = Path.Combine(current.FullName, "NuGet.config");
-                                    break;
-                                }
-                                current = current.Parent;
-                            }
-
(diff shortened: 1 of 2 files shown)


test/Shared/MsBuildProjectSetupHelper.cs
```diff
@@ -20,43 +20,10 @@ internal class MsBuildProjectSetupHelper
...
-                                if (File.Exists(Path.Combine(current.FullName, "Scaffolding.sln")))
-                                {
-                                    artifactsDir = Path.Combine(current.FullName, "artifacts", "build");
-                                    nugetConfigPath = Path.Combine(current.FullName, "NuGet.config");
-                                    break;
-                                }
-                                current = current.Parent;
-                            }
-
-                            _nugetConfigText = MsBuildProjectStrings.GetNugetConfigTxt(artifactsDir, nugetConfigPath);
```
prafullbhosale: @JunTaoLuo @natemcmaster 
We need this to make sure we are testing the packages we built. 
Whats the way of specifying local sources for restore using `<RestoreSources>` ?

I locally tried specifying the full path of the artifacts directory as one of the sources, but that doesn't seem to work.

natemcmaster: The full path should work. What error do you see?

prafullbhosale: Restore fails to find the package and the 'Nearest version ... ' error messages doesn't have the local feed in the list of things it checked.

natemcmaster: This looks like an error @JunTaoLuo and I hit while testing this. NuGet appears to have a bug in restore if you don't set RestoreSources to the same value in csproj *and* all ProjectReferences.

prafullbhosale: Yup, that seems to be the issue. 
Will send a PR with the changes to include the local artifacts folder in the sources.

JunTaoLuo: Oops, didn't notice we weren't testing with the packages that were just built. I'm not sure how restore succeeds since I thought the code generation versions are pinned: https://github.com/aspnet/Scaffolding/blob/rel/2.0.1/test/Shared/MsBuildProjectSetupHelper.cs#L44-L51

prafullbhosale: It doesn't http://aspnetci/viewLog.html?buildId=340768&buildTypeId=Lite_Scaffolding&tab=buildLog&_focus=1541
Looks like it resolves to a lower version.

JunTaoLuo: Ah I see. The patch builds were testing with the correct versions which is why I didn't notice this. Odd that the tests pass locally as well.

PR #640, opened 2017-10-05
Set output encoding to UTF8

### LINKED ISSUE #635: Set output encoding to Unicode
The redirected output shows up garbled in the VS output window 

image

prafullbhosale:
Fixes #635 

cc @mlorbetske


prafullbhosale:
Rebased from `rel/2.0.1` and removed the change for test projects as that's already made in `rel/2.0.1`. 
Will merge this if the CI build is green.

Issue #645, opened 2017-10-23
Microsoft.VisualStudio.Web.CodeGeneration.Design 2.0.1 causes package downgrade error NU1605 for Microsoft.NETCore.App


natemcmaster:
Microsoft.VisualStudio.Web.CodeGeneration.Design has Microsoft.NETCore.App listed in its nuspec dependencies. This causes issues when the .NET Core SDK picks an implicit reference to Microsoft.NETCore.App that is lower than what the .Design package depends on.

**Repro:**
Create app with the 2.0.0 SDK.
Add or upgrade Microsoft.VisualStudio.Web.CodeGeneration.Design to 2.0.1
Restore.

**Result:**
```
error NU1605: Detected package downgrade: Microsoft.NETCore.App from 2.0.3 to 2.0.0. Reference the package directly from the project to select a different version.
 WebApplication1 (>= 1.0.0) -> Microsoft.VisualStudio.Web.CodeGeneration.Design (>= 2.0.1) -> Microsoft.NETCore.App (>= 2.0.3)
```

**Workarounds:**

1. Upgrade Microsoft.NETCore.App (recommended)
```xml
<PropertyGroup>
  <RuntimeFrameworkVersion>2.0.3</RuntimeFrameworkVersion>
</PropertyGroup>
```

2. Suppress NU1605 (may hide other issues)
```xml
<PropertyGroup>
  <NoWarn>$(NoWarn);NU1605</NoWarn>
</PropertyGroup>
```

**Recommendation:**
We should remove Microsoft.NETCore.App from the nuspec dependencies and rely on the project to determine the runtime version use to launch dotnet-aspnet-codegenerator-design.exe

natemcmaster:
cc @mikeharder @prafullbhosale

natemcmaster:
Fixed in https://github.com/aspnet/Scaffolding/pull/646 in rel/2.0.1. Leaving open and on my plate to back-port to dev

natemcmaster:
This was brought back to dev in https://github.com/aspnet/Scaffolding/pull/651

mnmr:
The Microsoft.VisualStudio.Web.CodeGeneration.Tools package has the same problem (still present in latest v2.0.2 release). It'd be great if a similar fix could be applied there.

natemcmaster:
@mmr is your reference to Microsoft.VisualStudio.Web.CodeGeneration.Tools a DotNetCliToolReference or a PackageReference? It should be the former.

[...]

PR #646, opened 2017-10-24
[2.0.x] Remove Microsoft.NETCore.App from .Design nuspec

natemcmaster:
Resolves https://github.com/aspnet/Scaffolding/issues/645

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -1,18 +1,18 @@
 ﻿<Project>
   <PropertyGroup>
-    <InternalAspNetCoreSdkPackageVersion>2.0.2-rc1-15526</InternalAspNetCoreSdkPackageVersion>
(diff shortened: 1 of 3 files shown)


mikeharder:
@natemcmaster: Do we ship any other packages which target netcoreapp2.0?  Do we use the same pattern in their csproj?

natemcmaster:
:watch: investigating test failures...

natemcmaster:
@mikeharder this is the only package in 2.0.x that has this issue.

natemcmaster:
The test issue: https://github.com/aspnet/Scaffolding/issues/647. Not relate to this change. Moving forward with this update.

prafullbhosale:
> The test issue: #647. Not relate to this change. Moving forward with this update.

Makes sense, wasn't able to see the failures on my machine.

PR #648, opened 2017-10-30
Fix package structure for CodeGeneration.Design

prafullbhosale:
This fixes the issue with the architecture mismatch. 

cc @mlorbetske @natemcmaster @rynowak


prafullbhosale:
Related https://github.com/aspnet/BuildTools/issues/450

prafullbhosale:
cc @Eilon 

Merging this now as we have an approved TFS bug for this.

PR #651, opened 2017-11-02
Merge rel/2.0.1 back to dev

natemcmaster:
Changes:

 - Removes PackageLineup and pins to package and tool versions - https://github.com/aspnet/Universe/issues/575
 - Uses those versions to generate MsBuildProjectStrings
 - Includes other fixes from 2.0.1, such as the anycpu/x86/x64 improvements
 - Ensure global.json is copied to the e2e test context to avoid SDK lifting to new versions


.gitignore
```diff
@@ -37,4 +37,4 @@ project.fragment.lock.json
 backup/**/*.*
 .dotnet/**
 global.json
-korebuild-lock.txt
```
prafullbhosale: Intentional?

natemcmaster: Yup. We want builds to be reproducible, so we've started committing exact package and tools versions to source, both in dev and rel/2.0.x branches.

Issue #407, opened 2017-02-23
Scaffolding using separate assemblies for model/context 


jamiewest:
Scaffoling using dotnet aspnet-codgenerator --_commands_ only seems to work if the model and DbContext are located in the same assembly as the Mvc app.

## Steps to reproduce
```
1. Add domain model to it's own assembly, i.e. Blog class in AssemblyName.Domain
2. Add EF Context to its own assembly with DbSet<Blog>, i.e. ModelContext class in Assembly.Data
3. Create MvcApp referencing both assemblies above and load ModelContext into DI.
4. Run :: dotnet aspnet-codegenerator --project . controller -name BlogsController -m Blog -dc ModelContext -outDir Controllers
```
## Expected behavior
```
Writes BlogsController class to the Controllers folder using supplied model and DbContext.
```

## Actual behavior
```
Could not add Model type 'AssemblyName.Domain.Blog' to DbContext 'AssemblyName.Data.ModelContext'. Please make sure that 'AssemblyName.Data.ModelContext' has a DbS
et property for 'AssemblyName.Domain.Blog'.
   at Microsoft.VisualStudio.Web.CodeGeneration.ActionInvoker.<BuildCommandLine>b__6_0()
   at Microsoft.Extensions.CommandLineUtils.CommandLineApplication.Execute(String[] args)
   at Microsoft.VisualStudio.Web.CodeGeneration.CodeGenCommand.Execute(String[] args)
```
## Environment
dotnet --info output:
```
.NET Command Line Tools (1.0.0-rc4-004771)

Product Information:
 Version:            1.0.0-rc4-004771
 Commit SHA-1 hash:  d881d45b75

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\1.0.0-rc4-004771
```
.csproj file
```xml
<Project Sdk="Microsoft.NET.Sdk.Web">
  <PropertyGroup>
    <TargetFramework>netcoreapp1.1</TargetFramework>
    <PackageTargetFallback>portable-net45;win8;wp8;wpa81</PackageTargetFallback>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore" Version="1.1.0" />
    <PackageReference Include="Microsoft.AspNetCore.Mvc" Version="1.1.1" />
[...]

PR #658, opened 2017-11-28
Fix issue that caused scaffolding to fail if DbContext class was in a dependency

prafullbhosale:
This is a followup to to #407 

https://github.com/aspnet/Scaffolding/issues/407#issuecomment-284031010
The fix for #407 allowed the Model class to be in a separate assembly, but the `DbContext` class still had to be part of the web project. 

With this change, the `DbContext` can be present in a dependency of the web project.


prafullbhosale:
Fixes #663

Issue #689, opened 2018-02-15
Move to RazorProjectEngine.


NTaylorMullen:
MvcRazorTemplateEngine isn't used any more. Use RazorProjectEngine.

NTaylorMullen:
0176fab6446174125fd3289ef5a988641669f9c8

PR #691, opened 2018-02-16
Move to RazorProjectEngine.

NTaylorMullen:
#689

/cc @mkArtakMSFT

### DIFF
--- a/src/VS.Web.CG.Templating/RazorTemplating.cs
+++ b/src/VS.Web.CG.Templating/RazorTemplating.cs
@@ -4,6 +4,7 @@
 using System;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


src/VS.Web.CG.Templating/RazorTemplating.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.AspNetCore.Razor.Language;
 using Microsoft.AspNetCore.Mvc.Razor.Extensions;
 using Microsoft.VisualStudio.Web.CodeGeneration.Templating.Compilation;
+using System.Text;
```
prafullbhosale: `System.*` on top of others.

NTaylorMullen:
Awesome, will get this in once we have packages from coherence so local builds work.

PR #694, opened 2018-02-27
React to new dependencies

ryanbrandenburg:
This should unblock Update Universe.

### DIFF
--- a/NuGetPackageVerifier.json
+++ b/NuGetPackageVerifier.json
@@ -8,6 +8,7 @@
         "Exclusions": {
           "NEUTRAL_RESOURCES_LANGUAGE": {
             "tools/any/any/Newtonsoft.Json.dll": "Third party assembly bundled in our nupkg.",
+            "tools/any/any/System.Buffers.dll": "Third party assembly bundled in our nupkg.",
             "tools/any/any/System.Runtime.CompilerServices.Unsafe.dll": "Assembly built by another project but bundled in our nupkg."
           },
           "SERVICING_ATTRIBUTE": {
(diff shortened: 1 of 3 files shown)


ryanbrandenburg:
YOLO pushing this to fix Update Universe.

NuGetPackageVerifier.json
```diff
@@ -20,7 +21,12 @@
           "WRONG_PUBLICKEYTOKEN": {
             "tools/any/any/Newtonsoft.Json.dll": "Third party assembly bundled in our nupkg.",
             "tools/any/any/NuGet.Frameworks.dll": "Assembly built by another project but bundled in our nupkg.",
-            "tools/any/any/System.Runtime.CompilerServices.Unsafe.dll": "Assembly built by another project but bundled in our nupkg."
+            "tools/any/any/System.Buffers.dll": "Assembly built by another project but bundled in our nupkg.",
```
pranavkm: FYI @davidfowl - this is a result of the change to Primitives

davidfowl: Good to know, the span change?

pakrym: Why is it the result? Shouldn't it be trimmed?

pakrym:
```
NEUTRAL_RESOURCES_LANGUAGE @ tools/any/any/runtimes/unix/lib/netcoreapp2.0/System.Buffers.dll: The managed assembly 'tools/any/any/runtimes/unix/lib/netcoreapp2.0/System.Buffers.dll' in this package is missing the '[assembly: NeutralResourcesLanguage("en-us")]' attribute.
NEUTRAL_RESOURCES_LANGUAGE @ tools/any/any/runtimes/win/lib/netcoreapp2.0/System.Buffers.dll: The managed assembly 'tools/any/any/runtimes/win/lib/netcoreapp2.0/System.Buffers.dll' in this package is missing the '[assembly: NeutralResourcesLanguage("en-us")]' attribute.
```

NuGetPackageVerifier.json
pakrym: Should we cross compile primitives so we are not leaking package updates into 2.0 apps?

davidfowl: Uh, I think no. Otherwise we'd end up doing that for a bunch of stuff no?

pranavkm: Cross compile to what exactly? Doesn't it target `ns2.0` today?

pakrym: Cross-compile and reference 4.3.0 of system.memory in netstandard 2.0 version.

pakrym: OK, let's just fix the rest of validation issues then.

pranavkm: @ryanbrandenburg could you follow up on the fix? You'll likely need to port the changes to this file to Coherence-Signed  too

PR #708, opened 2018-03-07
Fix build warnings about missing template variables

natemcmaster:
The .cs.in file had undefined variables and was causing a build warning.


natemcmaster:
Merging as the changes are minor and tests are passing on CI.

natemcmaster:
cref https://github.com/aspnet/Coherence-Signed/issues/601

PR #710, opened 2018-03-14
Create or update baselines to match 2.0

ryanbrandenburg:


ryanbrandenburg:
@javiercn and/or @Eilon we don't ship anything in this repo as public API do we?

Eilon:
@ryanbrandenburg I believe that is correct, but adding @prafullbhosale . Prafull - is there any API in this repo that we need to account for when detecting breaking changes?

prafullbhosale:
>  Prafull - is there any API in this repo that we need to account for when detecting breaking changes?

No, we don't want to track any APIs (that we ship right now) for breaking changes.
Since the packages here are intended to be `PrivateAssets=All`, users should not be using the APIs from scaffolding in their code. 

We do have extension points, but they are largely convention based (and an empty interface) 
[ICodeGenerator](https://github.com/aspnet/Scaffolding/blob/dev/src/VS.Web.CG.Core/ICodeGenerator.cs)

/cc @seancpeters

ryanbrandenburg:
OK, I'm going to close this PR based off that then.

PR #713, opened 2018-03-23
Bug fixes for identity scaffolder.

prafullbhosale:
- Changed the default name for the generated DbContext to be `IdentityDataContext` 
  - Earlier it was `IdentityDbContext` and it would clash with the one from Identity assemblies.
- Enabled passing in an existing user class in case a new DbContext is being generated.
- Some pages under `Pages/Account/Mangage` have a dependency on `/Pages/Account/Manage/_ManageNav.cshtml`. 
  - Added support to scaffold dependency pages when they are not selected. 
  - Also added E2E tests that test scaffolding each individual page and ensure that the project still builds.


prafullbhosale:
Going to merge this now before we enter ask mode.

PR #716, opened 2018-03-30
Remove reference to Microsoft.AspNetCore.Mvc.Razor.Extensions

pranavkm:
Between preview1 and preview2, the bitness of .NET Framework targeting projects changed from x86 to AnyCPU [1]. Consequently the app is compiled with AnyCPU, but the Scaffolding binary is x86. In both preview1 and preview2, Mvc attempts to load Scaffolding binaries because it appears to be a candidate containing controllers[2]. However, the change in bitness of the app results in an exception when attempting to Assembly.Load scaffolding's binary. Possible suggested fixes:

* Make Scaffolding not appear to be a candidate for Mvc
* Put the correct bitness exe in the output directory.

[1] - Between Preview1 and Preview2, Asp.Net Core no longer depends on Libuv, a native dependency. This causes the PlatformTarget to change. Here's the culprit: https://github.com/dotnet/sdk/blob/34454e6b651d4e1f0f1149c13d8c1ec115e111a1/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets#L158-L167

[2] - This is because Scaffolding references Microsoft.AspNetCore.Mvc.Razor.Extensions, an Mvc specific library.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -12,7 +12,7 @@
     <MicrosoftAspNetCoreIdentityEntityFrameworkCorePackageVersion>2.1.0-preview2-30478</MicrosoftAspNetCoreIdentityEntityFrameworkCorePackageVersion>
     <MicrosoftAspNetCoreIdentityUIPackageVersion>2.1.0-preview2-30478</MicrosoftAspNetCoreIdentityUIPackageVersion>
     <MicrosoftAspNetCoreMvcPackageVersion>2.1.0-preview2-30478</MicrosoftAspNetCoreMvcPackageVersion>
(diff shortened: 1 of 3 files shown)


pranavkm:
@Eilon this changes a package dependency for Scaffolding which is added as a regular (not a  developer dependency) to user's projects. However, it's not the common path through which users acquire this dependency.

pranavkm:
Travis OSX keeps failing with no error message. AppV + Travis Ubuntu passed which means this is good vis-a-vis CI.

PR #726, opened 2018-04-05
Set projects under tools/ as IsPackable=false

natemcmaster:
It throws off Universe's groove because these projects under tools/ appear to produce the same package ID.


prafullbhosale:
cc @seancpeters

Issue #718, opened 2018-04-05
[Preview2-Validation] 'Add Area' scaffolding generates incorrect code in ScaffoldingReadme.txt


kichalla:
Repro:
- Create a new MVC application | right-click on the project | Add | Area | give a name like 'Admin'
- Above step creates a ScaffoldingReadme.txt with the following content:

```
Scaffolding has generated all the files and added the required dependencies.

However the Application's Startup code may required additional changes for things to work end to end.
Add the following code to the Configure method in your Application's Startup class if not already done:

        app.UseMvc(routes =>
        {
          route.MapRoute(
            name : "areas",
            template : "{area:exists}/{controller=Home}/{action=Index}/{id?}"
          );
        });
```

Note that the code should say `routes.MapRoute` and not `route.MapRoute`

Also could we add the ScaffoldingReadme.txt to .gitignore too?

PR #747, opened 2018-04-10
Readme fix route to routes  issue 718

seancpeters:
Fix for #718


src/VS.Web.CG.Mvc/Identity/IdentityGeneratorTemplateModelBuilder.cs
```diff
@@ -185,7 +185,7 @@ internal Type UserType
             var layout = _commandlineModel.Layout;
             if (_commandlineModel.GenerateLayout)
             {
-                layout = "/Pages/Shared/_Layout.chstml";
```
prafullbhosale: Is this change intentional here? 
There's another PR from release/2.1 that is bringing in this change. Might lead to conflicts.

Issue #724, opened 2018-04-05
[Preview2-Validation] Adding Identity scaffolding should add LoginPartial if there isn't one already


kichalla:
*VS version:**
15.7.Preview 3.0 [27604.0.d15.7]

**Repro:**
- Create 'WebApplication` based project (Choose to create a git repoistory as it makes it easier to see file changes) | right-click on project | Add | New Scaffolded Item.. | choose Identity on left pane | Identity | Add
- Choose `/Pages/Shared/_Layout.chstml` in the text box to choose existing layout | Add a new DbContext and User types | Add
- Ctrl + F5 and you should see the following error in the browser
![image](https://user-images.githubusercontent.com/6559784/38387038-6ec0ef84-38cb-11e8-8727-033927b70427.png)

**Expected**:
Ctrl +F5 works fine with the login partial (Register, Login link) on the navigation bar

**Actual**:
Error related to `_LoginPartial` not found.

**Suggestion**:
Can we check if a file called `_LoginPartial.cshtml` already exists under `Pages` or `Pages\Shared` folders. If it does not exist, we should automatically add the partial to the application.

Also noticed that there are bunch of edits to the `Pages\Shared\_Layout.cshtml` page. Many of these should be avoided and are incorrect.
![image](https://user-images.githubusercontent.com/6559784/38387281-20a1ed16-38cc-11e8-8103-254a0c323d1e.png)
![image](https://user-images.githubusercontent.com/6559784/38387303-2d80c426-38cc-11e8-9109-35a429656ab5.png)

seancpeters:
@javiercn @mkArtakMSFT @prafullbhosale 
Is it correct that the _LoginPartial.cshtml should be left alone if it already exists?
My plan is to check for it in both:
- `Pages\Shared\_LoginPartial.cshtml`
- `Pages\_LoginPartial.cshtml`
...and if it already exists, do nothing. If it doesn't exist, create it under `Pages\Shared`

PR #748, opened 2018-04-11
_LoginPartial is looked for in multiple places, not generated if exists

seancpeters:
Fixes part of issue #724

### DIFF
--- a/src/VS.Web.CG.Mvc/Identity/IdentityGenerator.cs
+++ b/src/VS.Web.CG.Mvc/Identity/IdentityGenerator.cs
@@ -201,7 +201,7 @@ namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Identity
             foreach (var template in templates)
             {
                 var outputPath = Path.Combine(projectDir, template.OutputPath);
-                if (template.ShouldOverWrite != OverWriteCondition.Never || !_fileSystem.FileExists(outputPath))
+                if (template.ShouldOverWrite != OverWriteCondition.Never || !DoesTemplateFileExist(template, projectDir))
                 {
                     // We never overwrite some files like _ViewImports.cshtml.
                     _logger.LogMessage($"Adding template: {template.Name}", LogMessageLevel.Trace);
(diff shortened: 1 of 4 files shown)


src/VS.Web.CG.Mvc/Identity/identitygeneratorfilesconfig.json
```diff
@@ -5,7 +5,9 @@
         "Name": "_LoginPartial",
         "SourcePath": "_LoginPartial.cshtml",
         "OutputPath": "Pages/Shared/_LoginPartial.cshtml",
-        "IsTemplate": true
+        "IsTemplate": true,
+        "AltPaths": [ "Pages/_LoginPartial.cshtml" ],
+        "ShouldOverWrite": "Never"
```
prafullbhosale: If this file should never be overwritten, then it shouldn't be a candidate for scaffolding individually.

seancpeters: @javiercn @mkArtakMSFT - thoughts on Prafull's comment?

javiercn: I'm not really an expert on the scaffolder, but sounds about right. We only want to generate the login partial if there isn't an existing one

src/VS.Web.CG.Mvc/Identity/IdentityGenerator.cs
```diff
@@ -222,6 +222,27 @@ private void ShowFileList()
...
+        // Returns true if the template file exists in it's output path, or in an alt path (if any are specified)
+        private bool DoesTemplateFileExist(IdentityGeneratorFile template, string projectDir)
+        {
+            string outputPath = Path.Combine(projectDir, template.OutputPath);
+            if (_fileSystem.FileExists(outputPath))
+            {
+                return true;
+            }
+
+            foreach (string altPath in template.AltPaths)
```
mkArtakMSFT: Minor: You can rewrite whole this section with:
```csharp
return template.AltPaths.Any(altPath=>_fileSystem.FileExists(Path.Combine(projectDir, altPath));
```

src/VS.Web.CG.Mvc/Identity/identitygeneratorfilesconfig.json
mkArtakMSFT: What does the "ShouldOverwrite":"Never" meam?

seancpeters: @javiercn @prafullbhosale @vijayrkn @mkArtakMSFT 
If we make _LoginPartial not be a candidate for scaffolding, then we'll also need to make a change in VS for the Identity Scaffolder UI - removing the checkbox for login partial.

seancpeters: When identity scaffolding occurs, files marked with ShouldOverWrite = Never are checked for existence in the project being scaffolded. If the file exists, then scaffolding doesn't make a new one - doesn't overwrite the same-named file that the user has already created. 

If the project being scaffolded does not have this file, it can be created.  But it only gets created if the user has chosen that file to be created (or selected all files).

PR #749, opened 2018-04-11
Razor validation scripts partial smarter conditional add

### LINKED ISSUE #721: [Preview2-Validation] Scaffolding razor pages using EF CRUD generates _ValidationScriptsPartial
_From @kichalla on April 4, 2018 23:3_

Repro:
- Create a new `WebApplication` template based project (This should create `_ValidationScriptsPartial.cshtml` under `Pages\Shared` folder)
- Add the following class in `Data` folder
```c#
public class Customer
{
    public int Id { get; set; }
    public int ZipCode { get; set; }
}
```
- Add a new folder `Customers` under `Pages` folder
- Right click on `Customers` folder | Add Razor Page.. | Razor Pages using EntityFramework (CRUD) | Add | Choose `Customer` model class and `ApplicationDbContext` | Add

**Expected**: `_ValidationScriptsPartial.cshtml` should not be regenerated as there is already one under `Pages\Shared` folder
**Actual**: A new `_ValidationScriptsPartial.cshtml` file is created under `Pages` folder.

_Copied from original issue: aspnet/templating#449_

seancpeters:
Fixes #721 

Checks for an existing _ValidationScriptsPartial in the Pages/Shared/ directory in addition to checking in Pages/

### DIFF
--- a/src/VS.Web.CG.Mvc/Common/RequiredFileEntity.cs
+++ b/src/VS.Web.CG.Mvc/Common/RequiredFileEntity.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 
 namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc
 {
@@ -23,6 +24,15 @@ namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc
             OutputPath = outputPath;
         }
         
+        public RequiredFileEntity(string outputPath, string templateName, List<string> altPaths)
+            : this(outputPath, templateName)
+        {
+            if (altPaths != null)
+            {
+                AltPaths = altPaths;
+            }
+        }
+
         /// <summary>
         /// Name of the template file.
         /// </summary>
@@ -32,5 +42,10 @@ namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc
         /// Path Relative to the .csproj file
         /// </summary>
         public string OutputPath { get; private set; }
+
+        /// <summary>
+        /// A list of other paths to check for file existence - to avoid generating new versions of existing files when not appropriate).
+        /// </summary>
+        public List<string> AltPaths { get; private set; } = new List<string>();
     }
 }
--- a/src/VS.Web.CG.Mvc/Common/RequiredFileEntity.cs
+++ b/src/VS.Web.CG.Mvc/Common/RequiredFileEntity.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc
 {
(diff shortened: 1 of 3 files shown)


seancpeters:
@prafullbhosale - no, it shouldn't have been against dev...I need to stop doing that.

src/VS.Web.CG.Mvc/Common/RequiredFileEntity.cs
```diff
@@ -23,6 +24,15 @@ public RequiredFileEntity(string outputPath, string templateName)
             OutputPath = outputPath;
         }
         
+        public RequiredFileEntity(string outputPath, string templateName, List<string> altPaths)
+            : this(outputPath, templateName)
+        {
+            if (altPaths != null)
```
mkArtakMSFT: Minor: Instead of those this check and the default value assignment, you can just have:
```csharp
    AltPaths = altPaths ?? new List<string>();
```

src/VS.Web.CG.Mvc/RazorPage/RazorPageScaffolderBase.cs
```diff
@@ -85,6 +85,27 @@ public virtual IEnumerable<string> TemplateFolders
...
         }
 
+        private bool ShouldFileBeAdded(RequiredFileEntity fileEntity)
+        {
+            if (File.Exists(Path.Combine(ApplicationInfo.ApplicationBasePath, fileEntity.OutputPath)))
+            {
+                return false;
+            }
+
+            if (fileEntity.AltPaths != null)
```
mkArtakMSFT: No need for this check, as it can't be ever null.

src/VS.Web.CG.Mvc/Common/RequiredFileEntity.cs
```diff
@@ -23,6 +24,15 @@ public RequiredFileEntity(string outputPath, string templateName)
             OutputPath = outputPath;
         }
         
+        public RequiredFileEntity(string outputPath, string templateName, List<string> altPaths)
```
mkArtakMSFT: Minor: Accepting `IEnumerable<string>` instead would simplify the work of the callers.

src/VS.Web.CG.Mvc/Common/RequiredFileEntity.cs
```diff
@@ -24,13 +25,10 @@ public RequiredFileEntity(string outputPath, string templateName)
...
         
-        public RequiredFileEntity(string outputPath, string templateName, List<string> altPaths)
+        public RequiredFileEntity(string outputPath, string templateName, IEnumerable<string> altPaths)
             : this(outputPath, templateName)
         {
-            if (altPaths != null)
-            {
-                AltPaths = altPaths;
-            }
+            AltPaths = altPaths.ToList() ?? new List<string>();
```
mkArtakMSFT: This may result in a NRE.
Use `altPaths?.ToList()` to avoid it.

mkArtakMSFT: Also, you don't need the `= new List<string>()` any more for the `AltPath` property

PR #775, opened 2018-04-30
Merge release/2.1

natemcmaster:
Are these changes expected to be brought back to dev?


natemcmaster:
Ok, thanks for clarifying. 

The general strategy is that any change made to release/x.y branches directly should be merged back to `dev`. This helps us ensure code stays mainline for future releases.

PR #787, opened 2018-05-11
Fixes the tilde directory problem with existing layout files

### LINKED ISSUE #786: Identity scaffolding adds extra '~' directory with support files
When scaffolding identity, in addition to the files which should be created, a top-level directory named '~' is created, containing the directory `Views/Shared/` which contains some support files:
image

This happens both when an existing _Layout.cshtml file is chosen, or if a new one is being generated.

@vijayrkn @javiercn

seancpeters:
Fixes #786

### DIFF
--- a/src/VS.Web.CG.Mvc/Identity/IdentityGeneratorTemplateModelBuilder.cs
+++ b/src/VS.Web.CG.Mvc/Identity/IdentityGeneratorTemplateModelBuilder.cs
@@ -261,8 +261,22 @@ namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Identity
 
             if (!string.IsNullOrEmpty(_commandlineModel.Layout))
             {
-                supportFileLocation = Path.GetDirectoryName(_commandlineModel.Layout);
-                layoutFile = _commandlineModel.Layout;
+                if (_commandlineModel.Layout.StartsWith("~"))
+                {
+                    layoutFile = _commandlineModel.Layout.Substring(1);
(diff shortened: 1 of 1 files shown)


seancpeters:
@javiercn @prafullbhosale - just hoping to get a quick look at this change - it's being taken to shiproom tomorrow.

src/VS.Web.CG.Mvc/Identity/IdentityGeneratorTemplateModelBuilder.cs
```diff
@@ -261,8 +261,22 @@ private bool DetermineSupportFileLocation(out string supportFileLocation, out st
...
+                if (_commandlineModel.Layout.StartsWith("~"))
+                {
+                    layoutFile = _commandlineModel.Layout.Substring(1);
+                }
+                else
+                {
+                    layoutFile = _commandlineModel.Layout;
+                }
+
+                if (layoutFile[0] == '\\' || layoutFile[0] == '/')
```
prafullbhosale: This already handles `~/` and `~\` cases, should this support crazy input like `\\\abc\layout` or `\/abc\layout`

seancpeters: Yes, it probably should. The path is assumed to always be relative to the project root, so all lead slashes should be removed. Fixing...

seancpeters:
Closing this, it's replaced with #789

PR #795, opened 2018-05-16
Update CDN to pull jQuery to 3.3 and update SRI hashes

natemcmaster:
Resolves https://github.com/aspnet/Scaffolding/issues/794

### DIFF
--- a/test/VS.Web.CG.Mvc.Test/CdnScriptTagTests.cs
+++ b/test/VS.Web.CG.Mvc.Test/CdnScriptTagTests.cs
@@ -58,7 +58,7 @@ namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc
 
             Assert.All(scriptTags, t =>
             {
-                Assert.True(shasum[t.Src] == t.Integrity, userMessage: $"Expected integrity on script tag to be {shasum[t.Src]} but it was {t.Integrity}. {t.FileName}");
+                Assert.True(shasum[t.Src] == t.Integrity, userMessage: $"Expected integrity on script tag to be {shasum[t.Src]} but it was {t.Integrity}: {t.Path}");
             });
         }
 
@@ -67,6 +67,7 @@ namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc
             public string Src;
             public string Integrity;
             public string FileName;
+            internal string Path;
         }
 
         private static readonly Regex _scriptRegex = new Regex(@"<script[^>]*src=""(?'src'http[^""]+)""[^>]*integrity=""(?'integrity'[^""]+)""([^>]*)>", RegexOptions.Multiline);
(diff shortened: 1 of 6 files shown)


natemcmaster:
found a few more. :up: :date:

natemcmaster:
Yeah, that's a good point. I'm somewhat concerned users will be burned by line-ending issues, which affects the shasum of the file.

By the way, what is this file for? https://github.com/aspnet/Scaffolding/blob/release/2.1/src/VS.Web.CG.Mvc/Templates/StaticFiles/Scripts/jquery-1.10.2.js

natemcmaster:
Found a few more. Not sure how they are used or if these need to be updated too. 

https://github.com/aspnet/Scaffolding/blob/release/2.1/src/VS.Web.CG.Mvc/Templates/StaticFiles/Scripts/jquery-1.10.2.js
https://github.com/aspnet/Scaffolding/blob/release/2.1/src/VS.Web.CG.Mvc/Templates/StaticFiles/Scripts/jquery.validate.js
https://github.com/aspnet/Scaffolding/blob/release/2.1/src/VS.Web.CG.Mvc/Templates/StaticFiles/Scripts/jquery-1.10.2.min.js
https://github.com/aspnet/Scaffolding/blob/release/2.1/src/VS.Web.CG.Mvc/Templates/StaticFiles/Scripts/jquery-1.10.2.intellisense.js
https://github.com/aspnet/Scaffolding/blob/release/2.1/src/VS.Web.CG.Mvc/Templates/StaticFiles/Scripts/jquery.validate-vsdoc.js
https://github.com/aspnet/Scaffolding/blob/release/2.1/src/VS.Web.CG.Mvc/Templates/StaticFiles/Scripts/jquery.validate.min.js

kichalla:
I have no idea when these files are used. @vijayrkn / @seancpeters probably know

vijayrkn:
@seancpeters & @prafullbhosale  - Can you comment on the new files?

javiercn:
@prafullbhosale Will probably have a better idea

test/VS.Web.CG.Mvc.Test/CdnScriptTagTests.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        private struct ScriptTag
+        {
+            public string Src;
+            public string Integrity;
+            public string FileName;
+            internal string Path;
+        }
+
+        private static readonly Regex _scriptRegex = new Regex(@"<script[^>]*src=""(?'src'http[^""]+)""[^>]*integrity=""(?'integrity'[^""]+)""([^>]*)>", RegexOptions.Multiline);
```
seancpeters: Nits:
The final capture group (unnamed) is unnecessary and can be removed, changing
`([^>]*)` to `[^>]*`

Also, performance will be better using lazy captures, changing each `*` to `*?`

An unformatted regex is difficult to read. Using RegexOptions.IgnorePatternWhitespace it can be formatted to be more readable. Including the above changes, something like:
```
        private static readonly Regex _scriptRegex = new Regex(
            @"<script[^>]*?
                src=""(?'src'http[^""]+)""[^>]*?
                integrity=""(?'integrity'[^""]+)""
                [^>]*?>"
                , RegexOptions.Multiline | RegexOptions.IgnorePatternWhitespace);
```

kichalla: Regexes are hard to understand. We should use AngleSharp in future.

natemcmaster: Thanks for the feedback. Feel free to change it in dev. The release/2.1 ship already sailed though.

natemcmaster:
For follow up on jQuery 1.10, see https://github.com/aspnet/Scaffolding/issues/796

PR #818, opened 2018-06-13
Set 2.1 baselines

ryanbrandenburg:


ryanbrandenburg:
@seancpeters you removed the following API's in aspnet/Scaffolding#807.

- Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Dependency.DependencyInstaller
- Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Dependency.DependencyGenerator
- Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Dependency.DependencyGeneratorModel
- Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Dependency.MvcLayoutDependencyInstaller : Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Dependency.DependencyInstaller
- Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Dependency.StaticFilesDependencyInstaller : Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Dependency.DependencyInstaller

Those classes were publicly visible so I'm just trying to confirm that the deletion as approved/intentional.

CC @natemcmaster who approved the PR and might remember.

seancpeters:
Removing the listed classes was the intent, as those are dead code in the sense that nothing from VS or command line scaffolding ever uses them.

But I don't think the possibility of external callers was considered when the decision was made - it was not on my radar in any case.
@vijayrkn - we may need to revert the removal of the dependency scaffolders #807 and #804

natemcmaster:
Historically we haven't cared about the API surface of scaffolding packages. In fact, IMO the entire API surface should just be made "internal" sinc we only expect users to interact with scaffolding through a small set of APIs in this repo.

cc @prafullbhosale

vijayrkn:
We should stick to the removal of those classes since nothing from VS or command line scaffolding is using them. 

We should do an inspection on the code base to make sure only the ones required are made public.

PR #873, opened 2018-09-11
Add missing Identity template ShowRecoveryCodes in config.

### LINKED ISSUE #799: Scaffold: Missing ShowRecoveryCodes
_From @adams-hub on May 17, 2018 11:29_

Scaffolding is missing the ShowRecoveryCodes page.

_Copied from original issue: aspnet/Identity#1796_

SteveVaneeckhout:
Fixes #799

### DIFF
--- a/src/VS.Web.CG.Mvc/Identity/bootstrap3_identitygeneratorfilesconfig.json
+++ b/src/VS.Web.CG.Mvc/Identity/bootstrap3_identitygeneratorfilesconfig.json
@@ -480,6 +480,32 @@
         "IsTemplate": true
       }
     ],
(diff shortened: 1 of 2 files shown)


seancpeters:
I'm merging this, but please be aware that with chis change, VS will not have an explicit option to select scaffolding the Show RecoveryCodes page. It'll still be generated when selecting scaffolding all pages. I need to make a change in VS to individually select the page. I expect it'll be available in 15.9 Preview3 or Preview4 but I can't guarantee that at this point.

With this change, when scaffolding from the command line, the page will be selectable in the --files list.

Thanks for the contribution!

PR #881, opened 2018-09-20
Fix for EF controller scaffolding when the model is defined on a parent of the selected db context

seancpeters:
Before this fix, if the model property is defined on a class that the selected db context is derived from, the model property would be added on the selected db context. This would hide the base classes model property.

### DIFF
--- a/Scaffolding.sln
+++ b/Scaffolding.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2024
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.28124.4003
 MinimumVisualStudioVersion = 15.0.26730.03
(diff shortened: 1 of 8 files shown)


prafullbhosale:
Might be good to add a test for this scenario.

seancpeters:
When a DbSet<[ModelType]> property already exists on the DbContext, that property's name will be used in the generated controller. Prior to the fix, when scaffolding your project, since the DbSet<Blog> was on a parent of the DbContext, it wasn't found, and so a new property with the same name as the model was created, and referenced in the controller. With it fix, since the DbSet<Blog> Blogs will be found in the base context class, the name "Blogs" will be used when generating the controller. Similar behavior can be seen in the existing code if you put the DbSet on the DbContext chosen for scaffolding - whatever it is named will be referenced in the generated controller.

This will take care of the issue where you now have to rename the property in the controller, in addition to having to remove the property from the derived db context.

seancpeters:
Good suggestion on the do loop, updated accordingly.

PR #882, opened 2018-09-20
make all scaffolding template files UTF-8 with BOM

seancpeters:
The files modified in this PR did not have a BOM, they do now.


ryanbrandenburg:
That actual code looks fine, I'm just wondering if we've got a bit more information on when we do and don't need a BOM. It seems this is fairly environment specific, some tools expect/demand a BOM to do things "right" and [others seem to have a problem with it](https://github.com/aspnet/mvc/issues/4671).

seancpeters:
@ryanbrandenburg - you're right, we should think about this a bit more.

ryanbrandenburg:
It seems to me that we're better off following the [Unicode spec](http://www.unicode.org/versions/Unicode5.0.0/ch02.pdf) which says:

> Use of a BOM is neither required nor recommended for UTF-8

I'd rather not include it, allowing anyone with need to add it themselves. I'm not against a configurable option being added to project generation, but I wonder how much use it would get because it seems that most things that would require you to have a BOM you would discover comparatively late in development.

natemcmaster:
This PR looks stale. Is this still needed for 2.2? If not, can we close?

seancpeters:
Yeah, this is stale. We decided not to do this, closing.

PR #887, opened 2018-09-26
Reference Microsoft.NET.Sdk.Razor in projects with Razor files

pranavkm:


seancpeters:
@pranavkm - thank you for looking into this!

PR #899, opened 2018-09-28
Fix microbuild signing configuration

natemcmaster:
Fix the file paths for code signing. cref https://github.com/aspnet/BuildTools/pull/766

### DIFF
--- a/src/dotnet-aspnet-codegenerator/dotnet-aspnet-codegenerator.csproj
+++ b/src/dotnet-aspnet-codegenerator/dotnet-aspnet-codegenerator.csproj
@@ -66,20 +66,19 @@
 
   <ItemGroup>
     <!-- Assemblies from this project -->
-    <SignedPackageFile Include="tools/$(TargetFramework)/any/toolassets/$(TargetFramework)/Microsoft.VisualStudio.Web.CodeGeneration.Msbuild.dll" Certificate="$(AssemblySigningCertName)" />
-    <SignedPackageFile Include="tools/$(TargetFramework)/any/toolassets/net461/Microsoft.VisualStudio.Web.CodeGeneration.Msbuild.dll" Certificate="$(AssemblySigningCertName)" />
-    <SignedPackageFile Include="tools/$(TargetFramework)/any/Microsoft.VisualStudio.Web.CodeGeneration.Contracts.dll" Certificate="$(AssemblySigningCertName)" />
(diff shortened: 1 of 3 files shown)


natemcmaster:
@HaoK this is the last one we need before we can merge https://github.com/aspnet/BuildTools/pull/766. Can you take a quick look?

src/dotnet-aspnet-codegenerator/dotnet-aspnet-codegenerator.csproj
```diff
@@ -66,20 +66,19 @@
...
 
-    <!-- Binaries from partner teams that should be signed already -->
-    <ExcludePackageFileFromSigning Include="tools/$(TargetFramework)/any/NuGet.Frameworks.dll" Certificate="$(AssemblySigningCertName)" />
-    <ExcludePackageFileFromSigning Include="tools/$(TargetFramework)/any/System.Runtime.CompilerServices.Unsafe.dll" Certificate="$(AssemblySigningCertName)" />
+    <!-- Binaries from other projects that should be signed already -->
+    <ExcludePackageFileFromSigning Include="$(PublishDir)Microsoft.Extensions.FileProviders.Physical.dll" PackagePath="tools/$(TargetFramework)/any/Microsoft.Extensions.FileProviders.Physical.dll" />
+    <ExcludePackageFileFromSigning Include="$(PublishDir)Microsoft.Extensions.FileSystemGlobbing.dll" PackagePath="tools/$(TargetFramework)/any/Microsoft.Extensions.FileSystemGlobbing.dll" />
+    <ExcludePackageFileFromSigning Include="$(PublishDir)Microsoft.Extensions.Primitives.dll" PackagePath="tools/$(TargetFramework)/any/Microsoft.Extensions.Primitives.dll" />
```
HaoK: Is this a change from before where we used to sign all the other packages that got pulled in?

natemcmaster: Yes. So each repo will sign in batches now. Before, these files came from the FileSystem repo unsigned and ended up in this package. With MicroBuild signing, they will be signed before the Scaffolding repo builds. Each repo only needs to codesign the binaries it produces.

HaoK: Cool sounds much better

PR #904, opened 2018-10-02
Fix path to signed package file

natemcmaster:
Fix broken build caused by mismatched path.


seancpeters:
Thanks, @natemcmaster

natemcmaster:
Hmm, don't merge yet. There are new issues on CI. Not sure what they are yet.

PR #924, opened 2018-11-07
Bump version to 2.1.7

natemcmaster:
Per my understanding of how VS picks the version packages to add as a PackageReference, this is still required everytime we patch aspnetcore 2.1.&, right @vijayrkn ?

### DIFF
--- a/version.props
+++ b/version.props
@@ -1,9 +1,9 @@
 ﻿<Project>
   <PropertyGroup>
-    <VersionPrefix>2.1.6</VersionPrefix>
-    <VersionSuffix>rtm</VersionSuffix>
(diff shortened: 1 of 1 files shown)


vijayrkn:
@seancpeters can correct me if I am wrong. 

Scaffolding looks at a json file to determine which all package versions to install for a given version of .NET core app. so, for Core app 2.1.7 verson,  if the json file list scaffolding package version 2.1.6, then it should ideally work provided the package closure makes sense. 

If we dont have an entry for 2.1.7, then it probably tries to match the version of scaffolding package and .NET core app package. Not sure if this logic is still there. 

@seancpeters ?

natemcmaster:
So, VS uses the package version of Microsoft.AspNetCore.App, regardless of whether it’s implicit or not? In that case, yes we need this.

seancpeters:
Correct, it doesn't matter whether AspNetCore.app is implicitly or explicitly referenced.

Issue #932, opened 2018-12-08
dotnet tool aspnet-codegenerator looks for executable at wrong path


tbureck:
I'm currently following [this guide](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/scaffold-identity?view=aspnetcore-2.1&tabs=netcore-cli#scaffold-identity-into-an-empty-project) in order to scaffold Identity in my existing ASP.NET Core project. When trying to run the scaffolder command with the help option, though, I'm getting this error message:

```
> dotnet aspnet-codegenerator --help
The application to execute does not exist: 'C:\Users\Tim\.dotnet\tools\.store\dotnet-aspnet-codegenerator\2.2.0-rtm-35687\dotnet-aspnet-codegener
ator\2.2.0-rtm-35687\tools\netcoreapp2.1\any\dotnet-aspnet-codegenerator.dll'.
```

I've checked the .dotnet\tools\.store directory and the tool seems to be installed correctly. BUT, it is installed in a `2.2.0` directory instead of the `2.2.0-rtm-35687` directory, which dotnet tool is looking in.

I've opened [a thread on Stackoverflow](https://stackoverflow.com/questions/53676499/dotnet-tool-aspnet-codegenerator-looks-for-executable-at-wrong-path) as well and someone else seems to have the same issue, which is why I'm creating this issue.

Why is it that `dotnet tool` is looking in the `2.2.0-rtm-35687` sub directory instead of the `2.2.0` where it is actually installed?

Edit: my dotnet version is `2.2.100`

maliming:
I also encountered the same problem.

seancpeters:
This issue was moved to dotnet/cli#10459

tbureck:
Since this was fixed in context of the issues linked above, I'm gonna close this issue. Thanks to everyone who has helped along the way! 💪

PR #934, opened 2018-12-11
Set VersionSuffix to empty instead of use prefix

wli3:


vijayrkn:
Fix for https://github.com/aspnet/Scaffolding/issues/932

vivmishra:
Approved for 2.2.1

seancpeters:
@wli3 - could you give a brief description of the intended effects of this change? Thanks!

version.props
```diff
@@ -38,12 +38,10 @@
 
     <!-- Run the build with /p:IsFinalBuild=true to produce the product with 'final' branding and versioning -->
     <IsFinalBuild Condition=" '$(IsFinalBuild)' == '' ">false</IsFinalBuild>
-    <IncludePreReleaseLabelInPackageVersion>true</IncludePreReleaseLabelInPackageVersion>
-    <IncludePreReleaseLabelInPackageVersion Condition=" '$(IsFinalBuild)' == 'true' AND ('$(PreReleaseLabel)' == 'servicing' OR '$(PreReleaseLabel)' == 'rtm')">false</IncludePreReleaseLabelInPackageVersion>
+    <IncludePreReleaseLabelInVersion>true</IncludePreReleaseLabelInVersion>
```
natemcmaster: Can you keep this variable as-is? It's used in other places.

version.props
```diff
@@ -38,12 +38,10 @@
...
+    <IncludePreReleaseLabelInVersion Condition=" '$(IsFinalBuild)' == 'true' AND ('$(PreReleaseLabel)' == 'servicing' OR '$(PreReleaseLabel)' == 'rtm')">false</IncludePreReleaseLabelInVersion>
 
-    <!-- The version in files -->
-    <PackageVersion>$(VersionPrefix)</PackageVersion>
-    <PackageVersion Condition=" '$(IncludePreReleaseLabelInPackageVersion)' == 'true' ">$(VersionPrefix)-$(VersionSuffix)</PackageVersion>
+    <VersionSuffix Condition=" '$(IncludePreReleaseLabelInVersion)' == 'false' "></VersionSuffix>
```
natemcmaster: I think a simpler fix would be to leave the file as is and just add 
```xml
<Version>$(PackageVersion)</Version>
```
VersionPrefix/VersionSuffix end up in other places, like informational version attributes, which is useful for figuring out the exact build number used to produce our packages and binaries.

vijayrkn:
Based on the conversation with Nate, the simplest change would be to add this to the CodeGenerator csproj so that the change is isolated to just this scenario.

`<Version>$(PackageVersion)</Version>`

wli3:
@natemcmaster I reset my change and only add the property change to csporj. I tested with `build.cmd /p:IsFinalBuild=true` and then global tool install locally to ensure no repro

vijayrkn:
Can we merge this change and get the PR ready to update the submodule?

seancpeters:
I want to do one more quick test. After that, if no one else has concerns, I’ll merge it and get the submodules PR setup.

PR #988, opened 2019-05-06
preview5 coherent package versions for breaking ef change

seancpeters:
Update the packages to build scaffolding to 3.0.0-preview5 versions.
Minor fixes for breaking changes in EF.


src/VS.Web.CG.EFCore/NavigationMetadata.cs
```diff
@@ -21,8 +21,9 @@ public NavigationMetadata(INavigation navigation, Type dbContextType)
             Contract.Assert(navigation.IsDependentToPrincipal());
 
             AssociationPropertyName = navigation.Name;
-
+#pragma warning disable EF1001 // Internal EF Core API usage.
             var otherEntityType = navigation.ForeignKey.ResolveOtherEntityType(navigation.DeclaringEntityType);
```
ajcvickers: @AndriySvyryd Is there a good alternative to using this internal API?

test/VS.Web.CG.EFCore.Test/TestModels/TestModel.cs
```diff
@@ -19,7 +19,9 @@ public static IModel CategoryProductModel
         {
             get
             {
+#pragma warning disable EF1001 // Internal EF Core API usage.
                 var builder = new ModelBuilder(SqlServerConventionSetBuilder.Build());
```
ajcvickers: What is the EF internal code that is being used here? As far as I can tell, everything here is now public.

seancpeters: SqlServerConventionSetBuilder.Build() is marked as [EntityFrameworkInternal]

ajcvickers: Ah, found the issue. These are public in master, but that that hasn't propagated to the release yet. So you should be able to remove these pragmas before RTM.

src/VS.Web.CG.EFCore/NavigationMetadata.cs
AndriySvyryd: No, I'll make it public in the next preview.

AndriySvyryd: Actually, in this case `navigation.GetTargetType()` can be used instead.

PR #1001, opened 2019-06-10
fix: bad create link

csutorasr:
The link is generated is not working most of the time. This should fix the problem, which was probably a typo.


vijayrkn:
@pranavkm - Can you please review this also?

pranavkm:
This doesn't look right. `asp-action` is the name of an action, not a path.

PR #1005, opened 2019-06-13
Move RazorPageGenerator SelectLists from GET handler to OnPageHandlerExecutionAsync

AtOMiCNebula:
`SelectList` initializers are only being code-generated into the GET handlers, but not the POST handlers.  Since `.ViewData` entries do not persist across responses, the page will not render correctly when the POST handler does not redirect (i.e. because validation failed).

This change moves the initialization into a shared `OnPageHandlerExecutionAsync`, which is invoked after model binding but before the handler itself.  No need to dual-generate the initializers into both handlers!

### DIFF
--- a/src/VS.Web.CG.Mvc/Templates/RazorPageGenerator/Bootstrap3/CreatePageModel.cshtml
+++ b/src/VS.Web.CG.Mvc/Templates/RazorPageGenerator/Bootstrap3/CreatePageModel.cshtml
@@ -40,12 +40,6 @@ namespace @Model.NamespaceName
 
         public IActionResult OnGet()
         {
-@{
-    foreach (var property in relatedProperties.Values)
-    {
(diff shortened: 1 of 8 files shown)


seancpeters:
@danroth27 @mkArtakMSFT - this is content fixes for the razor page scaffolding generated content.

AtOMiCNebula:
FYI I've been unable to run the `E2E_Test.RazorPageScaffolderTests` collection of tests locally, and the ones I expect would consume my updated reference `.txt`s appear to be being skipped in the Pipelines that get kicked off.  I believe the `.txt`s are correct, but still would like the chance to actually get the tests to pass before completing this.

Alternatively, I'll get the `.nupkg` installed locally and test the template generation by hand.  Still hoping to get the tests themselves working first...I have an email going with Sean who suggested some things for me to try to get the local tests unblocked.

PR #1010, opened 2019-06-18
Updating CodeAnalysis package versions to deal with newer CSharp versions

seancpeters:


build/dependencies.props
```diff
@@ -17,13 +17,13 @@
     <MicrosoftBuildUtilitiesCorePackageVersion>15.6.82</MicrosoftBuildUtilitiesCorePackageVersion>
 
     <!-- Microsoft.CodeAnalysis.CSharp -->
```
vijayrkn: I am glad we invested in documenting the package names

PR #1026, opened 2019-07-08
remove please from QR template

Rick-Anderson:
Per MS style guide:
Avoid please except in situations where the customer is asked to do something inconvenient or the application or site is to blame for the situation.


Rick-Anderson:
Replaced by #1027

PR #1030, opened 2019-07-15
Fixed missing .AddDefaultUI() in Identity template

CamSoper:
When scaffolding ASP.NET Identity, there's an issue where the  --useDefaultUI flag doesn't actually include the `.AddDefaultUI()` method call in *IdentityHostingStartup.cs*.

This change checks the `AddDefaultUI` flag on the model and, if true, adds the required `.AddDefaultUI()` method call.

Related to https://github.com/aspnet/Scaffolding/issues/885, but doesn't resolve the original issue.

cc: @scottaddie

### DIFF
--- a/src/VS.Web.CG.Mvc/Templates/Identity/Bootstrap4/IdentityHostingStartup.cshtml
+++ b/src/VS.Web.CG.Mvc/Templates/Identity/Bootstrap4/IdentityHostingStartup.cshtml
@@ -62,7 +62,7 @@ namespace @(thisNamespace)
 @:                services.AddDefaultIdentity<@(Model.UserClass)>()
         if (Model.UseDefaultUI)
         {
-@:                    .AddDefaultUI(UIFramework.Bootstrap4);
+@:                    .AddDefaultUI(UIFramework.Bootstrap4)
         }
 @:                    .AddEntityFrameworkStores<@Model.DbContextClass>();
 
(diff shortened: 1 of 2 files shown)


vijayrkn:
Adding @HaoK since this is a change to the identity content.

src/VS.Web.CG.Mvc/Templates/Identity/Bootstrap4/IdentityHostingStartup.cshtml
```diff
@@ -60,6 +60,10 @@ namespace @(thisNamespace)
 @:
 
 @:                services.AddDefaultIdentity<@(Model.UserClass)>()
+        if (Model.UseDefaultUI)
+        {
+@:                    .AddDefaultUI(UIFramework.Bootstrap4)
```
HaoK: Is this one really needed since Bootstrap4 is the default?

CamSoper: .AddDefaultUI() still needs to be present in the generated file or the correct layout never gets loaded and the rendered visuals are broken. Visual Studio's templates add this line in every case. It appears to be an oversight in the dotnet CLI scaffolding.

HaoK: Is this due to something on the VS/scaffolding side?  Because on the runtime side AddDefaultUI is called by `AddDefaultIdentity()` anyways

scottaddie: @HaoK The UI in our 2.2 Razor Pages app wasn't styled correctly unless the `AddDefaultUI` method call was chained on to `AddDefaultIdentity`.

CamSoper: @HaoK -- As @scottaddie mentioned, this was an issue in 2.2. Looking at it now in 3.0, it appears to no longer be an issue. Closing this PR.

vijayrkn: Thanks for getting to the bottom of this. 
@CamSoper Appreciate you help with this & we look forward to more contributions from your side.

PR #1056, opened 2019-08-28
Applying some best practices to scaffolding

ryanbrandenburg:


HaoK:
Should someone on the MVC team sign off on these changes, I'm not all that familiar with the model binding best practices?

mkArtakMSFT:
@ryanbrandenburg can you please add a comment here, showing what the final generated code by this scaffolder would look like. That would help out with the review.

PR #1061, opened 2019-09-01
Update jquery to 3.4.1

HaoK:


mkArtakMSFT:
Marking this as `blocked` to avoid accidental merge before the decision about which release this will go in.

PR #1064, opened 2019-09-03
[2.1] Update jquery to 3.4.1

HaoK:
Note there is no Bootstrapv4 in 2.1


mkArtakMSFT:
Marking as `blocked` as the decision about which release this should go into hasn't yet been made.

PR #1066, opened 2019-09-05
Remove cookie consent and cdn from script tags

HaoK:
@vijayrkn it looks like there might also need to be corresponding changes in https://github.com/aspnet/Scaffolding/blob/167a8b5c626db13896e1e4fad65b5695be37da20/src/VS.Web.CG.Mvc/Identity/IdentityGeneratorFilesConfig.cs#L284

Fixes https://github.com/aspnet/AspNetCore/issues/13699

### DIFF
--- a/src/VS.Web.CG.Mvc/Templates/Identity/Bootstrap3/Pages/_Layout.cshtml
+++ b/src/VS.Web.CG.Mvc/Templates/Identity/Bootstrap3/Pages/_Layout.cshtml
@@ -13,15 +13,7 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>@@ViewData["Title"] - @Model.ApplicationName</title>
-
-    <environment include="Development">
-        <link rel="stylesheet" href="~/Identity/lib/bootstrap/dist/css/bootstrap.css" />
-    </environment>
-    <environment exclude="Development">
-        <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/3.4.1/css/bootstrap.min.css"
-              asp-fallback-href="~/Identity/lib/bootstrap/dist/css/bootstrap.min.css"
-              asp-fallback-test-class="sr-only" asp-fallback-test-property="position" asp-fallback-test-value="absolute" />
-    </environment>
+    <link rel="stylesheet" href="~/Identity/lib/bootstrap/dist/css/bootstrap.css" />
     <link rel="stylesheet" href="~/Identity/css/site.css" />
 </head>
 <body>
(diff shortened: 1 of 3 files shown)


HaoK:
Updated to also remove cdn, cc @danroth27 for review of that

vijayrkn:
@HaoK Is this change required for 3.0? if so, we need qb approval before merging.

HaoK:
@danroth27 is this something we need for 3.0?

src/VS.Web.CG.Mvc/Templates/Identity/Bootstrap3/Pages/_Layout.cshtml
```diff
@@ -13,15 +13,7 @@
...
-
-    <environment include="Development">
-        <link rel="stylesheet" href="~/Identity/lib/bootstrap/dist/css/bootstrap.css" />
-    </environment>
-    <environment exclude="Development">
-        <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/3.4.1/css/bootstrap.min.css"
-              asp-fallback-href="~/Identity/lib/bootstrap/dist/css/bootstrap.min.css"
-              asp-fallback-test-class="sr-only" asp-fallback-test-property="position" asp-fallback-test-value="absolute" />
-    </environment>
+    <link rel="stylesheet" href="~/Identity/lib/bootstrap/dist/css/bootstrap.css" />
```
danroth27: These should be the minified files. Compare with the templates in release/3.0.

HaoK: Updated

danroth27:
@HaoK @vijayrkn Yes, we should get this fixed for 16.3.

danroth27:
@vijayrkn What needs to be done to ensure that *_CookieConsentPartial.cshtml* is no longer generated by the Identity scaffolder?

vijayrkn:
As @HaoK noted, there is probably corresponding changes needed here - https://github.com/aspnet/Scaffolding/blob/167a8b5c626db13896e1e4fad65b5695be37da20/src/VS.Web.CG.Mvc/Identity/IdentityGeneratorFilesConfig.cs#L284

vijayrkn:
@HaoK - it is probably just removing these 3 lines. Do we need to remove this file for all cases?  

https://github.com/aspnet/Scaffolding/blob/61114a665ee506804c3a648cb88b8ebc11ae9819/src/VS.Web.CG.Mvc/Identity/IdentityGeneratorTemplateModelBuilder.cs#L242-L245

HaoK:
Updated removing the 3 lines to stop generating the cookie consent page if it doesn't exist, I tried it on a private build and it appears to work (view no longer gets scaffolded)

PR #1077, opened 2019-09-17
Work around razor bug in layout

HaoK:
Fixes https://github.com/aspnet/AspNetCore/issues/14064

### DIFF
--- a/src/VS.Web.CG.Mvc/Templates/Identity/Bootstrap3/Pages/_Layout.cshtml
+++ b/src/VS.Web.CG.Mvc/Templates/Identity/Bootstrap3/Pages/_Layout.cshtml
@@ -3,10 +3,12 @@
     var currentYear = System.DateTime.Now.Year;
 }
 
-@@using Microsoft.AspNetCore.Hosting
-@@using Microsoft.AspNetCore.Mvc.ViewEngines
-@@inject IWebHostEnvironment Environment
-@@inject ICompositeViewEngine Engine
+@{
+@:@@using Microsoft.AspNetCore.Hosting
+@:@@using Microsoft.AspNetCore.Mvc.ViewEngines
+@:@@inject IWebHostEnvironment Environment
+@:@@inject ICompositeViewEngine Engine
+}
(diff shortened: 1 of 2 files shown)


mkArtakMSFT:
We want this fix for 3.1. Shouldn't this need to target a 3.1 specific branch?

mkArtakMSFT:
Also, I may be confused by the working here, but I assume this is the proper fix, and not simply a work around, isn't it?

vijayrkn:
For 3.1 - it should be cherry-picked.

mkArtakMSFT:
Based on https://github.com/aspnet/AspNetCore/issues/14064#issuecomment-532420102 it sounds like this is indeed a work around. @NTaylorMullen what is the bug being referred here?

HaoK:
The root bug is https://github.com/aspnet/AspNetCore/issues/10811

mkArtakMSFT:
@NTaylorMullen I prefer we fix the root cause, rather than making this change. Given it's `S` let's just do it now. I'm bringing the root-cause issue to 3.1 now. Any objections?

NTaylorMullen:
> @NTaylorMullen I prefer we fix the root cause, rather than making this change. Given it's `S` let's just do it now. I'm bringing the root-cause issue to 3.1 now. Any objections?

It's a super edge case so i'm not convinced it's worth the time. I think it makes more sense just to special case this here given it's already done.

mkArtakMSFT:
Aren't there any other side effects similar to this, caused by the actual root cause?

NTaylorMullen:
> Aren't there any other side effects similar to this, caused by the actual root cause?

Nope

mkArtakMSFT:
Let's move forward with this change then, @HaoK!

HaoK:
What's the 3.1 branch to cherry pick to?  I only see 3.0 and 3.0-preview9 I think

vijayrkn:
https://github.com/aspnet/Scaffolding/tree/release/3.1

Just created one off release/3.0

PR #1096, opened 2019-10-08
props updates for 3.1.0

deepchoudhery:
- had to update Microsoft.CodeAnalysis.CSharp.Workspaces to 3.4.0-beta2 as that dependency was required. 
image

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -69,7 +69,7 @@
     <MicrosoftAspNetCoreIdentityUIPackageVersion>3.0.0</MicrosoftAspNetCoreIdentityUIPackageVersion>
 
     <!-- Microsoft.Extensions.Identity.Stores -->
-    <MicrosoftExtensionsIdentityStoresPackageVersion>3.0.0</MicrosoftExtensionsIdentityStoresPackageVersion>
+    <MicrosoftExtensionsIdentityStoresPackageVersion>3.1.0-preview1.19506.2</MicrosoftExtensionsIdentityStoresPackageVersion>
 
     <!-- Microsoft.AspNetCore -->
     <!-- this is from aspnetcore-dev (test project only) -->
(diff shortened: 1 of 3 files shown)


version.props
```diff
@@ -1,9 +1,9 @@
 ﻿<Project>
   <PropertyGroup>
-    <MajorVersion>5</MajorVersion>
-    <MinorVersion>0</MinorVersion>
+    <MajorVersion>3</MajorVersion>
+    <MinorVersion>1</MinorVersion>
     <PatchVersion>0</PatchVersion>
-    <PreReleaseLabel>alpha1</PreReleaseLabel>
+    <PreReleaseLabel>rtm</PreReleaseLabel>
```
vijayrkn: This should be preview1

build/dependencies.props
```diff
@@ -99,7 +99,7 @@
     <MicrosoftEntityFrameworkCoreSqlServerPackageVersion>3.0.0-preview5.19227.1</MicrosoftEntityFrameworkCoreSqlServerPackageVersion>
 
     <!-- Microsoft.Extensions.Configuration.Abstractions -->
-    <MicrosoftExtensionsConfigurationAbstractionsPackageVersion>3.0.0</MicrosoftExtensionsConfigurationAbstractionsPackageVersion>
+    <MicrosoftExtensionsConfigurationAbstractionsPackageVersion>3.1.0-preview1.19506.1</MicrosoftExtensionsConfigurationAbstractionsPackageVersion>
 
     <!-- Microsoft.Extensions.Configuration.EnvironmentVariables -->
     <MicrosoftExtensionsConfigurationEnvironmentVariablesPackageVersion>3.0.0</MicrosoftExtensionsConfigurationEnvironmentVariablesPackageVersion>
```
vijayrkn: How about these extensions packages?

deepchoudhery: missed "Microsoft.Extensions.Identity.Stores" but only one I could find in the new feed.

version.props
deepchoudhery: fixed now

build/dependencies.props
vijayrkn: @JunTaoLuo - Is there plan to generate 3.1.0 versions for the extensions packages?

PR #1097, opened 2019-10-09
updated the feed file

deepchoudhery:


build/sources.props
```diff
@@ -8,7 +8,7 @@
       https://dotnet.myget.org/F/dotnet-core/api/v3/index.json;
       https://dotnet.myget.org/F/aspnetcore-dev/api/v3/index.json;
       https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json;
-      https://dotnetfeed.blob.core.windows.net/dotnet-core-3-1-preview1-014455/index.json
+      https://dotnetfeed.blob.core.windows.net/dotnet-core-3-1-preview1-014459/index.json
```
vijayrkn: are the versions still the same?

deepchoudhery: file wasn't saved, my bad. bunch of updates including extensions that were not before. local builds fine. will test the artifacts from this build

PR #1099, opened 2019-10-10
3.1 cleanup resend email

HaoK:
Scaffolding side matching fixes to https://github.com/aspnet/AspNetCore/pull/14901


vijayrkn:
If this is for 3.1, then this should be cherry-picked to the 3.1 branch. 

/cc @deepchoudhery

PR #1109, opened 2019-10-12
Update Resources.resx

Rick-Anderson:
Per MS style guide, 
> Avoid please except in situations where the customer is asked to do something inconvenient or the application or site is to blame for the situation.


Rick-Anderson:
#1120

PR #1124, opened 2019-10-30
updated arcade version manually

deepchoudhery:


vijayrkn:
@deepchoudhery - Can you check why the linux builds are failing.

PR #1132, opened 2019-11-01
Update 3.1.0-preview1 to 2

deepchoudhery:


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "3.1.100-preview1-014459"
+    "dotnet": "3.1.100-preview2-014569"
```
vijayrkn: As discussed, this need not change.

eng/Versions.props
```diff
@@ -16,7 +16,8 @@
 
   <PropertyGroup>
     <VersionPrefix>3.1.0</VersionPrefix>
-    <PreReleaseVersionLabel>rc1</PreReleaseVersionLabel>
+    <PreReleaseVersionLabel>preview</PreReleaseVersionLabel>
```
vijayrkn: @deepchoudhery - You should update the PreReleaseVersionLabel to Preview2 and remove the PreReleaseVersionIteration

PR #1138, opened 2019-11-03
Setting the Scaffolding design package as a developmentDependency

vijayrkn:
@deepchoudhery - This needs a corresponding change in VS before this can be merged.


vijayrkn:
@davidfowl - do you know if we can set something in the nuspec along with DevelopmentDependency to true in order to include 'compile' also as part of includeAssets? 

This is needed because Scaffolding needs to run the razor templates to generate the content. 

```
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive; compile</IncludeAssets>
```

PR #1158, opened 2019-11-14
Update 3.1 to GA

deepchoudhery:
updated apt packages.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -14,7 +14,7 @@
   </PropertyGroup>
   <PropertyGroup>
     <VersionPrefix>3.1.0</VersionPrefix>
-    <PreReleaseVersionLabel></PreReleaseVersionLabel>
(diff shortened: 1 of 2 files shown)


eng/Versions.props
```diff
@@ -14,7 +14,7 @@
   </PropertyGroup>
   <PropertyGroup>
     <VersionPrefix>3.1.0</VersionPrefix>
-    <PreReleaseVersionLabel>preview3</PreReleaseVersionLabel>
+    <PreReleaseVersionLabel></PreReleaseVersionLabel>
```
vijayrkn: You need to add an additional property here 

    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">true</StabilizePackageVersion>

vijayrkn: And PreReleaseVersionLabel should be rtm

PR #1164, opened 2019-11-15
Merging release/3.1 based branch into master

deepchoudhery:
Deleted additional eng files that were added to release/3.1.


eng/Versions.props
```diff
@@ -15,6 +15,7 @@
   <PropertyGroup>
     <VersionPrefix>5.0.0</VersionPrefix>
     <PreReleaseVersionLabel>alpha1</PreReleaseVersionLabel>
+    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">true</StabilizePackageVersion>
```
vijayrkn: This should be false since master is not stabilizing.

PR #1183, opened 2019-12-02
Fix ApiControllerWithActions Delete API comment

rabbitism:
It should use controller name instead of fixed string "ApiWithActions"


deepchoudhery:
thanks

PR #1215, opened 2020-01-03
Fix scaffolding file generation issues

deepchoudhery:
Scaffolding fails during file generation due to missing parens in this template file.


src/VS.Web.CG.Mvc/Templates/Identity/Bootstrap3/Pages/Account/Account.ConfirmEmailChange.cs.cshtml
```diff
@@ -34,9 +34,9 @@ namespace @thisNamespace
     public class ConfirmEmailChangeModel : PageModel
     {
         private readonly UserManager<@(Model.UserClass)> _userManager;
-        private readonly SignInManager<@(Model.UserClass> _signInManager;
+        private readonly SignInManager<@(Model.UserClass)> _signInManager;
```
vijayrkn: Can you update the version number in the branch to 3.1.1 servicing before taking this change?

deepchoudhery: should be good now

PR #1234, opened 2020-01-23
added sqlite support

deepchoudhery:
- added sqlite support for ef scaffolders
- added global tool helper for windows development
- added some tests


src/VS.Web.CG.Mvc/Common/EFValidationUtil.cs
```diff
@@ -10,13 +10,14 @@ namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc
 {
     internal static class EFValidationUtil
     {
+        const string EfDesignPackageName = "Microsoft.EntityFrameworkCore.Design";
+        const string SqlServerPackageName = "Microsoft.EntityFrameworkCore.SqlServer";
+        const string SqlitePackageName = "Microsoft.EntityFrameworkCore.Sqlite";
```
bricelam: Might be better to check for Microsoft.EntityFrameworkCore.Sqlite.**Core** (but still recommend installing Microsoft.EntityFrameworkCore.Sqlite)

bricelam: The main package depends on .Core and another package to pull in the native SQLite binaries:

* Project
  * Microsoft.Data.Sqlite
    * Microsoft.Data.Sqlite.Core
    * SQLitePCLRaw.bundle_e_sqlite

But you can use other SQLite binaries (like SQLCipher for encryption) by depending directly on .Core and a different bundle:

* Project
  * Microsoft.Data.Sqlite.Core
  * SQLitePCLRaw.bundle_e_sqlcipher

deepchoudhery: gotcha. so really Sqlite.Core and SQlitePCLRaw.bundle_e_sqlite should be installed as well when we wanna use sqlite

deepchoudhery: is checking for SQLitePCLRaw.bundle_e_sqlite or SQLitePCLRaw.bundle_e_sqlcipher necessary or do we leave that up to the user at that point?

src/VS.Web.CG.Mvc/Common/EFValidationUtil.cs
```diff
@@ -26,5 +29,27 @@ internal static void ValidateEFDependencies(IEnumerable<DependencyDescription> d
                     string.Format(MessageStrings.InstallEfPackages, $"{EfDesignPackageName}, {SqlServerPackageName}"));
             }
         }
+
+        internal static void ValidateSQLiteDependency(IEnumerable<DependencyDescription> dependencies)
+        { 
+            var isSqlitePackagePresent = dependencies
```
vijayrkn: Update per the message from Brice

src/VS.Web.CG.Mvc/Common/EFValidationUtil.cs
bricelam: Leave it up to the user. The code will compile just fine without it

PR #1243, opened 2020-01-30
Add new resend email confirmation

HaoK:
@deepchoudhery @vijayrkn I added two new files and updated the json file but I don't see them showing up in the dialog, did I miss updating something?


HaoK:
Oh never mind, I nuked them when I did git -xdff

PR #1255, opened 2020-02-13
Login page shows by address when already loggined

zandiarash:
If the user enter the login URL, the login page shows even when the user is already logged in.
I've changed to avoid this


vijayrkn:
@HaoK - Can you please review?

HaoK:
We wouldn't take any fixes in scaffolding first, @zandiarash please file an issue in https://github.com/dotnet/aspnetcore describing what you are encountering

PR #1264, opened 2020-02-20
Fix missing input

HaoK:
@deepchoudhery this should fix the compilation issues, thanks!


deepchoudhery:
Tested in VS and CLI. Works now.

PR #1299, opened 2020-03-25
sql package handling fix

deepchoudhery:
sqlserver is not mandatory if using sqlite now.


src/VS.Web.CG.Mvc/MessageStrings.resx
```diff
@@ -249,4 +249,7 @@
   <data name="InvalidBootstrapVersionForScaffolding" xml:space="preserve">
     <value>The specified bootstrap version '{0}' is not valid. The valid options are '{1}'.</value>
   </data>
+  <data name="InstallSqlPackages" xml:space="preserve">
+    <value>To scaffold controllers and views using models, install one of the Entity Framework core packages and try again: {0}</value>
```
vijayrkn: It shouldnt say "install one of the Entity Framework core packages". It should mention which package to install.

deepchoudhery: Removed it since the reworked logic doesn't need it and using other more apt. resource strings.

deepchoudhery:
@Rick-Anderson - added some new error messages.

PR #1312, opened 2020-04-03
removing all other targets from the design package

deepchoudhery:
Other platform targets such as x86, arm64, win64 aren't needed as net 5 should be able to handle all other platform targets. These targets were not being used in previous versions as well.


deepchoudhery:
Closing for now. Will come back to it after more testing.

PR #1315, opened 2020-04-07
removing nuspec for CodeGeneration.Design package

deepchoudhery:
removed nuspec and just using arcade's packing. 
- all package metadata is correct.
- tested the CLI tool and with VS, no trouble finding the correct assembly since package structure was kept intact.


src/VS.Web.CG.Design/VS.Web.CG.Design.nuspec
```diff
@@ -1,38 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
-  <metadata>
```
vijayrkn, line 3: I dont see this metadata copied over. You also need to add the dependency.

deepchoudhery: All metadata is picked up from arcade. Compared the generated nuspecs and looks good.

PR #1326, opened 2020-04-27
Razor page namespace fix

deepchoudhery:
Cli tool was still using application name over root namespace as namespace prefix when not provided a namespace. VS provides one.

### DIFF
--- a/src/VS.Web.CG.Mvc/RazorPage/RazorPageScaffolderBase.cs
+++ b/src/VS.Web.CG.Mvc/RazorPage/RazorPageScaffolderBase.cs
@@ -284,7 +284,7 @@ namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Razor
 
         private string GetDefaultPageModelNamespaceName(string relativeFolderPath)
         {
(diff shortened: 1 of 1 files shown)


src/VS.Web.CG.Mvc/RazorPage/RazorPageScaffolderBase.cs
```diff
@@ -284,7 +284,7 @@ protected RazorPageWithContextTemplateModel GetRazorPageWithContextTemplateModel
 
         private string GetDefaultPageModelNamespaceName(string relativeFolderPath)
         {
-            return NameSpaceUtilities.GetSafeNameSpaceFromPath(relativeFolderPath, ApplicationInfo.ApplicationName);
+            return NameSpaceUtilities.GetSafeNameSpaceFromPath(relativeFolderPath, _projectContext.RootNamespace);
```
vijayrkn, line 287: Is this used in commandline cases & when VS does not pass the namespace?

deepchoudhery: VS should always pass a namespace but yes in cmd line cases.

PR #1341, opened 2020-05-17
Fix execute permission bit on build.sh

MaxHorstmann:
Fixes https://github.com/dotnet/Scaffolding/issues/1340.

Git [stores the execute permission bit](https://stackoverflow.com/questions/39666585/does-git-store-the-read-write-execute-permissions-for-files) on files. Needs to be set on eng/common/build.sh, otherwise unable to build on MacOS/Linux.


deepchoudhery:
Thanks.

PR #1356, opened 2020-06-22
Updates for 5.0-Preview6

deepchoudhery:
- Version.props update
- Tfm changes from project system and sdk.

### DIFF
--- a/src/dotnet-aspnet-codegenerator/DotNetBuildCommandHelper.cs
+++ b/src/dotnet-aspnet-codegenerator/DotNetBuildCommandHelper.cs
@@ -11,6 +11,7 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration.Tools
 {
     internal class DotNetBuildCommandHelper
     {
+        private const string NetFrameworkShortFolderName = "net";
         internal static BuildResult Build(
             string project,
             string configuration,
@@ -63,7 +64,7 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration.Tools
                 {
                     if (framework.Version.Major >= 5)
                     {
-                        return "net5.0";
+                        return string.Concat(NetFrameworkShortFolderName, framework.Version.Major, ".", framework.Version.Minor);
                     }
                 }
                 shortFramework = framework.GetShortFolderName();
--- a/src/dotnet-aspnet-codegenerator/DotNetBuildCommandHelper.cs
+++ b/src/dotnet-aspnet-codegenerator/DotNetBuildCommandHelper.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -21,7 +21,7 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration.Tools
             {
                 project,
                 "--configuration", configuration,
-                "--framework", framework.GetShortFolderName(),
+                "--framework", GetFrameworkShortName(framework),
             };
 
             if (buildBasePath != null)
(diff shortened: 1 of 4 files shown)


src/dotnet-aspnet-codegenerator/DotNetBuildCommandHelper.cs
```diff
@@ -54,6 +54,21 @@ internal class DotNetBuildCommandHelper
                 StdOut = stdOutMsgs
             };
         }
+
+        private static string GetFrameworkShortName(NuGetFramework framework)
+        {
+            string shortFramework = string.Empty;
+            if (framework != null)
+            {
+                if (framework.Framework.Equals(FrameworkConstants.FrameworkIdentifiers.Net) &&
```
vijayrkn, line 63: You should be checking for the frameworkidentifier

deepchoudhery: framework.Framework is the framework identifier(".NETFramework"/ ".NETCore").

src/dotnet-aspnet-codegenerator/DotNetBuildCommandHelper.cs
```diff
@@ -54,6 +54,21 @@ internal class DotNetBuildCommandHelper
...
+                    framework.Version.Major >= 5)
+                {
+                    return string.Concat(NetFrameworkShortFolderName, framework.Version.Major, ".", framework.Version.Minor);
+                }
+                shortFramework = framework.GetShortFolderName();
```
vijayrkn, line 68: Why is this not handling the short name in all cases - GetShortFolderName?

src/dotnet-aspnet-codegenerator/Program.cs
```diff
@@ -231,14 +230,13 @@ internal int Execute(string[] args, bool isNoBuild)
             var depsFile = Path.Combine(targetDir, context.DepsFile);
 
             string dotnetCodeGenInsideManPath = string.Empty;
-            
-            if (IsNetCoreAppFramework(frameworkToUse))
+            if (IsNetCoreAppFramework(frameworkToUse) || IsNet5OrAbove(frameworkToUse))
```
vijayrkn, line 233: Why separate check for IsNet5OrAbove? If you are checking for the TFI, then net5.0 & above would also return true for the .NETCore check. 

The net5 check is only needed if you have just the shortname and not the TFI

src/dotnet-aspnet-codegenerator/DotNetBuildCommandHelper.cs
deepchoudhery: For .NETFrameworks, NuGet.Frameworks.NuGetFramework.GetShortFolderName returns the version without dots. for example, NET 4.6.1 => net461 and NET 5.0 => net50. For NETCoreApp it returns the version with dots for example NET Core APp 3.1 => netcoreapp3.1. Since the short folder name for NET 5.0 should be net5.0 instead of net50, I'm handling it myself. I might not be consuming the latest Nuget.Frameworks though but can look into that.

vijayrkn: NET 5.0 should be returned under NETCoreApp. Is that not what you are seeing?

src/dotnet-aspnet-codegenerator/Program.cs
deepchoudhery: TFI for net5.0 is .NETFramework and netcoreapp is .NETCoreApp. We do something similar in WTE. Need to have checks for both separately.

vijayrkn: TFI for net5.0 should be .NETCoreApp. What we do in WTE is only if you start with the shortname. That is different. Let me know when you have some time so that we can chat in more detail on this.

src/dotnet-aspnet-codegenerator/DotNetBuildCommandHelper.cs
deepchoudhery: NET 5.0 does not return under NETCoreApp since 5.0-Preview5

deepchoudhery:
Closed. All issues were solved by updating NuGet.Frameworks package

PR #1367, opened 2020-07-15
using projects.assets.json to get project dependencies

deepchoudhery:
since the project system changes in the sdk 3.1.4+ and 5.0.0-preview7+, we need to populate dependencies using the projects.assets.json file.

- Using msbuild property to get the assets file location and then parsing the json.

- added unit tests and one to test dependencies using a bare net 5.0 project with the CodeGeneration.Design dependency.

Will port over VS fairly soon as well (until we are ready to use the Nuget API).

### DIFF
--- a/src/VS.Web.CG.Msbuild/ProjectContextWriter.cs
+++ b/src/VS.Web.CG.Msbuild/ProjectContextWriter.cs
@@ -1,19 +1,25 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Text.Json;
 using Microsoft.Build.Framework;
 using Microsoft.VisualStudio.Web.CodeGeneration.Contracts.ProjectModel;
 using Microsoft.VisualStudio.Web.CodeGeneration.Utils;
-using Newtonsoft.Json;
+using NuGet.Frameworks;
 
 namespace Microsoft.VisualStudio.Web.CodeGeneration.Msbuild
 {
     public class ProjectContextWriter : Build.Utilities.Task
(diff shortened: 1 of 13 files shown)


src/VS.Web.CG.Msbuild/ProjectContextWriter.cs
```diff
@@ -98,91 +105,125 @@ public override bool Execute()
...
 
-        private IEnumerable<ProjectReferenceInformation> GetProjectDependency(
-            IEnumerable<string> projectReferences,
-            string rootProjectFullpath)
+        private IEnumerable<DependencyDescription> GetPackageDependencies(string projectAssetsFile)
         {
-            return ProjectReferenceInformationProvider.GetProjectReferenceInformation(
-                rootProjectFullpath,
-                projectReferences);
+            IList<DependencyDescription> packageDependencies = new List<DependencyDescription>();
```
vijayrkn, line 117: This will be replaced with the NuGet api when ready correct?

deepchoudhery: I'm still thinking about it since it still does not give me all the information I need. I would still be dependent on project.assets.json for transitive dependencies so for now, just taking a full dependency on that file. They have not removed much from that file the nuget team says. I will add a couple more tests before shipping this so it'll be easy to react to changes as the sdk evolves. Hopefully the Nuget API catches up in a couple more releases before we go GA.

jahnog:
Hi, it seems that this change generated several issues involving the aspnet-codegenerator. Basically it fails to scaffold files when the .nuget cache folder resides in a Case Sensitive filesystem (Linux & Mac):
#1393 and #1384 are just two of the issues.
There is a hot fix proposed by several users (mounting the .nuget folder in a Case Insensitive filesystem like vfat) but it is by no means a final solution: [https://github.com/dotnet/Scaffolding/issues/1393#issuecomment-752041418](https://github.com/dotnet/Scaffolding/issues/1393#issuecomment-752041418)

deepchoudhery:
@jahnog Yup I'll get a fix in ASAP.

PR #1372, opened 2020-07-22
Fix formatting

Youssef1313:
I'm working on an ASP.NET Core project, and after I've added scaffolding for register and login pages, the default code inside `IdentityHostingStartup` gives IDE0055 warning. I hope this is the correct place to fix this.

image

I've also a question: in my case, the `IdentityHostingStartup` doesn't seem to be doing anything, why it was generated in the first place? I think it should only be generated if there is no existing db context?


Youssef1313:
Let me know if you want the changes here and I'll resolve the conflicts.

vijayrkn:
The dotnet repo's does not accept just the formatting changes. Thank for your contribution.

PR #1375, opened 2020-07-24
Delete ResendEmailConfirmation from 3.1

Youssef1313:
It was removed in #1339, but it didn't target version 3.1.
Because 3.1 is a LTS, I think it's reasonable to add this fix to it.
This PR is just the same as #1339 but targets the 3.1 branch, so the fix will be included in the next minor version of 3.1

Fixes https://github.com/dotnet/aspnetcore/issues/24273


Youssef1313:
Based on the discussion with @HaoK in https://github.com/dotnet/aspnetcore/issues/24273, I updated the PR to completely delete the ResendEmailConfirmation page and its link, instead of removing the abstract keyword.

deepchoudhery:
Sounds good.

PR #1378, opened 2020-07-31
fixing path for dependencies

deepchoudhery:


deepchoudhery:
Minor change so just getting it in.

PR #1380, opened 2020-08-03
removing Nuget.Frameworks dependency

deepchoudhery:
removing Nuget.Frameworks dependency.
- only use for to get TargetFrameworkMoniker from short folder name.
- can get both properties from design time build


src/dotnet-aspnet-codegenerator/DotNetBuildCommandHelper.cs
```diff
@@ -40,9 +39,7 @@ internal class DotNetBuildCommandHelper
 
             var command = Command.CreateDotNet(
                     "build",
-                    args,
-                    framework,
-                    configuration: configuration)
```
vijayrkn, line 45: why remove the configuration?

deepchoudhery: it was not being used. couple more minor edits like this.

PR #1413, opened 2020-09-11
correcting identity help text (#1409)

deepchoudhery:


deepchoudhery:
Cherry-picking from master.

PR #1421, opened 2020-09-25
replaced check for referencing assemblies

deepchoudhery:
Tested against VS and CLI projects. 
- allows us to remove some unnecessary dependency code on VS.


deepchoudhery:
Closing. Will be covered in a different PR

PR #1548, opened 2021-04-12
dotnet-msidentity Preview 1

deepchoudhery:
Preview 1 version of the tool to be used on the command line and with VS 16.10 Preview 2.

--update-project command --client-id [--calls-graph] [--calls-downstream-api] [--update-user-secrets]
  - updates appsettings.json
  - creates secret, adds to azure ad app, updates local user-secrets

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -46,7 +46,7 @@
     <MicrosoftExtensionsDependencyInjectionPackageVersion>6.0.0-preview.3.21201.4</MicrosoftExtensionsDependencyInjectionPackageVersion>
     <!-- Microsoft.Extensions.FileProviders.Physical -->
     <MicrosoftExtensionsFileProvidersPhysicalPackageVersion>6.0.0-preview.3.21201.4</MicrosoftExtensionsFileProvidersPhysicalPackageVersion>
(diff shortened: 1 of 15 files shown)


eng/Versions.props
```diff
@@ -46,7 +46,7 @@
     <MicrosoftExtensionsDependencyInjectionPackageVersion>6.0.0-preview.3.21201.4</MicrosoftExtensionsDependencyInjectionPackageVersion>
     <!-- Microsoft.Extensions.FileProviders.Physical -->
     <MicrosoftExtensionsFileProvidersPhysicalPackageVersion>6.0.0-preview.3.21201.4</MicrosoftExtensionsFileProvidersPhysicalPackageVersion>
-    <NewtonsoftJsonPackageVersion>11.0.2</NewtonsoftJsonPackageVersion>
+    <NewtonsoftJsonPackageVersion>13.0.1</NewtonsoftJsonPackageVersion>
```
vijayrkn, line 49: Is this changing for scaffolding as well? That might have impact if VS is shipping another version. Can we confine this just to Identity?

deepchoudhery: yup it is used by scaffolding a bunch. Reverted it for now but adding ticket to track the upgrade for next preview. 
#1550

PR #1565, opened 2021-04-29
finding all aad apps through NextPageRequest on GraphServiceClient

deepchoudhery:
- Only the first 50 (sorted by creation time) AAD apps and Service Principals were being returned in the --list-* commands.
- Fixed to return all available apps and service principals.

### DIFF
--- a/src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Tool/MsAADTool.cs
+++ b/src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Tool/MsAADTool.cs
@@ -85,9 +85,14 @@ namespace Microsoft.DotNet.MSIdentity.Tool
                                 graphObjectsList.AddRange(additionalGraphObjects.ToList());
                                 nextPage = additionalGraphObjects.NextPageRequest;
                             }
+                            else
+                            {
+                                nextPage = null;
+                            }
                         }
                         catch (ServiceException)
                         {
+                            nextPage = null;
                             ConsoleLogger.LogMessage("Failed to retrieve all Azure AD/AD B2C objects(apps/service principals", LogMessageType.Error);
                         }
                     }
--- a/src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Tool/MsAADTool.cs
+++ b/src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Tool/MsAADTool.cs
@@ -67,42 +67,58 @@ namespace Microsoft.DotNet.MSIdentity.Tool
             var graphObjects = await GraphServiceClient.Me.OwnedObjects
                             .Request()
                             .GetAsync();
-            IList<Application> applicationList = new List<Application>();
-            if (graphObjects != null && graphObjects.Any())
+
+            if (graphObjects != null)
             {
-                foreach (var graphObj in graphObjects)
+                var graphObjectsList = graphObjects.ToList();
+
+                var nextPage = graphObjects.NextPageRequest;
+                while (nextPage != null)
                 {
-                    if (graphObj is Application app)
-                    {
-                        {
-                            applicationList.Add(app);
-                        }
-                    }
+                    var additionalGraphObjects = await nextPage.GetAsync();
+                    graphObjectsList.AddRange(additionalGraphObjects.ToList());
+                    nextPage = additionalGraphObjects.NextPageRequest;
                 }
 
-                if (applicationList.Any())
-                {
-                    //order list by created date.
-                    applicationList = applicationList.OrderByDescending(app => app.CreatedDateTime).ToList();
+                IList<Application> applicationList = new List<Application>();
 
-                    if (ProvisioningToolOptions.Json)
+                if (graphObjectsList != null && graphObjectsList.Any())
+                {
+                    foreach (var graphObj in graphObjectsList)
                     {
-                        JsonResponse jsonResponse = new JsonResponse(CommandName, State.Success, applicationList);
-                        outputJsonString = jsonResponse.ToJsonString(); 
-                    }
-                    else 
-                    {
-                        Console.Write(
-                            "--------------------------------------------------------------\n" + 
-                            "Application Name\t\t\t\tApplication ID\n" +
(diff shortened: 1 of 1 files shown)


zahalzel (approved):

src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Tool/MsAADTool.cs
```diff
@@ -67,42 +67,58 @@ internal async Task<string> PrintApplicationsList()
...
-                    if (ProvisioningToolOptions.Json)
+                if (graphObjectsList != null && graphObjectsList.Any())
+                {
+                    foreach (var graphObj in graphObjectsList)
                     {
-                        JsonResponse jsonResponse = new JsonResponse(CommandName, State.Success, applicationList);
-                        outputJsonString = jsonResponse.ToJsonString(); 
+                        if (graphObj is Application app)
+                        {
+                            {
```
phil-allen-msft, line 91: (nit) remove extra braces?

src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Tool/MsAADTool.cs
```diff
@@ -113,31 +129,45 @@ internal async Task<string> PrintServicePrincipalList()
...
+                            JsonResponse jsonResponse = new JsonResponse(CommandName, State.Success, servicePrincipalList);
+                            outputJsonString = jsonResponse.ToJsonString();
+                        }
+                        else
+                        {
+                            Console.Write(
+                                "--------------------------------------------------------------\n" +
+                                "Application Name\t\t\t\tApplication ID\n" +
+                                "--------------------------------------------------------------\n\n");
+                            foreach (var app in servicePrincipalList)
```
phil-allen-msft, line 167: instead of app, sp?

src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Tool/MsAADTool.cs
```diff
@@ -67,42 +67,58 @@ internal async Task<string> PrintApplicationsList()
             var graphObjects = await GraphServiceClient.Me.OwnedObjects
                             .Request()
                             .GetAsync();
-            IList<Application> applicationList = new List<Application>();
-            if (graphObjects != null && graphObjects.Any())
+
+            if (graphObjects != null)
```
phil-allen-msft, line 71: Seems like these two algorithms are slightly different on "T" and on the way to print them out after the querying is done.  Factor together the "T" parts, and then have two separate methods that can print them?

src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Tool/MsAADTool.cs
```diff
@@ -67,42 +67,58 @@ internal async Task<string> PrintApplicationsList()
...
+                var nextPage = graphObjects.NextPageRequest;
+                while (nextPage != null)
                 {
-                    if (graphObj is Application app)
-                    {
-                        {
-                            applicationList.Add(app);
-                        }
-                    }
+                    var additionalGraphObjects = await nextPage.GetAsync();
```
phil-allen-msft, line 78: Error handling: Right now, if page 1 succeeds and page 2 fails, this throws and there are no results.
Generally, the best way to handle it is to return the partial results AND the fact that there was an error.  For commandline, you can use stderr for the body of the error (the thrown exception, I think) and stdout for the (partial) results.  For JSON, you might need to introduce a top level property for 'is partial results' and/or 'exception'.
For this PR, consider returning partial results (catch/stderr the exception/return what you have); I don't think Zach is planning on you changing the interface in this PR, but let's adjust for the next one that has an expected interface change.

src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Tool/MsAADTool.cs
deepchoudhery: fixed

src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Tool/MsAADTool.cs
deepchoudhery: fixed

src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Tool/MsAADTool.cs
deepchoudhery: added try catch blocks with console messaging.

src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Tool/MsAADTool.cs
deepchoudhery: done.

src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Tool/MsAADTool.cs
```diff
@@ -61,22 +61,58 @@ public MsAADTool(string commandName, ProvisioningToolOptions provisioningToolOpt
...
+                            var additionalGraphObjects = await nextPage.GetAsync();
+                            if (additionalGraphObjects != null)
+                            {
+                                graphObjectsList.AddRange(additionalGraphObjects.ToList());
+                                nextPage = additionalGraphObjects.NextPageRequest;
+                            }
+                        }
+                        catch (ServiceException)
+                        {
+                            ConsoleLogger.LogMessage("Failed to retrieve all Azure AD/AD B2C objects(apps/service principals", LogMessageType.Error);
```
phil-allen-msft, line 91: nextPage = null;
Otherwise, you have an infinite loop here (I think)

deepchoudhery: good catch fixed

phil-allen-msft (approved):

PR #1574, opened 2021-05-15
updating install-msidentity.cmd to work from any folder

sayedihashimi:
I updated the `install-msidentity.cmd` to work from any folder. How it was written required the user to be in the `scaffolding` folder.


deepchoudhery (approved):
LGTM

PR #1612, opened 2021-08-06
Propagate correct tenant ID from command line args

zahalzel:
When calling the CLI tool from VS, the tenant ID parameter sent in was not being used for creating app registrations, updating app registrations, etc. So in certain cases the CLI tool will infer the wrong tenant ID from the VS project rather than using the tenant that the VS user has currently selected. This change makes it so the CLI tool uses the tenant ID parameter sent in from VS for these operations.


deepchoudhery (approved):
LGTM.

vijayrkn (approved):

vijayrkn:
@zahalzel - Lets bump the version to preview 4 in a follow up PR.  https://github.com/dotnet/Scaffolding/blob/main/eng/Versions.MSIdentity.props#L11

PR #1631, opened 2021-08-31
Log exception for provisioning failures

zahalzel:
Adding logging for exceptions when Provisioning fails, for example due to exceeding the directory quota.

### DIFF
--- a/src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Tool/AppProvisioningTool.cs
+++ b/src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Tool/AppProvisioningTool.cs
@@ -231,12 +231,13 @@ namespace Microsoft.DotNet.MSIdentity
                         ConsoleLogger.LogMessage(failMessage, LogMessageType.Error);
                     }
                 }
-                catch (Exception ex)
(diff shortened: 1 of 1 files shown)


src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Tool/AppProvisioningTool.cs
```diff
@@ -231,12 +231,13 @@ public AppProvisioningTool(string commandName, ProvisioningToolOptions provision
                         ConsoleLogger.LogMessage(failMessage, LogMessageType.Error);
                     }
                 }
-                catch (Exception ex)
+                catch (ServiceException ex)
                 {
+                    string jsonErrorContent = $"{ex.Error?.Code ?? string.Empty} {Environment.NewLine} {ex.Error?.Message ?? string.Empty}";
                     JsonResponse jsonResponse = new JsonResponse(CommandName)
                     {
                         State = State.Fail,
```
phil-allen-msft, line 239: Why not JsonResponse has a State, Content, ErrorCode set of properties?  The Content of an Fail could be the error message quite reasonably, and ErrorCode only set during Fail.

PR #1686, opened 2021-11-03
Update Web API code modifier config

zahalzel:
Adds Scopes attribute when Calls Microsoft Graph or Calls a downstream API option is selected

Adds necessary using statements

Adds code for OnGet method


deepchoudhery (approved):
LGTM

PR #1721, opened 2021-12-07
not adding/forcing Account.Logout.cshtml if blazor project already adding it

deepchoudhery:
Fix for [#38640](https://github.com/dotnet/aspnetcore/issues/38640). 
If Blazor project w/ individual auth adds a custom Accout\Logout.cshtml file, don't try to add/suggest overwrite for that file.

### DIFF
--- a/src/Scaffolding/VS.Web.CG.Mvc/Identity/IdentityGeneratorTemplateModel2.cs
+++ b/src/Scaffolding/VS.Web.CG.Mvc/Identity/IdentityGeneratorTemplateModel2.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Identity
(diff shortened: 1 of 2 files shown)


src/Scaffolding/VS.Web.CG.Mvc/Identity/IdentityGeneratorTemplateModel2.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Identity
```
HaoK, line 4: nit: is it worth putting a comment in here for what the distinction is between IdentityGeneratorTemplateModel2 and IdentityGeneratorTemplateModel does this map to a specific version i.e. pre 6 or something?

deepchoudhery: We actually always use IdentityGeneratorTemplateModel2 as seen [here](https://github.com/dotnet/Scaffolding/blob/82f93025d6f1dd2f20dcac5319e722407e165647/src/Scaffolding/VS.Web.CG.Mvc/Identity/IdentityGeneratorTemplateModelBuilder.cs#L192). 
I didn't write this but I imagine it for extensibility for other developers to add a different template model and use IdentityGeneratorTemplateModel as a base class.

PR #1723, opened 2021-12-10
Adding #nullable disable to scaffolding templates

deepchoudhery:
- Checking for `<Nullable>enable/disable</Nullable>` MsBuild property.
- Adding `#nullable disable` to EF Scaffolding templates if nullable is enabled for the project.
- Helps with current warnings.
- Will be handled with newer templates in .NET 7 timeline likely.


zahalzel (approved):
LGTM

PR #1727, opened 2021-12-11
[WiP] Make sure ActionInvoker throws with InnerException

### LINKED ISSUE #1726: NPE in ActionInvoker.BuildCommandLine
I am trying to generate Razor code for my model.

The scaffolding tool `dotnet-aspnet-codegenerator` should not say
> Object reference not set to an instance of an object.

It should say what is missing and, preferably, how to fix it.
At the very least, the exception thrown should be have the caught exception chained.
Here is a test for the latter feature:
```sh
dotnet-aspnet-codegenerator razorpage Name List |
 wc -l |
 { read L && test "${L}" -gt 8; }
```

yecril71pl:
1. `ActionInvoker` constructs its `InvalidOperationException` chained with the caught exception.
2. Add a test to check that this is the case.

Does not fix #1726.

### DIFF
--- a/src/Scaffolding/VS.Web.CG.Core/ActionInvoker.cs
+++ b/src/Scaffolding/VS.Web.CG.Core/ActionInvoker.cs
@@ -87,7 +87,7 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration
                         ex = ex.GetBaseException();
                     }
 
-                    throw new InvalidOperationException(ex.Message);
+                    throw new InvalidOperationException(ex.Message, ex);
                 }
 
                 return 0;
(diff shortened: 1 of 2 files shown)


deepchoudhery:
Cool yeah that sounds very helpful. I'll quickly try to invoke a couple exceptions to see the formatting of them on the command line. Make sure it doesn't look too messy but should be good.

Thanks a lot,
Deep

yecril71pl:
> Cool yeah that sounds very helpful. I'll quickly try to invoke a couple exceptions to see the formatting of them on the command line. Make sure it doesn't look too messy but should be good.
> 
> Thanks a lot, Deep

I do not think we display the whole chain anyway, this is just the first step.  But as a coding rule of :-1: : if you catch an exception and do nothing with it, you are leaving the users in the dark.  There should be a warning about doing that.

deepchoudhery (approved):
LGTM

PR #1737, opened 2021-12-13
Add tracing to Command.Execute

### LINKED ISSUE #1736: Add tracing to Command.Execute
Failed to get Project Context.

`Command.Execute` should trace the command so that I can repeat it independently.

yecril71pl:
Add Command.ToString (returning file name and arguments)
Add Command._logger
Add LogMessage to Execute (before starting)
Fixes #1736

Tracing is not essential, so no test.


deepchoudhery (changes requested):
I like the idea, especially with just tracing enabled (codegen_trace=1).
After testing it, would like an additional newline added at the end of the statement ala
`$"Executing external command:\n{ this }\n", LogMessageLevel.Trace`.

Also extra whitespace after `LogMessageLevel`. 

Thanks,
Deep

deepchoudhery (approved):
LGTM.

PR #1739, opened 2021-12-14
Do not intercept output from MSBuild

### LINKED ISSUE #1738: MsBuildProjectContextBuilder drops MSBuild output on the floor
## File a bug

`Build` intercepts the output of `MSBuild` and drops it on the floor.

### Include your code

```C#
        public IProjectContext Build()
        {
            var errors = new List<string>();
            var output = new List<string>();
            var tmpFile = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
            var result = 
             Command.CreateDotNet(
                "msbuild",
                new string[]
                {
                    _projectPath,
                    $"/t:EvaluateProjectInfoForCodeGeneration",
                    $"/p:OutputFile={tmpFile};CodeGenerationTargetLocation={_targetLocation};Configuration={_configuration}"
                })
                .OnErrorLine(e => errors.Add(e))
                .OnOutputLine(o => output.Add(o))
                .Execute();

            if (result.ExitCode != 0)
            {
                throw CreateProjectContextCreationFailedException(_projectPath, errors);
            }
            try
            {
                var info = File.ReadAllText(tmpFile);

                var buildContext = JsonConvert.DeserializeObject<CommonProjectContext>(info);

                return buildContext;
            }
            catch (Exception ex)
            {
                throw new InvalidOperationException("Failed to read the BuildContext information.", ex);
            }
        }

```

### Include stack traces

```
 at Microsoft.Extensions.ProjectModel.MsBuildProjectContextBuilder.Build() in /tmp/Scaffolding/src/Shared/Microsoft.DotNet.Scaffolding.Shared/MsBuild/MsBuildProjectContextBuilder.cs:line 58
 at Microsoft.VisualStudio.Web.CodeGeneration.Tools.Program.GetProjectInformation(String projectPath, String configuration) in /tmp/Scaffolding/tools/dotnet-aspnet-codegenerator/Program.cs:line 298
```

### Include provider and version information

<DL ><DT >Microsoft.VisualStudio.Web.CodeGeneration.Design/Microsoft.DotNet.MSIdentity version: <DD >1.0.0
<DT >Target framework: (e.g. .NET 5.0)<DD >net6.0
<DT >Operating system:<DD >Linux x64 5.15.6-1-default
<DT >IDE: (e.g. Visual Studio 2019 16.3)<DD >1.63.0</DL >

yecril71pl:
Do not intercept output from MSBuild
Fixes #1738

Class `MsBuildProjectContextBuilder` is not being tested.  Moreover, we have no control over what `MSBuild` writes to standard output, so there is nothing to assert.

### DIFF
--- a/src/Shared/Microsoft.DotNet.Scaffolding.Shared/MsBuild/MsBuildProjectContextBuilder.cs
+++ b/src/Shared/Microsoft.DotNet.Scaffolding.Shared/MsBuild/MsBuildProjectContextBuilder.cs
@@ -36,7 +36,6 @@ namespace Microsoft.Extensions.ProjectModel
         public IProjectContext Build()
         {
             var errors = new List<string>();
-            var output = new List<string>();
             var tmpFile = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
             var result = Command.CreateDotNet(
                 "msbuild",
@@ -47,7 +46,6 @@ namespace Microsoft.Extensions.ProjectModel
                     $"/p:OutputFile={tmpFile};CodeGenerationTargetLocation={_targetLocation};Configuration={_configuration}"
                 })
                 .OnErrorLine(e => errors.Add(e))
-                .OnOutputLine(o => output.Add(o))
                 .Execute();
 
             if (result.ExitCode != 0)


deepchoudhery:
We can discuss here but I tend to prefer the standard output being redirected. Seeing the build output provides more clarity to VS and command line users as to the scaffolding process. It is not particularly verbose and can highlight warnings that may cause scaffolding to fail. Helps users and me alike to solve any issues with scaffolders and especially custom scaffolders. That's just my take on it.
Let me know how you feel @yecril71pl.

yecril71pl:
> Seeing the build output 

You can see the build output in two ways:

1. Just let the subprocess do its thing.
2. Intercept all output from the subprocess and present it ourselves.

This change switches from mode 2 to mode 1.  Mode 2 makes sense when we do something to the output, like code page conversion, injecting time stamps, teeing into a secondary logger or whatever.  We do not do any of these things, so I would rather prefer it raw, if possible.

PR #1750, opened 2021-12-15
Include projectReferenceString in exception

yecril71pl:
`RoslynWorkspaceHelper.GetProjectReferenceInformation`: when `new Project` throws an [Exception](https://docs.microsoft.com/en-us/dotnet/api/system.exception), wrap it with an [InvalidOperationException](https://docs.microsoft.com/en-us/dotnet/api/system.invalidoperationexception) holding the value of `projectReferenceString`.


deepchoudhery (approved):
LGTM. projectReferenceString is the just the path to the .csproj file and will add that in a comment in a future update.

PR #1751, opened 2021-12-15
OnExecute: log inner exceptions

yecril71pl:
Log the whole [exception chain](https://docs.microsoft.com/en-us/dotnet/api/system.exception.innerexception) when an [Exception](https://docs.microsoft.com/en-us/dotnet/api/system.exception) gets caught by a program.


deepchoudhery (approved):
LGTM. This will be helpful more than it will make formatting annoying in VS error dialogs so I'm cool with it. Might have some VS specific logic if need be later. Ty

PR #1767, opened 2021-12-23
Code updates for blazor server Downstream API option

zahalzel:
1. Adds code updates for blazor server "calls downstream API" option including CallWebApi.razor
2. Adds more code file filtering -> Startup.cs will be ignored when "NonMinimalApp" is false, Program.cs will be ignored when "MinimalApp" is false
3. Fix bug in ProjectModifierHelper. UpdateDocument where code files are added by FileName rather than by FilePath

### DIFF
--- a/src/Shared/Microsoft.DotNet.Scaffolding.Shared/Project/ProjectModifierHelper.cs
+++ b/src/Shared/Microsoft.DotNet.Scaffolding.Shared/Project/ProjectModifierHelper.cs
@@ -38,6 +38,7 @@ namespace Microsoft.DotNet.Scaffolding.Shared.Project
             return startupType == null;
         }
 
+        // Returns true when there is no Startup.cs or equivalent
         internal static async Task<bool> IsMinimalApp(CodeAnalysis.Project project)
         {
             if (project != null)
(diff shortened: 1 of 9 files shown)


src/Shared/Microsoft.DotNet.Scaffolding.Shared/Project/ProjectModifierHelper.cs
```diff
@@ -420,14 +428,14 @@ internal static async Task<Document> ModifyDocumentText(Document fileDoc, IEnume
...
-            return fileDoc.WithText(sourceTextToAdd);
+            var updatedSourceText = SourceText.From(sourceFileString);
+            return fileDoc.WithText(updatedSourceText);
         }
 
         internal static async Task UpdateDocument(Document fileDoc, Document editedDocument, IConsoleLogger consoleLogger)
         {
             var classFileTxt = await editedDocument.GetTextAsync();
-            File.WriteAllText(fileDoc.Name, classFileTxt.ToString());
+            File.WriteAllText(fileDoc.FilePath, classFileTxt.ToString());
```
deepchoudhery, line 438: double check the value for Document.Name vs Document.FilePath. I remember distinctly using Document.Name but blanking rn. 
Maybe add a comment showing the difference between their values.

deepchoudhery (approved):
LGTM. quick concern if you could address it.

PR #1778, opened 2022-01-05
Mark replaced islands in messages

yecril71pl:
Messages where replaced values are unmarked are hard to read when the replaced values are empty or read like a part of running text.


haileymck:
hi, thanks for putting up this PR. I am closing because of the date on this, but feel free to reopen if you would like to continue with these changes.

PR #1798, opened 2022-01-14
Blazor Server: Code updates for Microsoft Graph and Downstream API options

zahalzel:
Updates code modifier configs to add the code to the blazor server templates when authentication is not selected in Visual Studio


deepchoudhery (approved):
LGTM

PR #1814, opened 2022-02-02
not throwing exception on invalid bootstrap versions.

deepchoudhery:
- improved string check for commandline passed bootstrap versions.
- not throwing InvalidOperation exception if bootstrap version not found, just returning default.

Tested .net6 projects with bootstrap 3, 4, 5 and invalid versions. All work well.


zahalzel (approved):
LGTM

PR #1819, opened 2022-02-07
Update Account.Manage.EnableAuthenticator.cs.cshtml

danirzv:
@HaoK


HaoK:
note: I'm not sure why this wasn't an issue in the old bootstrap4 for scaffolding

PR #1823, opened 2022-02-08
throwing new mysql exception mssg with no sensitive information.

deepchoudhery:
In the scenario when using a MySQL db with dotnet webapp/webapi, scaffolding will 
- try to register and activate the DbContext. 
- If the current user/dev does not have the correct MySQL db credentials set/access, MySQL will throw an AccessDenied exception with error code 1045. 
- This exception can contain sensitive information about the server being connected to and passwords in connection strings. 
- Will discard the entire exception message and keep the user id from the MySQL connection string if possible. 

Two possible exception messages:
1. Unable to get DbContext Instance. Access denied for user 'USER' (MySQL db).
2. Unable to get DbContext Instance. Access denied for user (MySQL db).

### DIFF
--- a/src/Scaffolding/VS.Web.CG.EFCore/EntityFrameworkModelProcessor.cs
+++ b/src/Scaffolding/VS.Web.CG.EFCore/EntityFrameworkModelProcessor.cs
@@ -8,6 +8,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.Build.Locator;
 using Microsoft.CodeAnalysis;
@@ -24,6 +25,7 @@ namespace Microsoft.VisualStudio.Web.CodeGeneration.EntityFrameworkCore
     internal class EntityFrameworkModelProcessor
     {
         private const string EFSqlServerPackageName = "Microsoft.EntityFrameworkCore.SqlServer";
+        private const string MySqlException = nameof(MySqlException);
         private const string NewDbContextFolderName = "Data";
         private bool _useSqlite;
         private string _dbContextFullTypeName;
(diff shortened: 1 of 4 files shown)


src/Scaffolding/VS.Web.CG.EFCore/EntityFrameworkModelProcessor.cs
```diff
@@ -554,10 +556,62 @@ private DbContext TryCreateContextUsingAppCode(Type dbContextType, Type startupT
             }
             catch (Exception ex)
             {
+                System.Diagnostics.Debugger.Launch();
+                var exceptionType = ex.GetType();
+                // if MySQL exception with error code 1045, discard error message since it contains sensitive dev information
+                if (exceptionType.Name.Equals(MySqlException, StringComparison.OrdinalIgnoreCase) || exceptionType.FullName.Contains(MySqlException, StringComparison.OrdinalIgnoreCase))
+                {
+                    if (ex.Data.Keys.Count > 0 &&
+                        ex.Data["Server Error Code"] != null &&
```
phil-allen-msft, line 565: ex.Data.TryGetValue("Server Error Code", out object serverErrorCode)......?
Some dictionary implementations will throw if you try to reference a non-existent key, and then you don't need to repeat the constant.

src/Scaffolding/VS.Web.CG.EFCore/EntityFrameworkModelProcessor.cs
```diff
@@ -554,10 +556,62 @@ private DbContext TryCreateContextUsingAppCode(Type dbContextType, Type startupT
             }
             catch (Exception ex)
             {
+                System.Diagnostics.Debugger.Launch();
+                var exceptionType = ex.GetType();
+                // if MySQL exception with error code 1045, discard error message since it contains sensitive dev information
+                if (exceptionType.Name.Equals(MySqlException, StringComparison.OrdinalIgnoreCase) || exceptionType.FullName.Contains(MySqlException, StringComparison.OrdinalIgnoreCase))
```
phil-allen-msft, line 562: Is there a reason to do something other than a full string comparison against  exceptionType.FullName?

src/Scaffolding/VS.Web.CG.EFCore/EntityFrameworkModelProcessor.cs
```diff
@@ -554,10 +556,62 @@ private DbContext TryCreateContextUsingAppCode(Type dbContextType, Type startupT
...
+        /// <returns>username from connection strings found in exception messages.</returns>
+        internal static string GetUsername(string exceptionMssg)
+        {
+            string username = string.Empty;
+            if (!string.IsNullOrEmpty(exceptionMssg))
+            {
+                //if the exception message is in the format `Access denied for user 'admin'@'localhost',
+                //discard the entire message since username and machine name could be both sensitive based on
+                //our collected telemetry.
+                if (exceptionMssg.Contains("Access denied for user", StringComparison.OrdinalIgnoreCase))
```
phil-allen-msft, line 592: Does this work outside of English?

src/Scaffolding/VS.Web.CG.EFCore/EntityFrameworkModelProcessor.cs
```diff
@@ -554,10 +556,62 @@ private DbContext TryCreateContextUsingAppCode(Type dbContextType, Type startupT
...
                 throw ex.Unwrap(_logger);
             }
         }
 
+        /// <summary>
+        /// return the username/user id from exception messages including sensitive information.
+        /// Currently accounting for exception messages using MySQL scenarios.
+        /// </summary>
+        /// <returns>username from connection strings found in exception messages.</returns>
+        internal static string GetUsername(string exceptionMssg)
```
phil-allen-msft, line 584: (nit) Mssg -> Msg?

src/Scaffolding/VS.Web.CG.EFCore/EntityFrameworkModelProcessor.cs
deepchoudhery: TryGetValue is part of Dictionary and casting [Exception.Data](https://docs.microsoft.com/en-us/dotnet/api/system.exception?view=net-6.0) from  IDictionary (not IDictionary<string,string>) is not pretty. Also the untyped IDictionary property in the Exception does guarantee not throwing on non-existent keys as seen [here](https://docs.microsoft.com/en-us/dotnet/api/system.collections.idictionary.item?redirectedfrom=MSDN&view=net-6.0#System_Collections_IDictionary_Item_System_Object_) in all the dotnet versions we will be supporting. We can still do the cast, just not very pretty.

src/Scaffolding/VS.Web.CG.EFCore/EntityFrameworkModelProcessor.cs
deepchoudhery: The exception comes from the MySQL provider and there are plenty of options out there. I tried two different providers and based on that, a FullName check seem to satisfy both of them (they were the two most popular).

src/Scaffolding/VS.Web.CG.EFCore/EntityFrameworkModelProcessor.cs
deepchoudhery: No it does not. The tool itself has no localization support as well but have not tried using MySQL providers in different languages.

phil-allen-msft (approved):

src/Scaffolding/VS.Web.CG.EFCore/EntityFrameworkModelProcessor.cs
deepchoudhery: gotcha, you mean if someone overrides the Exception.Data property and it becomes a different implementation. I'll do the cast.

PR #1851, opened 2022-03-15
Add Calls Graph support to Blazor WASM config

zahalzel:
Adds Calls Microsoft Graph support to Blazor WASM config

test command:
dotnet msidentity --update-project --username $username$--tenant-id $tenant-id$ --client-id $app-id$ --project-file-path $path$ --code-update=true --calls-graph=true


deepchoudhery (approved):
LGTM

PR #1876, opened 2022-04-13
throwing exception if GetConnectionString returns null.

deepchoudhery:
throwing exception if GetConnectionString returns null, adding to identity config for when adding a new DbContext and DbContextConnectionString. 
[Context ](https://github.com/dotnet/Scaffolding/pull/1830)


zahalzel (approved):
Looks good

PR #1901, opened 2022-05-18
Fix spacing and bug fix for Minimal Api Scopes code update

zahalzel:
- Fixes spacing issue for code update configs
- Fixes issue with Minimal Web API code modifier config where "scopeRequiredByApi" variable would not be generated if the Program.cs file does not contain "app.UseAuthorization" line, resulting in a failure to compile.


deepchoudhery (approved):
LGTM

PR #1913, opened 2022-06-05
Fix position of the Logout button near the profile link

jacob-l:
The "btn-link" class has 1px border and, as a result, a button looks misplaced.

Before:
Снимок экрана 2022-06-04 в 21 25 36

After:
Снимок экрана 2022-06-05 в 15 51 54


deepchoudhery:
@HaoK feedback here pls

HaoK:
Any fixes for code in the scaffolding should go through updating the runtime code first, https://github.com/dotnet/aspnetcore/tree/main/src/Identity/UI/src/Areas/Identity

Can you open a PR against that first @jacob-l

jacob-l:
Hi @HaoK , Thank you very much for your assistance. I can not find "_LoginPartial.cshtml" in that folder. However, I have created a new project with .Net 7 Preview 6 and I can see that there is _LoginPartial in the generated project. Where does it take the template?

deepchoudhery:
> Hi @HaoK , Thank you very much for your assistance. I can not find "_LoginPartial.cshtml" in that folder. However, I have created a new project with .Net 7 Preview 6 and I can see that there is _LoginPartial in the generated project. Where does it take the template?

thanks a lot, looks good. 
Appreciate the fix.

Thanks,
Deep

PR #1927, opened 2022-06-10
Fix code update config for blazor wasm

zahalzel:
The "DownstreamApi" option needs to be specified for modifying the "FetchData" page for blazor wasm projects, otherwise the file gets modified even if there is not a downstream API configured.

See bug: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1553434


deepchoudhery (approved):
LGTM.

PR #1931, opened 2022-06-13
Add service principal when exposing API scopes

zahalzel:
Fixes a regression where service principals were not being created when creating app registrations, causing problems with accessing Web APIs. Also small fix to appsettings parsing, to prevent null refs.


deepchoudhery (approved):
LGTM

PR #1939, opened 2022-06-17
Update code modifierconfigs for Blazor Hosted Calls Graph, Calls Downstream API

zahalzel:
Adds the code updates to Program.cs for Blazor Hosted scenario when either or both of the options checkboxes are checked in VS.


deepchoudhery (approved):
LGTM.

PR #1978, opened 2022-08-06
Fix B2C Bugs

zahalzel:
Adds check for "Downstream API" before adding API permissions for B2C apps
Adds "IsB2C" field to app registrations, so that "calls-graph" checkbox is disabled for B2C Apps


deepchoudhery (approved):
LGTM

PR #1983, opened 2022-08-09
init. Removed dependency on MSBuildLocator.RegisterDefaults().

deepchoudhery:
We were using Microsoft.Build.Locator to load in requisite Microsoft.Build assemblies (using MSBuildLocator.RegisterDefaults()) to : 
- initialize Microsoft.Build.Evaluation.Project object
- use `Project.GetProperty("ms build property name")?.EvaluatedValue` to get needed MSBuild project properties
- we needed the "Nullable" property.

Now:
- Added Nullable string property to IProjectContext
- Initialize IProjectContext during runtime msbuild operation using updated `VS.Web.CG.Msbuild/Target/build/Microsoft.VisualStudio.Web.CodeGeneration.Tools.targets` file.
- Nullable property populated at runtime without performing additional Microsoft.Build initialization and calls.

Changes : 
- Removed package references to `Microsoft.Build.Locator`
- Removed all calls to MSBuildLocator.RegisterDefaults().
- Added tests with nullable = enable, disable and the nullable property missing (disable by default).

### DIFF
--- a/src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/CodeReaderWriter/ProjectModifier.cs
+++ b/src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/CodeReaderWriter/ProjectModifier.cs
@@ -6,7 +6,6 @@ using System.Reflection;
 using System.Text;
 using System.Text.Json;
 using System.Threading.Tasks;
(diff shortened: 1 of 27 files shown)


zahalzel (approved):
LGTM

PR #1987, opened 2022-08-16
Use AdhocWorkspace instead of MSBuildWorkspace

zahalzel:
Changes CodeAnalysisHelper to use AdhocWorkspace instead of MSBuildWorkspace

With this change, Document filepaths are always found via Document.Name whereas before, .cs files had the filepath stored in Document.FilePath

### DIFF
--- a/test/Shared/Microsoft.DotNet.Scaffolding.Shared.Tests/CodeAnalysisHelperTests.cs
+++ b/test/Shared/Microsoft.DotNet.Scaffolding.Shared.Tests/CodeAnalysisHelperTests.cs
@@ -1,0 +1,18 @@
+using System.Collections.Generic;
+using Microsoft.DotNet.Scaffolding.Shared.CodeModifier;
+using Xunit;
+
+namespace Microsoft.DotNet.Scaffolding.Shared.Tests
+{
+    public class CodeAnalysisHelperTests
+    {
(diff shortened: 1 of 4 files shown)


test/Shared/Microsoft.DotNet.Scaffolding.Shared.Tests/CodeAnalysisHelperTests.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Microsoft.DotNet.Scaffolding.Shared.Tests
+{
+    public class CodeAnalysisHelperTests
+    {
+        [Fact]
+        public void LoadCodeAnalysisProjectTest()
+        {
+            var files = new List<string>();
+            var projectPath = string.Empty; // TODO test project
```
zahalzel, line 13: @deepchoudhery what do you think would make sense for this?

deepchoudhery: This is good since all we want to test is the initialization of the AdhocWorkspace without any msbuild assembly exceptions. We can add one for loading documents later.

deepchoudhery (approved):
LGTM.

PR #2002, opened 2022-09-01
checking if endpoints class is static, if not, creating one.

deepchoudhery:
Check if endpoints class is static,
- if it **IS**, continue as usual, create endpoints method in that class.
- if it **IS NOT**, create a new static class in the same file (at the end of the file) and add the scaffolded endpoints method to the newly created class.

- additional fix where `Microsoft.EntityFrameworkCore` namespace is added if a DbContext is being used, not just when a DbContext namespace is present (in the endpoints class).

### DIFF
--- a/src/Scaffolding/VS.Web.CG.Mvc/Minimal Api/MinimalApiGenerator.cs
+++ b/src/Scaffolding/VS.Web.CG.Mvc/Minimal Api/MinimalApiGenerator.cs
@@ -154,13 +154,16 @@ namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc.MinimalApi
                     //Get class syntax node to add members to the class
                     var docRoot = docEditor.OriginalRoot as CompilationUnitSyntax;
                     //create CodeFile just to add usings
-
                     var usings = new List<string>();
                     //add usings for DbContext related actins.
                     if (!string.IsNullOrEmpty(templateModel.DbContextNamespace))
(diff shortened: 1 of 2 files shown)


src/Scaffolding/VS.Web.CG.Mvc/Minimal Api/MinimalApiGenerator.cs
```diff
@@ -179,11 +181,31 @@ internal async Task AddEndpointsMethod(string membersBlockText, string endpoints
 
                     if (classNode != null && classNode is ClassDeclarationSyntax classDeclaration)
                     {
+                        SyntaxNode classParentSyntax = null;
+                        //if class is not static, create a new class in the same file
+                        if (!classDeclaration.Modifiers.Any(x => x.Text.Equals(SyntaxFactory.Token(SyntaxKind.StaticKeyword).Text)))
```
zahalzel, line 186: Can x.Text be null?

zahalzel (approved):
Just the one question on string equality, otherwise looks good

src/Scaffolding/VS.Web.CG.Mvc/Minimal Api/MinimalApiGenerator.cs
deepchoudhery: No, it can string.Empty but can't be null.

PR #2013, opened 2022-09-07
not adding extra whitespace and checking for Main method correctly

deepchoudhery:
- not adding extra whitespace when using top level statements.
- checking FileScopedNamespace and regular Namespace for finding methods in `GetMethodFromSyntaxRoot`
  - adding extra logic for `IsUsingTopLevelStatements`


zahalzel (approved):
Looks good 👍🏻

PR #2022, opened 2022-09-21
adding target for and carrying over project refs.

deepchoudhery:
After removing MSBuildLocator to get msbuild assemblies and loading msbuild projects, project references fail for models, db contexts etc.
- added `ResolveProjectReferences` task.


zahalzel (approved):
Looks good to me

PR #2035, opened 2022-10-05
Add Susi Policy Id parameter for Code Update 

zahalzel:
Adds the optional parameter "--susi-policy-id" for the code update command. Used for B2C scenarios to update the appsettings.json file with the correct B2C Sign up sign in policy ID.

Also makes "dotnet-webapp" the default project type in case the project type cannot be inferred.


deepchoudhery (approved):
LGTM.

Issue #1884, opened 2022-05-05
minimalapi -outdir inconsistent behaviour


mrward:
## File a bug

The -outdir argument for the minimalapi is inconsistent with other scaffolders. Other scaffolders support a full path but minimalapi does not.

Specifying a full path ends up creating a file with the filename which is the full path.

Relative paths such as "./" end up creating this as part of the filename.

This works:

dotnet-aspnet-codegenerator minimalapi -m FooModel -e EndPointsClass -outDir "Foo"  --endpointsNamespace "TestNS"

These do not:

dotnet-aspnet-codegenerator minimalapi -m FooModel -e EndPointsClass -outDir "./"  --endpointsNamespace "TestNS"
dotnet-aspnet-codegenerator minimalapi -m FooModel -e EndPointsClass -outDir ""  --endpointsNamespace "TestNS"
dotnet-aspnet-codegenerator minimalapi -m FooModel -e EndPointsClass -outDir "/Full/Path/To/OutDir"  --endpointsNamespace "TestNS"

VS Mac currently uses the full path for -outDir to other scaffolders.

Workaround here is to only use a relative path and remove the -outDir argument if the project directory is being used.

### Include provider and version information

Microsoft.VisualStudio.Web.CodeGeneration.Design version: 6.0.3
Target framework: .NET 6.0.201
Operating system: macOS
IDE: command line

mrward:
The problem seems to be the minimal api scaffolder is using Path.Join not Path.Combine. This does not handle paths in the same way.

https://github.com/dotnet/Scaffolding/blob/main/src/Scaffolding/VS.Web.CG.Mvc/Minimal%20Api/MinimalApiGenerator.cs#L132

```
string path1 = "/Users/test/foo";
string path2 = "/Users/test/foo/bar";

string result = Path.Combine(path1, path2);
Console.WriteLine (result);

string result2 = Path.Join(path1, path2);
Console.WriteLine (result2);
```

Output:

```
/Users/test/foo/bar
/Users/test/foo/Users/test/foo/bar
```

deepchoudhery:
@mrward  Hey, should be fixed in .NET 7 scenarios. Apologies didn't get serviced for .NET 6, will get fixed next month.

PR #2036, opened 2022-10-05
fix minimal api --outDir bug

deepchoudhery:
Addresses #1884 
- Changed `Path.Join` to `Path.Combine`
- Not skipping test for this use.


zahalzel (approved):
Looks good

Issue #2037, opened 2022-10-06
Changes for .NET 7 minimal API scaffolders


DamianEdwards:
- Add whitespace route group definition and first API definition
- Remove all `Produces` extension method calls (they aren't needed when using `TypedResults` or returning the object to be serialized directly)
- Remove all calls to `WithTags` from API definitions and instead have single call on the group definition
- Move call to `WithOpenApi` from group definition to API definitions (that method is designed to allow editing a specific `OpenApiOperation` so is more useful on the endpoint definitions directly IMO)
- Update non-EF Core scaffolder to also use `TypedResults`
- Question: Do we want to add an option for using strongly typed results?
  - My thinking is "yes" because they represent an alternate of writing the API but not necessarily a "better" way and some folks may prefer using `Results` and the `Produces` approach

DamianEdwards:
FYI @deepchoudhery @vijayrkn @JeremyLikness

deepchoudhery:
@DamianEdwards Working on getting these in for 7.0 GA. I do not have enough feedback on your question regarding strongly typed results, so I think you all should make that call.

DamianEdwards:
@deepchoudhery if we're able to get that option added (simple boolean checkbox "Use TypedResults") then I think we should do it. If it's going to increase the work such that it will push this out past GA then we should leave it out for now and just do the TypedResults version.

@JeremyLikness thoughts?

JeremyLikness:
+1 on the checkbox. Let's give the flexibility to choose unless it compromises GA. Also concur fallback s/b TypedResults.

deepchoudhery:
some sample dialogs for the proposed change on the VS side. On the cmd line, it'll be another option that is passed.

Non ef dialog, showing checkbox for TypedResults:
<img width="488" alt="NeMK9OfMsT" src="https://user-images.githubusercontent.com/54324771/196535617-1d072494-625b-4d38-b20d-40958b9e280f.png">

Ef dialog, showing as launched:
<img width="488" alt="yZGtSBWpJX" src="https://user-images.githubusercontent.com/54324771/196535703-4f6a9e8f-0b68-476d-b8db-9d1c44e3800d.png">

Ef dialog, showing with existing db, sqlite disabled as its only allowed for new dbs:
<img width="488" alt="e2UPOoN34M" src="https://user-images.githubusercontent.com/54324771/196535781-3a33a502-95de-48ce-865e-e87b03e19156.png">

[...]

PR #2046, opened 2022-10-12
minimal api net7 changes

deepchoudhery:
Addressing changes from #2037.

Notes (@DamianEdwards) : 
> `Update non-EF Core scaffolder to also use TypedResults`
- the non-EF scaffolder has commented out the Results return so just have changed it TypedResults but still commented out.

### DIFF
--- a/src/Scaffolding/VS.Web.CG.Mvc/Templates/MinimalApi/MinimalApi.cshtml
+++ b/src/Scaffolding/VS.Web.CG.Mvc/Templates/MinimalApi/MinimalApi.cshtml
@@ -44,8 +44,8 @@ public static class @endPointsClassName
     @{
     if(Model.OpenAPI)
     {
-        @:.WithOpenApi()
-        @:.WithName("@getAllModels");
+        @:.WithName("@getAllModels")
+        @:.WithOpenApi();
     }
     else
     {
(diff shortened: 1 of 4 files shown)


src/Scaffolding/VS.Web.CG.Mvc/Templates/MinimalApi/MinimalApi.cshtml
```diff
@@ -103,14 +112,22 @@ public static class @endPointsClassName
...
+        }
         })
     @{
     if(Model.OpenAPI)
     {
-        @:.WithTags(nameof(@modelName))
-        @:.WithName("@deleteModel")
-        @:.Produces<@modelName>(StatusCodes.Status200OK);
+        @:.WithOpenApi()
+        @:.WithName("@deleteModel");
```
DamianEdwards, line 130: These should all be reversed so that `WithName` is before `WithOpenApi`

DamianEdwards:
Are you we planning on adding the option for `TypedResults` as discussed in the issue?

src/Scaffolding/VS.Web.CG.Mvc/Templates/MinimalApi/MinimalApi.cshtml
deepchoudhery: yep ty fixed.

deepchoudhery:
> Are you we planning on adding the option for `TypedResults` as discussed in the issue?

Yep, that'll be in the next PR and that will have VS work alongside it. Won't be closing the issue for now.

DamianEdwards (approved):

PR #2129, opened 2022-11-30
Output error for app registrations, and small formatting nit

zahalzel:
- Adding an error output when app registrations fail to load
- Fixing a redundant newline formatting nit in DocumentBuilder.cs


deepchoudhery (approved):
LGTM! ty

PR #2144, opened 2022-12-07
Fix output for app registrations

zahalzel:
A few small fixes:
- if GetGraphObjects method fails, puts the output in the JsonResponse Output field rather than the Content field
- if there are no app registrations, but everything succeeded, returns a JsonResponse with empty list rather than failing


deepchoudhery (approved):
LGTM! Let's add a test for the empty app registration scenario (or an issue open is good)

PR #2205, opened 2023-01-06
bump system commandline version

deepchoudhery:
bumping System.CommandLine version to 2.0.0-beta4.22613.1.
- added `dotnet-libraries` nuget feed for the latest System.CommandLine package
- added `ProvisioningToolOptionsBinder` for using `SetHandler` instead of `CommandHandler.Create`
- changed all the Options for dotnet-msidentity from new objects to readonly properties.


zahalzel (approved):
Looks good!

PR #2210, opened 2023-01-09
Fix authentication bug (regenerate msal.cache)

zahalzel:
The tool has been relying on Visual Studio to populate or create the msal.cache file for storing account information, but since VS has transitioned to using WAM it no longer populates the msal.cache file. To work around this, we can make a call to AcquireTokenInteractive when an account is not found, and this will pop up a browser window allowing for picking an account.


deepchoudhery (approved):
LGTM!

Issue #2143, opened 2022-12-07
Unnecessary null checks in scaffolded MVC Controllers using EF Core Read/Write Actions


DanielStout5:
## Ask a question

This [PR](https://github.com/dotnet/Scaffolding/pull/1887/files) added null checks for the DbSet.

Some are conditional upon nullable reference types being enabled, but others, like the one in the Details endpoint, are not:

![image](https://user-images.githubusercontent.com/13984239/206216308-68bf5380-8c96-427b-967b-ce216cba1970.png)

The [EF Core documentation on nullable reference types](https://learn.microsoft.com/en-us/ef/core/miscellaneous/nullable-reference-types) says **"The DbContext base constructor ensures that all DbSet properties will get initialized, and null will never be observed on them"** so why is the scaffolding adding this unnecessary clutter?

I'm sure the performance impact is minimal since it's just a null check, but the bigger issue is the confusion this generates since constantly checking for the DbSet properties being null implies that's a reasonable state for the DbContext to be in, despite the official documentation saying the DbSet properties **should never be null**.

Unless there's some reason I'm missing, I think that PR should be reverted and scaffolded controllers should assume that the DbSet properties are non-null.

ajcvickers:
As @DanielStout5 says, null checks on DbSet properties are not appropriate. Use either:

```C#
public DbSet<Foo> Foos => Set<Foo>();
```

or

```C#
public DbSet<Foo> Foos { get; set; } = null!;
```

or reply on the [suppressor](https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-7.0/whatsnew#suppressor-for-uninitialized-dbset-properties).

deepchoudhery:
addressed #2211. Will close after it's live in 7.0.2

deepchoudhery:
Closing, fixed in 7.0.2. Will be serviced for 6.0.12 in Feb 2023.

PR #2211, opened 2023-01-09
removing DbSet null checks from #1887

deepchoudhery:
- addressing #2143 
- removed DbSet null checks
- will be serviced for 7.0.2, later for .NET 6.0 servicing.


zahalzel (approved):
Looks good 👍🏻

PR #2314, opened 2023-03-06
Add web app MVC code modifier

zahalzel:
- Adds code modifier config for DotNet Web App MVC template
- Adds functionality for modifiying method attributes


deepchoudhery (approved):
LGTM

PR #2359, opened 2023-03-27
Fix code modifier config

zahalzel:
Fixing a spacing issue with the MVC code modifier config


deepchoudhery (approved):
LGTM, we might have some spacing issues where Program.cs has a main method with a namespace (no top level statements scenario)

PR #2400, opened 2023-05-08
Add input parameter for Instance

zahalzel:
Adds an optional "Instance" input parameter, so that we can use the tool on other clouds besides the Azure Public cloud, such as the government cloud, the China cloud, etc.


deepchoudhery (approved):
LGTM

PR #2401, opened 2023-05-08
pluralizing model name in minimalapi scenarios using humanizer

deepchoudhery:
using Humanizer v2.14.1
- will be porting to .NET 6, 7 and will be in the next .NET 8 Preview


zahalzel (approved):
Looks good :)

PR #2410, opened 2023-05-12
Add instance parameter to Release/7.0 branch

zahalzel:
Required some different code from main, as a result of a different System.CommandLine version, so making a separate PR instead of cherry-picking from main.


deepchoudhery (approved):
LGTM.

PR #2496, opened 2023-08-15
supporting netstandard2.0 for Microsoft.DotNet.Scaffolding.Shared

deepchoudhery:
adding support for netstandard2.0 for MIcrosoft.DotNet.Scaffolding.Shared so it can be consumed by VS.
- added `ContainsIgnoreCase` since netstandard2.0 does not have a string comparison with StringOrdinal parameters.

### DIFF
--- a/src/Shared/Microsoft.DotNet.Scaffolding.Shared/General/StringUtil.cs
+++ b/src/Shared/Microsoft.DotNet.Scaffolding.Shared/General/StringUtil.cs
@@ -3,7 +3,7 @@
 
 using System;
 
-namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc
+namespace Microsoft.DotNet.Scaffolding.Shared
 {
     internal static class StringUtil
     {
@@ -19,6 +19,11 @@ namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc
                 return input.Substring(0, length: 1).ToLowerInvariant() + input.Substring(1);
             }
             return input;
+        }
+
+        public static bool ContainsIgnoreCase(this string input, string value)
+        {
+            return input.ToLowerInvariant().Equals(value.ToLowerInvariant());
         }
     }
 }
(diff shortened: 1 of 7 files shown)


src/Shared/Microsoft.DotNet.Scaffolding.Shared/General/StringUtil.cs
```diff
@@ -20,5 +20,10 @@ public static string ToLowerInvariantFirstChar(this string input)
             }
             return input;
         }
+
+        public static bool ContainsIgnoreCase(this string input, string value)
+        {
+            return input.ToLowerInvariant().Equals(value.ToLowerInvariant());
```
anvillan, line 26: Do we want to verify `input` and/or `value` is not null before doing `ToLowerInvariant()`? Not sure if that will throw when the string value is `null`.

test/Shared/Microsoft.DotNet.Scaffolding.Shared.Tests/StringUtilTests.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Microsoft.DotNet.Scaffolding.Shared.Tests
+{
+    public class StringUtilTests
+    {
+        [Fact]
+        public void ContainsIgnoreCase_ShouldReturnCorrectResults()
+        {
+            // Arrange
+            string input1 = "Hello";
```
anvillan, line 14: Add a case for when either value, or both, is `null`

src/Shared/Microsoft.DotNet.Scaffolding.Shared/General/StringUtil.cs
deepchoudhery: yes ty will do

test/Shared/Microsoft.DotNet.Scaffolding.Shared.Tests/StringUtilTests.cs
deepchoudhery: will add check and test case tyty

deepchoudhery:
Will create a different one, this one is having pipeline funkyness.

PR #2522, opened 2023-09-11
Add try/catch for adding owner to app registration

zahalzel:
For some tenants, the owner is automaticaly added to the app registration, so we want to continue and not throw an error

### DIFF
--- a/src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/MicrosoftIdentityPlatform/MicrosoftIdentityPlatformApplicationManager.cs
+++ b/src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/MicrosoftIdentityPlatform/MicrosoftIdentityPlatformApplicationManager.cs
@@ -81,18 +81,11 @@ namespace Microsoft.DotNet.MSIdentity.MicrosoftIdentityPlatform
                     consoleLogger.LogFailureAndExit(Resources.FailedToCreateApp);
                 }
 
-                // Add the current user as a owner.
-                User? me = await graphServiceClient.Me.GetAsync();
-                var requestBody = new ReferenceCreate
-                {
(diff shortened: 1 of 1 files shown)


src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/MicrosoftIdentityPlatform/MicrosoftIdentityPlatformApplicationManager.cs
```diff
@@ -155,6 +148,31 @@ public class MicrosoftIdentityPlatformApplicationManager
...
+                var requestBody = new ReferenceCreate
+                {
+                    OdataId = $"https://graph.microsoft.com/beta/directoryObjects/{me?.Id}",
+                };
+
+                await graphServiceClient.Applications[createdApplication!.Id].Owners.Ref.PostAsync(requestBody);
+            }
+            catch (Graph.Models.ODataErrors.ODataError)
+            {
+                // Owner already exists in app registration
```
deepchoudhery, line 172: should we do something in this catch, maybe just console output?

deepchoudhery (approved):
LGTM!, minor comment.

src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/MicrosoftIdentityPlatform/MicrosoftIdentityPlatformApplicationManager.cs
zahalzel: In this case, I don't think it's worth logging, as the info doesn't really affect the customer and the scenario where the exception is thrown seems to be limited to only the microsoft tenant at this point

PR #2530, opened 2023-09-14
supporting netstandard2.0 for Microsoft.DotNet.Scaffolding.Shared

deepchoudhery:
adding support for netstandard2.0 for MIcrosoft.DotNet.Scaffolding.Shared so it can be consumed by VS.
- removing net6.0, net7.0, net8.0 build
- added ContainsIgnoreCase since netstandard2.0 does not have a string comparison with StringOrdinal parameters.


anvillan (approved):

zahalzel (approved):
LGTM 👍🏻

PR #2568, opened 2023-10-24
Downstream api fix on release7

zahalzel:
Fixes downstream API scenario, and updates strings for removing AAD references


deepchoudhery (approved):
LGTM!

PR #2577, opened 2023-11-09
Remove comment about DotNet-Blob-Feed

akoeplinger:
Makes it easier to search for repos still using the variable.

/cc @deepchoudhery


akoeplinger:
@deepchoudhery ping

deepchoudhery:
> Makes it easier to search for repos still using the variable.
> 
> /cc @deepchoudhery

looks good, merging. tyty

PR #2627, opened 2024-01-29
Add blazor identity

deepchoudhery:
Adding 'dotnet-aspnet-codegenerator blazor-identity' generator.
Official [documentation ](https://learn.microsoft.com/en-us/aspnet/core/blazor/security/?view=aspnetcore-8.0)for ASP.NET Core Blazor Authentication and Authorization.
Help:
[
Screenshot 2024-02-05 155354
](url)
Scaffold blazor identity components and files.
- create/use an existing IdentityDbContext
- create/use an existing IdentityUser
- include/exclude files, or scaffold all of them
- necessary Program.cs changes are applied as well.
- necessary App.razor, Routes.razor and NavMenu.razor changes are applied as well.

[Sample ](https://github.com/deepchoudhery/Sample-BlazorWebApp-Identity)blazor web app with scaffolded identity


src/Scaffolding/VS.Web.CG.Mvc/BlazorIdentity/BlazorIdentityGenerator.cs
```diff
@@ -0,0 +1,544 @@
...
+                blazorIdentityModel.FilesToGenerate = AllBlazorIdentityFiles.Keys.ToList();
+            }
+
+            blazorIdentityModel.BaseOutputPath = Path.Combine(AppInfo.ApplicationBasePath, commandlineModel.RelativeFolderPath);
+            return blazorIdentityModel;
+        }
+
+        private void ShowFileList()
+        {
+            Logger.LogMessage("File List:\n");
```
tlmii, line 302: This might be just my lack of knowledge on this codebase, but why are some things written to Logger and others (like down in ModifyFilesAsync) written to ConsoleLogger?

src/Scaffolding/VS.Web.CG.Mvc/BlazorIdentity/BlazorIdentityGenerator.cs
```diff
@@ -0,0 +1,544 @@
...
+            //Modifying Program.cs document
+            var docEditor = await DocumentEditor.CreateAsync(programDocument);
+            var docRoot = docEditor.OriginalRoot as CompilationUnitSyntax;
+            programCsFile.Usings = programCsFile.Usings.Union(new string[] { blazorIdentityModel.DbContextNamespace, blazorIdentityModel.BlazorIdentityNamespace }).ToArray();
+            var docBuilder = new DocumentBuilder(docEditor, programCsFile, ConsoleLogger);
+            var newRoot = docBuilder.AddUsings(new CodeChangeOptions());
+            var useTopLevelsStatements = await ProjectModifierHelper.IsUsingTopLevelStatements(project.Documents.ToList());
+            var builderIdentifier = ProjectModifierHelper.GetBuilderVariableIdentifierTransformation(newRoot.Members);
+            //add code snippets/changes.
+            if (programCsFile.Methods != null && programCsFile.Methods.Count != 0)
```
tlmii, line 389: `if (programCsFile.Methods?.Count > 0)` ?

src/Scaffolding/VS.Web.CG.Mvc/BlazorIdentity/BlazorIdentityGenerator.cs
```diff
@@ -0,0 +1,544 @@
...
+        private const string Main = nameof(Main);
+        private bool CalledFromCommandline => !(FileSystem is SimulationModeFileSystem);
+
+        private IEnumerable<string> _templateFolders;
+        //Folders where the .tt templates for Blazor Identity scenario live. Should be in VS.Web.CG.Mvc\Templates\BlazorIdentity
+        private IEnumerable<string> TemplateFolders
+        {
+            get
+            {
+                if (_templateFolders is null || !_templateFolders.Any())
```
tlmii, line 53: For this condition, is the intention that if the `TemplateFoldersUtilities.GetTemplateFolders` call below returned an empty result set, that we'd keep checking it? Should that scenario even really happen? Or should this effectively just be a lazy load (if null check only, no empty check)?

I guess the same question applies to all the below collections as well (`AllBlazorIdentityFiles`, `GeneralT4Files`, `BlazorIdentityTemplateTypes`).

src/Scaffolding/VS.Web.CG.Mvc/BlazorIdentity/BlazorIdentityGenerator.cs
```diff
@@ -0,0 +1,544 @@
...
+                    UserClassModel = applicationUserModel
+                };
+
+                ExecuteDbContextTemplate(dbContextModel);
+                var connectionStringsWriter = new ConnectionStringsWriter(AppInfo, FileSystem);
+                connectionStringsWriter.AddConnectionString("DefaultConnection", dbContextModel.DbContextName, commandlineModel.DatabaseProvider);
+            }
+
+            //gather blazor identity files to generate
+            blazorIdentityModel.FilesToGenerate = new List<string>();
```
tlmii, line 265: Could you put this in an else, so you don't new it up if you're going to just replace it on line 268 or 288?

src/Scaffolding/VS.Web.CG.Mvc/BlazorIdentity/BlazorIdentityGenerator.cs
```diff
@@ -0,0 +1,544 @@
...
+                    throw new InvalidOperationException(string.Join(Environment.NewLine, errors));
+                }
+
+                //get files to overwrite
+                blazorIdentityModel.FilesToGenerate = AllBlazorIdentityFiles.Keys.Except(excludedFiles).ToList();
+            }
+
+            if (blazorIdentityModel.FilesToGenerate.Count == 0)
+            {
+                blazorIdentityModel.FilesToGenerate = AllBlazorIdentityFiles.Keys.ToList();
```
tlmii, line 293: This might impact the answer to my last comment about line 265, but is the intention here to do this even if one of the two conditions above executed and the result of them was an empty set? Or only if neither condition was true?

src/Scaffolding/VS.Web.CG.Mvc/BlazorIdentity/BlazorIdentityGenerator.cs
```diff
@@ -0,0 +1,544 @@
...
+            var programType = ModelTypesLocator.GetType("<Program>$").FirstOrDefault() ?? ModelTypesLocator.GetType("Program").FirstOrDefault();
+            var project = Workspace.CurrentSolution.Projects.FirstOrDefault(p => p.AssemblyName.Equals(ProjectContext.AssemblyName, StringComparison.OrdinalIgnoreCase));
+            var programDocument = project.GetUpdatedDocument(FileSystem, programType);
+            //Modifying Program.cs document
+            var docEditor = await DocumentEditor.CreateAsync(programDocument);
+            var docRoot = docEditor.OriginalRoot as CompilationUnitSyntax;
+            programCsFile.Usings = programCsFile.Usings.Union(new string[] { blazorIdentityModel.DbContextNamespace, blazorIdentityModel.BlazorIdentityNamespace }).ToArray();
+            var docBuilder = new DocumentBuilder(docEditor, programCsFile, ConsoleLogger);
+            var newRoot = docBuilder.AddUsings(new CodeChangeOptions());
+            var useTopLevelsStatements = await ProjectModifierHelper.IsUsingTopLevelStatements(project.Documents.ToList());
```
tlmii, line 386: typo? should be `useTopLevelStatements`?

src/Scaffolding/VS.Web.CG.Mvc/Identity/IdentityHelper.cs
```diff
@@ -0,0 +1,154 @@
...
+            {
+                return null;
+            }
+
+            return dbContext.Substring(0, dbContext.LastIndexOf('.'));
+        }
+
+        internal static string GetClassNameFromTypeName(string dbContext)
+        {
+            return dbContext.Split(new char[] { '.' }, StringSplitOptions.RemoveEmptyEntries).Last();
```
tlmii, line 46: I'd generally suggest not doing a Split if you don't care about all of the parts. If you're just going after the last one, you could do effectively the opposite of what you did above in `GetNamespaceFromTypeName` above. And then the two methods next to each other would be more consistent as well.

src/Scaffolding/VS.Web.CG.Mvc/BlazorIdentity/BlazorIdentityGenerator.cs
deepchoudhery: yep that's better tyty.

src/Scaffolding/VS.Web.CG.Mvc/BlazorIdentity/BlazorIdentityGenerator.cs
deepchoudhery: it should just be a lazy load, null check only. We don't need to keep to checking, tyty.

src/Scaffolding/VS.Web.CG.Mvc/BlazorIdentity/BlazorIdentityGenerator.cs
deepchoudhery: It's needed for the DocumentBuilder class. Ideally there should only be one logger but this created earlier for dotnet-msidentity (DocumentBuilder was written alongside). Changed all references of ConsoleLogger to Logger other than the one needed for DocumentBuilder. These should be combined in our rework soon.

src/Scaffolding/VS.Web.CG.Mvc/BlazorIdentity/BlazorIdentityGenerator.cs
deepchoudhery: If neither condition is true, then we just go ahead scaffold all identity files is the intent. If we end up with an empty set from the 2 conditions, it's ok to scaffold I presume. I think that new `FilesToGenerate` list is probably unnecessary so will just add a final else statement. tyty

src/Scaffolding/VS.Web.CG.Mvc/BlazorIdentity/BlazorIdentityGenerator.cs
deepchoudhery: addressed, check comment below

src/Scaffolding/VS.Web.CG.Mvc/Identity/IdentityHelper.cs
deepchoudhery: sounds good, this file also has no unit tests so adding those.

src/Scaffolding/VS.Web.CG.Mvc/BlazorIdentity/BlazorIdentityGenerator.cs
```diff
@@ -80,7 +80,7 @@ private IEnumerable<string> GeneralT4Files
         {
             get
             {
-                if (_generalT4Files == null || !_generalT4Files.Any())
+                if (_generalT4Files == null)
```
tlmii, line 83: Should have commented on this earlier, but its more obvious now. The back and forth of `== null` sometimes and `is null` other times makes my OCD hurt :) Can we standardize on `is null`?

src/Scaffolding/VS.Web.CG.Mvc/Identity/IdentityHelper.cs
```diff
@@ -43,7 +43,13 @@ internal static string GetNamespaceFromTypeName(string dbContext)
...
         internal static string GetClassNameFromTypeName(string dbContext)
         {
-            return dbContext.Split(new char[] { '.' }, StringSplitOptions.RemoveEmptyEntries).Last();
+            var lastIndexOfDot = dbContext.LastIndexOf('.');
+            if (lastIndexOfDot == -1)
+            {
+                return string.Empty;
+            }
+
+            return dbContext.Substring(lastIndexOfDot + 1, dbContext.Length - lastIndexOfDot - 1);
```
tlmii, line 52: You should just be able to use the [Substring overload that just takes `startIndex`](https://learn.microsoft.com/en-us/dotnet/api/system.string.substring?view=net-8.0#system-string-substring(system-int32)). That'll return everything after that.

tlmii (approved):

src/Scaffolding/VS.Web.CG.Mvc/Identity/IdentityHelper.cs
deepchoudhery: fixed tyty

PR #2665, opened 2024-03-07
Fork build pipeline for pr/ci

tlmii:
This is just the first step of getting the build pipelines ready for 1ES pipeline template migration. Here we create a copy of the build pipeline that will be used for public/pr builds. Modifications to the file after copy are limited to removing conditionals that were relevant only to internal/ci builds. For review purposes its better to compare this new file with azure-pipelines.yml since its treated as a new file here.

Note that the PR build for this PR will still run against the old yml file. I'll swap over the public build to the new file once this PR completes.

### DIFF
--- a/azure-pipelines-pr.yml
+++ b/azure-pipelines-pr.yml
@@ -1,0 +1,114 @@
+pr:
+  branches:
+    include:
+    - '*'
+
+variables:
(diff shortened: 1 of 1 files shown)


deepchoudhery (approved):
LGTM!

PR #2667, opened 2024-03-07
Restore CI trigger for PR pipeline

tlmii:
Quick fix from #2665 - I removed the trigger altogether, which actually made CI _always_ run. On second thought I'll just put it back as it was.


deepchoudhery (approved):
sgsg

PR #2668, opened 2024-03-08
Migrate internal pipeline to MicroBuild template

tlmii:
This is (hopefully) the final step of migrating the pipelines to the new 1ES/MicroBuild templates. The deletion of SetupNugetSources.ps1 is because that file fails the SDLSources step. That issue is fixed with https://github.com/dotnet/arcade/pull/14553 but that hasn't flowed here yet. When it does, we'll get a new copy.


tlmii:
@deepchoudhery If you could take a look, but don't merge until we've finished validating the internal builds.

deepchoudhery (approved):
LGTM! will wait for you to merge

PR #2730, opened 2024-05-06
Remove -pt images which are no longer available

tlmii:
These images are no longer available in the build pools


deepankarchoudhery (approved):
LGTM!

PR #2737, opened 2024-05-08
adding 'dotnet-scaffold-aspire' and 'dotnet-scaffold-aspnet'

deepchoudhery:
adding mostly hollow dotnet-scaffold-aspire and dotnet-scaffold-aspnet
- there is duplicate code for Specter.Console helpers. Currently do not want to add a dependency of Spectre.Console onto the Shared helper projects (Will abstract this work to a Spectre.Console.Helper project).
- Mostly hollow cmdline apps that execute 'get-commands' and will be filled out soon.

### DIFF
--- a/eng/Versions.Scaffold.AspNet.props
+++ b/eng/Versions.Scaffold.AspNet.props
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- Opt out of certain Arcade features -->
+  <PropertyGroup>
+    <UsingToolXliff>false</UsingToolXliff>
+    <UsingToolNetFrameworkReferenceAssemblies>true</UsingToolNetFrameworkReferenceAssemblies>
+  </PropertyGroup>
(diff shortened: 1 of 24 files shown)


eng/Versions.Scaffold.AspNet.props
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
```
tlmii, line 1: Do we need to have these (this file and `eng/Versions.Scaffold.Aspire.props` and the existing `eng/Versions.Scaffold.props`) separate immediately since they're identical (or nearly identical in the case of the existing file)? Or can we split them when necessary?

tlmii (approved):
Good shell to get us going. I'm mostly ignoring the aspnet one and focused on the aspire one as I assume you know the plan for the aspnet one better than me :)

eng/Versions.Scaffold.AspNet.props
deepchoudhery: These are going to be different pretty quickly I think as their dependencies and properties will diverge. Also since we might develop on these tools separately, its nice to have this separation for clarity and if we need to release them not in sync.

PR #2736, opened 2024-05-08
Consolidate LangVersion on 'latest'

tlmii:
We have several different LangVersions specified in different places and overridden in others (including defaulting to an old old version 7.3 in D.B.props). Just a quick attempt to consolidate on `latest`, which appears to work for all projects currently.

Issue #2732, opened 2024-05-07
Make all projects nullable friendly


deepchoudhery:
Title, convert all csproj and project contents to be nullable enabled.

PR #2739, opened 2024-05-09
Invert nullable default

tlmii:
Related to #2732 

Previously the default for `<Nullable>` in the repo was not specified (thus making it `disable`). This meant that each time we added a new project we needed to remember to add `<Nullable>enable</Nullable>` to it, even though that really should be the default.

This PR inverts the default, setting `<Nullable>enable</Nullable>` at the root in D.B.props, removing the same from any existing projects that have it and adding `<Nullable>disable</Nullable>` to any projects that require code changes.

I intentionally did not make any code changes here to keep this PR clean (since it already touched a lot of files) even though some projects likely required minimal changes to support nullable types.

Note that I intentionally did not try to use deeper D.B.props files even when all or most of the projects within a subfolder needed nullable types disabled. This will make it easier for us to know how many projects are left and keep the changes isolated to one project at a time as we go through and do it. 

I also dropped a few `<LangVersion>latest</LangVersion>` that snuck in via #2737 since that was in flight already when I did #2736

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -16,6 +16,7 @@
   <PropertyGroup>
     <RepositoryRoot>$(MSBuildThisFileDirectory)</RepositoryRoot>
     <LangVersion>latest</LangVersion>
+    <Nullable>enable</Nullable>
   </PropertyGroup>
 
   <PropertyGroup Label="Package and Assembly Metadata">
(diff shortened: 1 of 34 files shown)


deepchoudhery (approved):

deepchoudhery (approved):
sounds good, LGTM!

PR #2741, opened 2024-05-10
using ParameterDiscovery and InteractivePickerType.ProjectPicker 

deepchoudhery:
Changing project discovery for 1st and 3rd party components (dotnet tools)
- remove `SourceProjectFlowStep.cs` (from the list of flow steps from ScaffoldingCommand as well) and `ProjectDiscovery.cs`
- move `ProjectDiscovery` logic to `ParameterDiscovery` under the `InteractivePickerType.ProjectPicker` switch case.


tlmii (approved):
I think I see what you mean about ParameterDiscovery. We can iterate.

PR #2745, opened 2024-05-15
Change AppHost verbiage

tlmii:
My comment on a prior PR (https://github.com/dotnet/Scaffolding/pull/2744#discussion_r1600715024) was too brief and didn't convey what I intended. So just a quick PR to make the full change here. We can get the PM/docs folks later on whether it should be `App host` (like in the docs) or `AppHost` (like in the project name) or what, but it should probably have both parts (see [Aspire docs](https://learn.microsoft.com/en-us/dotnet/aspire/fundamentals/app-host-overview#app-host-project) for reference).


deepchoudhery (approved):
yes makes sense, LGTM!

PR #2748, opened 2024-05-15
Fix database apiservice package add

tlmii:
Looks like a simple copy/paste error here - seems to be using the AppHost dictionary for the ApiService reference install.


deepchoudhery (approved):
oops tyty

PR #2750, opened 2024-05-16
Various quick tweaks

tlmii:
- More cleanup on App host project command argument and its naming relative to the other project argument.
- Consolidating the web/api project command arguments into one Project command argument. It really doesn't matter in Aspire what type of thing it is - it could be web, web api, worker, even a WPF app or something. I started to just call it `.NET Project` in the display name but that felt _too_ generic. Went with Brady's suggestion of Web or worker for now.
- Missed one place when changing the version from 0.1.0 back to 9.0.0
- Turned off showing the selected options on launch at least for now since we're rarely doing it and it just clutters the demo
- Changed the way we displayed the output from the commands so that we always display both stdout/stderr regardless of exit code (there can be both regardless of exit code) but only display the exit code if it is non-zero. 
- Hid the command being executed as it clutters the demo at the moment
- Left a few TODOs in the command execute flow step

### DIFF
--- a/src/Shared/Microsoft.DotNet.Scaffolding.Helpers/General/DotnetCommands.cs
+++ b/src/Shared/Microsoft.DotNet.Scaffolding.Helpers/General/DotnetCommands.cs
@@ -25,7 +25,7 @@ public static class DotnetCommands
             }
 
             arguments.Add("--prerelease");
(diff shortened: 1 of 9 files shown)


deepchoudhery (approved):
LGTM! tyty

PR #2754, opened 2024-05-17
Tweak AddPackage output

tlmii:
I don't think we can assume that the output in the failure case is on stderr - I think it can (will?) be on stdout. So we should write out both to get the full picture of what's going on in the failure case.

### DIFF
--- a/src/Shared/Microsoft.DotNet.Scaffolding.Helpers/General/DotnetCommands.cs
+++ b/src/Shared/Microsoft.DotNet.Scaffolding.Helpers/General/DotnetCommands.cs
@@ -10,8 +10,6 @@ public static class DotnetCommands
     {
         if (!string.IsNullOrEmpty(packageName))
         {
-            var errors = new List<string>();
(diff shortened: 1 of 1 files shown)


src/Shared/Microsoft.DotNet.Scaffolding.Helpers/General/DotnetCommands.cs
```diff
@@ -32,8 +30,15 @@ public static void AddPackage(string packageName, ILogger logger, string? projec
...
             if (exitCode != 0)
             {
+                if (!string.IsNullOrWhiteSpace(stdOut))
+                {
+                    logger.LogMessage($"\n{stdOut}");
+                }
+                if (!string.IsNullOrWhiteSpace(stdErr))
+                {
+                    logger.LogMessage($"\n{stdErr}");
+                }
```
deepchoudhery, line 40: nit: newline after line 40

deepchoudhery (approved):
LGTM!

PR #2756, opened 2024-05-18
Stream output from tools

tlmii:
- Changes the output from the dependent tools to be displayed as it happens rather than all at once when it exits.
- Tweaks the output of the aspire tool slightly just to clean things up a bit

This is what it looks like at the end now (it has the status spinner thing while it's working, and these come in one at a time now):

image


deepchoudhery (approved):
LGTM!

PR #2758, opened 2024-05-19
Relative path for project selection

tlmii:
The primary purpose of this PR is to change the display of the project picker so that it shows the relative path to the project instead of the full path. At least for demo purposes I think this is cleaner. We can discuss how we want this all displayed long-term later.

Before:
image

After:
image

Along the way I also changed out the use of the Tuple to a concrete class, which makes the code a little easier to reason with. We'll have more changes here in the future so I'm not sure that class is permanent but I think it makes this more readable in the interim.


deepchoudhery (approved):
LGTM!

PR #2762, opened 2024-05-20
Escape tool output

tlmii:
Per the [docs](https://spectreconsole.net/markup):

### Escaping Interpolated Strings
When working with interpolated strings, you can use the `MarkupInterpolated` and `MarkupLineInterpolated` methods to automatically escape the values in the interpolated string "holes".

The output from the tools often contains `[...]` sequences, which throws off the markup. This escapes that output.


deepchoudhery (approved):
LGTM! nice find

PR #2763, opened 2024-05-20
Add --prerelease option

tlmii:
- Adds a yes/no picker to keep in the "flow" of moving up/down through the options that we have for all the other selections rather than having to move to the y/n keys to answer the question with a confirmation prompt.
- Uses the yes/no picker to add a `--prerelease` option for each of the aspire commands that determines if we install the latest prerelease package (if you say Yes) or the latest release package (if you say No).


deepchoudhery (approved):
LGTM!

PR #2766, opened 2024-05-20
Remove -pt images from 6.0 branch

tlmii:
The `-pt` images were removed once they were redundant with the base images. We should no longer reference them. This change was already made in main but needed to be made in servicing branches as well.


deepchoudhery (approved):
LGTM!

PR #2776, opened 2024-05-30
Change PR builds to use macos-13 image

tlmii:
MacOS 11 was EOL in September 2023, s we should probably move to newer supported machines.


akoeplinger:
@deepchoudhery FYI Azure DevOps will remove macos-11 images in about two weeks so this should be backported to the servicing branches.

PR #2780, opened 2024-05-31
Bring back strong naming for dotnet-scaffolding packages

tlmii:
This gets around Spectre.Console not being strong named while still keeping our stuff strong named. Since strong naming doesn't provide any actual value in .NET (Core/5+) and we don't ship this for .NET Framework, there shouldn't be any harm in ignoring these warnings.


deepchoudhery (approved):
awesome, good to know.

PR #2783, opened 2024-05-31
cleaning up configs regarding dotnet-scaffold

deepchoudhery:
changing versioning for all projects in `src\dotnet-scaffolding`
- relying on the common `Versions.props`, all packages for `dotnet-scaffold` will be 9.0.0-preview.X and 9.0.x when we GA.
- removed `Versions.Scaffold.props`, `Versions.Scaffold.Aspire.props`, and `Versions.Scaffold.AspNet.props` and their references
- bumped preview version to 5


tlmii (approved):
Looks good!

PR #2792, opened 2024-06-06
Remove encoding override

tlmii:
Spectre.Console has [logic in it](https://github.com/spectreconsole/spectre.console/blob/5c87d7fa04424faadf54686f92709af277ec28b3/src/Spectre.Console/AnsiConsoleFactory.cs#L31) to detect the output encoding of the shell being used, but we had a line in AnsiConsoleLogger that always overrode that to make it be Utf8. That worked ok in a lot of shells that were Utf8 natively (like powershell core) but some (e.g. Windows Powershell) did not have Utf8 enabled by default and this led to rendering bugs like this:

image

Removing this one line appears to allow the default detection through


deepchoudhery (approved):
LGTM!

PR #2797, opened 2024-06-10
fix project discovery (for macOS and linux)

deepchoudhery:
Duplicate project files were being displayed when using `InteractivePickerType.ProjectPicker`
- project files discovered rom `.sln` were not being normalized (based on platform) and duplicates were being added.
- fixed scripts for installing tools on macOS and linux.

### DIFF
--- a/src/dotnet-scaffolding/dotnet-scaffold/Flow/Steps/ParameterDiscovery.cs
+++ b/src/dotnet-scaffolding/dotnet-scaffold/Flow/Steps/ParameterDiscovery.cs
@@ -266,8 +266,8 @@ namespace Microsoft.DotNet.Tools.Scaffold.Flow.Steps
             {
                 var normalizedPath = StringUtil.NormalizePathSeparators(projectPath);
                 var projectName = Path.GetFileName(normalizedPath);
-                //not using case comparison, seems like macOS/Linux project names in the .sln file are not case sensitive. 
-                if (!baseProjectNames.Contains(projectName))
+                //not making a case-sensitive comparison, seems like macOS/Linux project names in the .sln file are not case sensitive. 
+                if (!baseProjectNames.Contains(projectName, StringComparer.OrdinalIgnoreCase))
                 {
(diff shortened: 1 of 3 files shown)


src/dotnet-scaffolding/dotnet-scaffold/Flow/Steps/ParameterDiscovery.cs
```diff
@@ -258,6 +259,22 @@ internal IList<string> GetProjectsFromSolutionFiles(List<string> solutionFiles,
...
         }
 
+        internal List<string> AddUniqueProjects(List<string> baseList, List<string> projectPathsToAdd)
+        {
+            var baseProjectNames = baseList.Select(x => Path.GetFileName(x));
+            foreach(var projectPath in projectPathsToAdd)
+            {
+                var normalizedPath = StringUtil.NormalizePathSeparators(projectPath);
+                var projectName = Path.GetFileName(normalizedPath);
+                if (!baseProjectNames.Contains(projectName))
```
tlmii, line 269: We should be explicit about what Contains is going to do here. That probably means setting up BaseProjectNames as a List<string> with a specified comparer first, so you know it'll do the appropriate comparison. One question would be what is the appropriate comparison... technically linux/mac are case-sensitive and windows is case-insensitive.

tlmii (changes requested):

src/dotnet-scaffolding/dotnet-scaffold/Flow/Steps/ParameterDiscovery.cs
deepchoudhery: oh yeah totally forgot unix being case-sensitive. I think we should make this case-sensitive, thoughts?

deepchoudhery: added comment, keeping it case-insensitive

PR #2801, opened 2024-06-11
Remove redundant msbuild props

tlmii:
Thes are already set in Versions.props: 

https://github.com/dotnet/Scaffolding/blob/aa14b4c5fa67c21cd1093d242cd7a537a5317981/eng/Versions.props#L3-L7

which is imported first into the same projects as part of the default arcade imports.. Essentially the things in this file (Versions.MSIdentity.props) just override a few things in Versions.props.

Just something small I noticed while preparing for Centralized Package Management


deepchoudhery (approved):
LGTM!

PR #2802, opened 2024-06-11
Update CodeAnalysis packages, some cleanup

tlmii:
A few small tweaks:

- Update the `Microsoft.CodeAnalysis.*` references to `4.10.0` final which was just released yesterday
- Consolidate `Microsoft.CodeAnalysis.*` versioning references so all use the same version and can be updated together. Note that I left them as separate entries (for now at least)
- Remove some unused package version information in `Versions.props`
- Updated one reference in `src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Microsoft.DotNet.MSIdentity.csproj` that was a dependency of the `Microsoft.CodeAnalysis.*` packages
- Cleaned up the `src/MSIdentityScaffolding/Microsoft.DotNet.MSIdentity/Microsoft.DotNet.MSIdentity.csproj` a bit to separate out the package refs, project refs, folder includes, etc.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -7,7 +7,6 @@
   </PropertyGroup>
   <!-- Production Dependencies -->
   <PropertyGroup>
-    <MicrosoftBuildFrameworkPackageVersion>$(MicrosoftBuildPackageVersion)</MicrosoftBuildFrameworkPackageVersion>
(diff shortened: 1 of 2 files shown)


deepchoudhery (approved):
LGTM!

PR #2817, opened 2024-06-19
Implement central package management

tlmii:
This started off as updating the versions of some dependencies from .NET 9 Preview 1 to .NET 9 Preview 5 because we were going to add a few more in an upcoming PR and wanted to add the latest. But doing so caused version conflicts and that seemed like a good time to go ahead and get CPM out of the way.


deepchoudhery (approved):
LGTM! tyty

PR #2821, opened 2024-06-21
fixing MSBuild initialization

deepchoudhery:
instead of using `MSBuildLocator.RegisterDefaults()`, using `MSBuildLocator.RegisterMSBuildPath()`
- `RegisterDefaults()` registers the newest version of MSBuild that ships with the newest version of VS. This is not necessarily in line with the version of .NET running on commandline 
- we run dotnet commands on the commandline so would like to register MSBuild off of the newest sdk
- find the newest sdk path on disk (using default sdk locations for now, will adapt to custom install locations) and then use `RegisterMSBuildPath`.


src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Helpers/Services/MsBuildInitializer.cs
```diff
@@ -1,19 +1,81 @@
...
         {
-            MSBuildLocator.RegisterDefaults();
+            // Path to the directory containing the SDKs
+            string sdkBasePath = GetDefaultSdkPath();
+            //register newest MSBuild from the newest dotnet sdk installed.
+            var sdkPath = Directory.GetDirectories(sdkBasePath)
+                                  .OrderByDescending(d => new DirectoryInfo(d).Name)
+                                  .FirstOrDefault();
+
+            if (!Directory.Exists(sdkBasePath) || string.IsNullOrEmpty(sdkPath))
```
BillHiebert, line 36: You check for the existence of sdkBasePath after you called Directory.GetDirectories() on it. Should this be called first? I think Directories.GetDirectories() will throw if the path doesn't exist

deepchoudhery: yep not sure why I tried to combine these checks, separated them.

BillHiebert (approved):

PR #2822, opened 2024-06-21
added NOTICE.txt file to project root

deepchoudhery:
Just created one file for the root of the project. Also they seem to be the same for all branches since it's based on the pipeline itself. 
- will just backport it to all GA branches after merge.


tlmii (approved):
I believe we should exclude the things that we don't ship from this file.

PR #2834, opened 2024-06-28
switch AnsiConsole.Status to calling StartAsync instead of Start

deepchoudhery:
was causing an issue where the call `MSBuildWorkspace.OpenProjectAsync` was blocking the main thread and the AnsiConsole.Status progress UI was dropped. Switching to `AnsiConsole.Status.StartAsync` to allow async Tasks. Fixed the issue!


src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Helpers/Services/CodeService.cs
```diff
@@ -126,7 +126,7 @@ public void Dispose()
         UnloadWorkspace();
     }
 
-    public async Task<IList<ISymbol>> GetAllClassSymbolsAsync()
+    public async Task<List<ISymbol>> GetAllClassSymbolsAsync()
```
tlmii, line 129: We'll want to API review this if this class (or a derivative of it) gets opened up as part of the public API. For now this is fine.

tlmii (approved):

PR #2835, opened 2024-07-01
added code changes in json config and using IDbContextFactory now

deepchoudhery:
- updated `blazorWebCrudChanges.json` to include some additional code changes.
- updated all blazor crud T4 templates to use `IDbContextFactory` instead of `DbContext`
- updated `IEntityFrameworkService.cs` and relating files to allow optional use of `AddDbContextFactory` instead of `AddDbContext`
- added use of `ApplyTextReplacements` to `BlazorWebCRUDGenerator.cs` to allow 'Replacements' in the code modifier configs.
- removed `ModifyProgramCsTests` since the test is invalid due to above change.


tlmii (approved):
Looks reasonable from my perspective

PR #2866, opened 2024-07-11
added AddNewDbContextStep

deepchoudhery:
added `AddNewDbContextStep` 
- inherits from `ScaffoldStep`
- currently in `Microsoft.DotNet.Scaffolding.Helpers` but will potentially move this to an internal library (only to be used by 1st party tools), unsure as to if this is a good step to expose. Thoughts @tlmii?

minor updates : 
- added `dotnet-scaffold-aspire/Helpers/DatabaseProperties.cs` to separate aspire database properties from `DbContextProperties` (should only have properties relating to creating a new DbContext)


tlmii (approved):
LGTM. As far as exposing the step or not... I think its a good idea for us to list all of our intended steps and bring that to @sayedihashimi and see if we can come up with a clear idea of what's really "general purpose" vs "special purpose". This seems like it is probably reusable. But we can also err on the side of making it internal initially and opening it up later as we're more confident in the API/shape

PR #2874, opened 2024-07-16
added CodeChangeStep

deepchoudhery:
added `CodeChangeStep`
- still using `ProjectModifier.ExecuteAsync()` for applying code changes from a `CodeModifierConfig` to files in a project
- using a very simple variable injection, using the `CodeModifierProperties` dictionary as part of the step
  - simply go through all the `CodeModifierConfig.Files.Methods.CodeChanges` and replace keys with values
- using it in all the `dotnet-scaffold-aspire` commands currently
  - moved `GetAutoGeneratedProjectNamesAsync()` logic to `AspireHelpers.GetAutoGeneratedProjectNamesAsync()`
  - had to edit the logic since the roslyn workspace was not loading all the `*.ProjectMetadata.g.cs` files so finding them more manually. Will need a better fix (will search for MSBuild variables that will help with this)
- will have a PR soon for incorporating this step into `dotnet-scaffold-aspnet` commands
- minor edits to `ProjectModifier`


src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Helpers/Steps/CodeChangeStep.cs
```diff
@@ -0,0 +1,75 @@
...
+            codeSnippet.Parent = ReplaceString(codeSnippet.Parent);
+            codeSnippet.Block = ReplaceString(codeSnippet.Block) ?? string.Empty;
+        }
+    }
+
+    private string? ReplaceString(string? input)
+    {
+        if (string.IsNullOrEmpty(input))
+        {
+            return null;
```
tlmii, line 65: Not sure it matters in practice, but would it make sense to maintain the original value here (e.g. return input rather than null)? If you did that you could also potentially flip this if check to a ! and move the foreach inside it and just have one return on input since you'd be returning the same thing in all cases? Just a suggestion.

tlmii (approved):
I know you're going to rebase this on mine before merging, but other than one suggestion this looks good to me. Assuming the rebase goes fine, approved!

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Helpers/Steps/CodeChangeStep.cs
deepchoudhery: I like just returning input here. Not a fan of nesting if I can avoid it.

tlmii (approved):

PR #2877, opened 2024-07-18
Small builder cleanup/docs changes

tlmii:
Small follow up to #2875 

- Adds some doc comments
- Deletes some unused code and unused usings
- Tweaks the way scaffolder name is stored so it can more reliably be used as a command line command (e.g. can't contain spaces).


src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldBuilder.cs
```diff
@@ -74,4 +74,7 @@ public IScaffolder Build(IServiceProvider serviceProvider)
 
         return new Scaffolder(Name, DisplayName, Category, Description, _options, steps, _stepPreparers);
     }
+
+    private static string FixName(string name)
```
tlmii, line 78: Similar to the option naming, we'll want to flesh this out and add tests around it.

deepchoudhery: sounds good

deepchoudhery (approved):
LGTM

PR #2879, opened 2024-07-18
Quick dotnet-scaffold-aspire cleanup

tlmii:
- Removed TypeRegistrar and TypeResolver as they weren't being used anymore (we're using normal .NET DI now)
- Removed SpectreExtensions since we're not doing a spinner from within dotnet-scaffold-aspire
- Remove and Sort usings


deepchoudhery (approved):
LGTM

PR #2887, opened 2024-07-23
adding internal `ScaffolderSteps` for dotnet-scaffold-aspnet scaffolders

deepchoudhery:
- added `ScaffoldStep.SkipStep`
- added `AddConnectionStringStep.cs`
  - given a connection string name and value, add it to the `appsettings.json`, very simple use-case rn
- added `WithConnectionStringStep` to  `dotnet-scaffold-aspnet\ScaffolderBuilderExtensions.cs` which uses `AddConnectionStringStep`
- removed `AddNewDbContextStep.cs` and moved logic to `dotnet-scaffold-aspnet\ScaffolderBuilderExtensions.WithDbContextStep`
- added `ScaffolderContext` parameter to `ICommandWithSettings.ExecuteAsync`
- reverted `ProjectModifier` to having `CodeChangeOptions` as a parameter (interim change, will be reworked in next roslyn-focused PR)
- using new `IScaffoldBuilder` extensions in `Program.cs`


src/dotnet-scaffolding/dotnet-scaffold-aspnet/ScaffolderBuilderExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+        var allT4TemplatePaths = new TemplateFoldersUtilities().GetAllT4Templates(["BlazorCrud"]);
+        foreach (var templatePath in allT4TemplatePaths)
+        {
+            var templateName = Path.GetFileNameWithoutExtension(templatePath);
+            var templateType = BlazorCrudHelper.GetTemplateType(templatePath);
+            if (!string.IsNullOrEmpty(templatePath) && templateType is not null)
+            {
+                builder = builder.WithStep<TextTemplatingStep>(config =>
+                {
+                    config.Step.SkipStep = false;
```
deepchoudhery, line 72: having to reset the `SkipStep` variable because if set to true, it remains true for cascading steps. So if I have 3 `TextTemplatingStep` steps chained together in Program.cs and set `TextTemplatingStep1.SkipStep = true`, the 2nd and 3rd steps default to true as well. Why is that? Assuming I'm setting it up wrong. Is it due to the singleton nature of initialization? @tlmii

deepchoudhery: using `AddTransient` in the correct spots

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Helpers/ScaffolderBuilderExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+internal static class ScaffolderBuilderExtensions
+{
+    /// <summary>
+    /// Adds a new DbContext class 
+    /// </summary>
+    /// <param name="builder"></param>
+    /// <returns></returns>
+    public static IScaffoldBuilder WithDbContextStep(
+        this IScaffoldBuilder builder,
+        DbContextProperties? dbContextProperties = null)
```
tlmii, line 19: `dbContextProperties` doesn't appear to be passed in anymore in any of the cases where this is currently called. Can it be removed as a parameter? Or do you know for sure it'll be used in future uses?

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Helpers/ScaffolderBuilderExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+            step.TemplateModel = dbContextProperties;
+            step.OutputPath = dbContextProperties.DbContextPath;
+        });
+
+        return builder;
+    }
+
+    public static IScaffoldBuilder WithConnectionStringStep(
+        this IScaffoldBuilder builder,
+        string? baseProjectPath = null,
```
tlmii, line 52: Same question about `baseProjectPath` and `dbContextProperties` here as I had about `dbContextProperties` in the other extension method - doesn't appear to be used in any of the current calls.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.TextTemplating/ScaffolderBuilderExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    public static IScaffoldBuilder WithTextTemplatingStep(
+        this IScaffoldBuilder builder,
+        string templateFilePath,
+        Type templateType,
+        string templateModelName,
+        object templateModel,
+        string outputPath)
+    {
+        return builder.WithStep<TextTemplatingStep>(config =>
```
tlmii, line 17: It doesn't seem like this is used? And it doesn't remove as much extra code as the others do. Consider cutting it?

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.TextTemplating/TextTemplatingStep.cs
```diff
@@ -1,28 +1,25 @@
...
-    public required string OutputPath { get; init; }
+    public required Type TemplateType { get; set; }
+    //output file path where the templated content will be written (should include the extension if one is wanted)
+    public required string OutputPath { get; set; }
     //the 'name' property of <#@ parameter #> in provided .tt template
-    public required string TemplateModelName { get; init; }
+    public required string TemplateModelName { get; set; }
     //the 'type' property of <#@ parameter #> in provided .tt template
-    public required object TemplateModel { get; init; }
-    public required IFileSystem FileSystem { get; init; }
```
tlmii, line 23: These should be able to be imported still since the step comes from DI. Just add a constructor to TextTemplatingStep that takes them as parameters. Then you can undo some of the other changes in this file I think.

src/dotnet-scaffolding/dotnet-scaffold-aspnet/Program.cs
```diff
@@ -147,17 +155,17 @@ public static void Main(string[] args)
 
         var runner = builder.Build();
         runner.RunAsync(args).Wait();
-        //    config.AddCommand<AreaCommand>("area");
-
     }
 
     static void ConfigureServices(IServiceCollection services)
```
tlmii, line 160: We should probably consider having a separate AddSteps method (or something similar) specifically for adding all the steps to the collection that's separate from the ConfigureServices just to keep the semantic intent obvious.

src/dotnet-scaffolding/dotnet-scaffold-aspnet/ScaffolderBuilderExtensions.cs
```diff
@@ -0,0 +1,99 @@
...
+using Microsoft.DotNet.Tools.Scaffold.AspNet.Helpers;
+
+namespace Microsoft.DotNet.Scaffolding.Core.Builder;
+
+internal static class ScaffolderBuilderExtensions
+{
+    public static IScaffoldBuilder WithMinimalApiTextTemplatingStep(this IScaffoldBuilder builder)
+    {
+        var allT4Templates = new TemplateFoldersUtilities().GetAllT4Templates(["MinimalApi"]);
+        string? t4TemplatePath = null;
```
tlmii, line 19: Do these lines need to be outside the WithStep call? Thinking about it, calling GetAllT4Templates now would enumerate the files every time the tool ran, even when just running `get-commands` or whatever. If it was inside the WithStep call, it'd be deferred until execution.

src/dotnet-scaffolding/dotnet-scaffold-aspnet/ScaffolderBuilderExtensions.cs
```diff
@@ -0,0 +1,99 @@
...
+            step.TemplateModelName = "Model";
+            step.OutputPath = minimalApiModel.EndpointsPath;
+        });
+        
+    }
+
+    public static IScaffoldBuilder WithBlazorCrudTextTemplatingStep(
+        this IScaffoldBuilder builder)
+    {
+        var allT4TemplatePaths = new TemplateFoldersUtilities().GetAllT4Templates(["BlazorCrud"]);
```
tlmii, line 62: Same question as above

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Scaffolders/Scaffolder.cs
```diff
@@ -39,9 +39,12 @@ public async Task ExecuteAsync(ScaffolderContext context)
         {
             var step = _steps[stepIndex];
             var preparer = _preparers[stepIndex];
-
             preparer.RunPreExecute(step, context);
-            await step.ExecuteAsync(context);
+            if (!step.SkipStep)
```
tlmii, line 43: I'm still not entirely sure about the SkipStep property. But we can iterate.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Steps/ScaffoldStep.cs
```diff
@@ -9,4 +9,5 @@ public abstract class ScaffoldStep
 {
     public abstract Task ExecuteAsync(ScaffolderContext context, CancellationToken cancellationToken = default);
     public bool ContinueOnError { get; set; }
+    public bool SkipStep { get; set; } = false;
```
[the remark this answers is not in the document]
deepchoudhery: yep I do this by habit for some reason, removing.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Helpers/General/DbContextProperties.cs
```diff
@@ -5,6 +5,6 @@ internal class DbContextProperties
 {
     public string DbContextName { get; set; } = "NewDbContext";
     public string? DbContextPath { get; set; }
-    public string? DbSetStatement { get; set; }
+    public string? DbSetStatement { get; set; }
```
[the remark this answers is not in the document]
deepchoudhery: fixed ty

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Scaffolders/Scaffolder.cs
deepchoudhery: sounds good

src/dotnet-scaffolding/dotnet-scaffold-aspnet/ScaffolderBuilderExtensions.cs
deepchoudhery: moved these inside the step since only one `WithStep` is needed.

src/dotnet-scaffolding/dotnet-scaffold-aspnet/Program.cs
deepchoudhery: yup sounds good, adding a TODO there.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Helpers/Microsoft.DotNet.Scaffolding.Helpers.csproj
```diff
@@ -20,5 +20,6 @@
   <ItemGroup>
     <ProjectReference Include="$(RepoRoot)src\dotnet-scaffolding\Microsoft.DotNet.Scaffolding.ComponentModel\Microsoft.DotNet.Scaffolding.ComponentModel.csproj" />
     <ProjectReference Include="$(RepoRoot)src\dotnet-scaffolding\Microsoft.DotNet.Scaffolding.TextTemplating\Microsoft.DotNet.Scaffolding.TextTemplating.csproj" />
+    <ProjectReference Include="$(RepoRoot)src\dotnet-scaffolding\Microsoft.DotNet.Scaffolding.Core\Microsoft.DotNet.Scaffolding.Core.csproj" />
```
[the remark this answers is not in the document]
deepchoudhery: yup will do ty

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Helpers/ScaffolderBuilderExtensions.cs
deepchoudhery: yep removed, we can iterate on this later. I think I try to anticipate too much flexibility on what other devs will want.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Helpers/ScaffolderBuilderExtensions.cs
deepchoudhery: yep removed as well

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.TextTemplating/TextTemplatingStep.cs
deepchoudhery: but IFileSystem comes from the `Helpers` library which is not referenced in this project? I was avoiding adding that reference until our dependency stuff is sorted (next PR)

src/dotnet-scaffolding/dotnet-scaffold-aspnet/ScaffolderBuilderExtensions.cs
deepchoudhery: easier in the `WithMinimalApiTextTemplatingStep` but harder here since we are returning builder with multiple steps. Unless I can add steps to the builder from within an WithStep pre execute block. Any thoughts?

tlmii: Ah yes, you're right, my bad. We may want to consider looking at this again later (perhaps some way to have the Step itself expand this and iterate within itself?) to avoid the perf hit on get-command, --version, etc calls. But you're right I misread it as being similar to the last one.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.TextTemplating/TextTemplatingStep.cs
tlmii: Ah, yeah, you're right about IFileSystem. ILogger should be fine though. Maybe leave a TODO comment about the FileSystem so we don't forget and put ILogger back in?

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.TextTemplating/ScaffolderBuilderExtensions.cs
tlmii: I think this is the only thing that wasn't changed/commented on? Do you think its important to leave it?

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.TextTemplating/TextTemplatingStep.cs
tlmii: (note the import in the constructor here would be `ILogger<TextTemplatingStep>`, not just `ILogger`, though you can store it as an ILogger if you want, it doesn't really matter other than for import.

deepchoudhery: yep I already have those changes stashed actually, was avoiding adding it in this PR and will do it together for all logging.

src/dotnet-scaffolding/dotnet-scaffold-aspnet/ScaffolderBuilderExtensions.cs
deepchoudhery: yes I will definitely want to rework this but couldn't come up with anything better rn. I'll sit with and a TODO to make this better.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.TextTemplating/ScaffolderBuilderExtensions.cs
deepchoudhery: removed yes

tlmii (approved):

PR #2889, opened 2024-07-23
ILogger Refactoring + Cleanup

tlmii:
The primary thing here is removing the old AnsiConsoleLogger ILogger implementation and replacing it with a Microsoft.Extensions.Logging.ILogger implementation. I created a CleanConsoleFormatter logger implementation that is _very_ rudimentary for now that we can extend as we need to. Then I plumbed through the Microsoft.Extensions.Logging.ILogger to all the places the old ILogger was used and removed all references to Spectre.Console from dotnet-scaffold-aspire and dotnet-scaffold-aspnet. 

Along the way I did a few other things:
1. Added a new install-scaffold-all script that builds/installs all three of the new packages at once
2. Enabled Implicit Usings on dotnet-scaffold to match .Core because I needed to share some code there. We should do the same on -aspnet and -aspire but I'll do that in another PR
3. Sorted removed/usings on all files I could find that were impacted by any of the above changes.

There should be no real user-facing changes to how things work by this PR but it'll enable us to have a better multi-targeted logging story in the future.


src/dotnet-scaffolding/dotnet-scaffold/dotnet-scaffold.csproj
```diff
@@ -8,19 +8,30 @@
     <PackageTags>dotnet;scaffold</PackageTags>
     <PackageId>Microsoft.dotnet-scaffold</PackageId>
     <RootNamespace>Microsoft.DotNet.Tools.Scaffold</RootNamespace>
+    <ImplicitUsings>enable</ImplicitUsings>
     <NoWarn>$(NoWarn);8002</NoWarn>
   </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="..\Microsoft.DotNet.Scaffolding.Core\Logging\CleanConsoleFormatter.cs" Link="Logging\CleanConsoleFormatter.cs" />
```
deepchoudhery, line 15: Is this for the interim, wondering if this will complicate things in the future as these expand?

deepchoudhery (approved):
LGTM, one question

src/dotnet-scaffolding/dotnet-scaffold/dotnet-scaffold.csproj
tlmii: Yeah, sorry, should have clarified that. The only thing "shared" between `dotnet-scaffold` and `.Core` is the `ComponentModel` library, but that's not the right place for this. I made it a shared file for the moment to get around that (since we don't need binary compat) but when we have further discussions about what exactly our package/library layout looks like we'll need to discuss where something might live that is shared between these layers.

tlmii: I added a comment to the .csproj file to hopefully remind us to change that

deepchoudhery: cool sgsg

PR #2894, opened 2024-07-26
updated AddPackagesStep and AddConnectionStringStep

deepchoudhery:
- updated `AddPackagesStep` to inherit from the `Microsoft.DotNet.Scaffolding.Core.Steps.ScaffoldStep`
- updated `AddConnectionStringStep` to use System.Text.Json instead of Newtonsoft.Json.
- added `ValidateContextStep` to allow easy validation by passing a validate function to this step (we can rework this)
- added `ValidationHelper.cs` to help with dotnet-scaffold-aspire cli option validation (moved out of the commands)
- added `ScaffolderContext.GetOptionResult(string optionCliName` for getting option results with just the cli flags.
- removed `IAppSettings` use from `CodeService`. All we really needed from it was the project path so added that as readonly variable on the class.
- changed `dotnet-scaffold-aspire\Commands\GetCmdsHelper.cs` to `AspireCommandHelpers.cs`
- removed `dotnet-scaffold-aspire` add package logic from the commands to `ScaffolderBuilderExtensions.cs` and using `AddPackagesStep` with `IScaffolderBuilder`
- same with `dotnet-scaffold-aspnet` commands.


src/dotnet-scaffolding/dotnet-scaffold-aspire/ScaffoldSteps/ValidateContextStep.cs
```diff
@@ -0,0 +1,34 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.DotNet.Scaffolding.Core.Scaffolders;
+using Microsoft.DotNet.Scaffolding.Core.Steps;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DotNet.Tools.Scaffold.Aspire.ScaffoldSteps;
+
+public class ValidateContextStep : ScaffoldStep
```
tlmii, line 13: What do you think about ValidateOptionsStep as a name instead? At the point this would execute, that's the only part of the context that would be relevant (the step is just itself, the property bag would be empty since no other steps have run). And really that's what you're using it for.

src/dotnet-scaffolding/dotnet-scaffold-aspire/ScaffoldSteps/ValidateContextStep.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        _logger = logger;
+    }
+
+    public override Task ExecuteAsync(ScaffolderContext context, CancellationToken cancellationToken = default)
+    {
+        if (!ValidateMethod(context, _logger))
+        {
+            _logger.LogError("Validation failed.");
+            return Task.FromResult(false);
```
tlmii, line 28: ScaffoldStep.ExecuteAsync doesn't actually return a value so that result doesn't go anywhere. We need to either change the signature to return Task<bool> (or some other way to indicate a result) or throw an exception. Probably Task<bool> I guess. And then we need to actually check the result in Scaffolder.ExecuteAsync.

tlmii (commented):
I'm not sure about the Validation step idea. I _do_ think its a good idea to have something specific to enable (and encourage!) validation. There are some other options we could consider:

1. An explicit validation procedure outside of the normal step flow. 
2. Similar to what you have, but instead make the ExecuteAsync abstract and have people override it with their own validation implementation. I'm not thrilled with that because they'd need to add the validation to services for each of their validators. 

So I think this is a good start. But we might want to see what we think after a couple more iterations.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Helpers/Steps/AddConnectionStringStep.cs
```diff
@@ -55,16 +63,11 @@ public override Task ExecuteAsync(ScaffolderContext context, CancellationToken c
...
         }
 
-        // Json.Net loses comments so the above code if requires any changes loses
-        // comments in the file. The writeContent bool is for saving
-        // a specific case without losing comments - when no changes are needed.
         if (writeContent && !string.IsNullOrEmpty(appSettingsFile))
         {
-            File.WriteAllText(appSettingsFile, content.ToString());
+            var options = new JsonSerializerOptions { WriteIndented = true };
+            await File.WriteAllTextAsync(appSettingsFile, content.ToJsonString(options));
```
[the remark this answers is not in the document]
deepchoudhery: yep will switch to IFIleSystem.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Helpers/Steps/AddConnectionStringStep.cs
```diff
@@ -16,24 +17,31 @@ internal class AddConnectionStringStep : ScaffoldingStep
...
         if (string.IsNullOrEmpty(appSettingsFile) || !File.Exists(appSettingsFile))
         {
-            content = [];
+            content = new JsonObject();
             writeContent = true;
         }
         else
         {
-            content = JObject.Parse(File.ReadAllText(appSettingsFile));
+            var jsonString = await File.ReadAllTextAsync(appSettingsFile, cancellationToken);
```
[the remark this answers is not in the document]
deepchoudhery: yep will switch to IFIleSystem.

src/dotnet-scaffolding/dotnet-scaffold-aspire/ScaffoldSteps/ValidateContextStep.cs
deepchoudhery: yep thats a better name for sure.

src/dotnet-scaffolding/dotnet-scaffold-aspire/ScaffoldSteps/ValidateContextStep.cs
deepchoudhery: returning bool on steps now and checking for ContinueOnError as well

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Scaffolders/Scaffolder.cs
```diff
@@ -15,7 +14,7 @@ public class Scaffolder : IScaffolder
     private readonly List<ScaffolderOption> _options;
     private readonly List<ScaffoldStep> _steps;
     private readonly List<ScaffoldStepPreparer> _preparers;
-
+
```
tlmii, line 17: nit: extra spaces

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Scaffolders/Scaffolder.cs
```diff
@@ -42,7 +41,11 @@ public async Task ExecuteAsync(ScaffolderContext context)
...
             {
-                await step.ExecuteAsync(context);
+                var stepResult = await step.ExecuteAsync(context);
+                if (!stepResult && !step.ContinueOnError)
+                {
+                    break;
+                }
             }
 
             preparer.RunPostExecute(step, context);
```
tlmii, line 51: Probably a discussion we should have at some point about whether the PostExecute should run (with a failed flag?) even if there was an error. Not for now though

src/dotnet-scaffolding/dotnet-scaffold-aspire/Program.cs
```diff
@@ -96,7 +96,7 @@
 //TODO separate adding transient steps from singleton services.
 static void ConfigureServices(IServiceCollection services)
 {
-    services.AddTransient<ValidateContextStep>();
+    services.AddTransient<ValidateOptionsStep>();
```
tlmii, line 99: One thing to consider if we do keep the ValidateOptionsStep concept, is:
1. Move it into Core
2. Register it in services as part of the core services so that users don't need to register it themselves to use it.

tlmii (approved):

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Scaffolders/Scaffolder.cs
deepchoudhery: fixed ty

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Scaffolders/Scaffolder.cs
deepchoudhery: ok added a TODO for now

src/dotnet-scaffolding/dotnet-scaffold-aspire/Program.cs
deepchoudhery: ok added a TODO for now

PR #2898, opened 2024-07-29
refactoring dotnet-scaffolding part 1

deepchoudhery:
- removed `Microsoft.DotNet.Scaffolding.ComponentModel`
  - moved it to `Microsoft.DotNet.Scaffolding.Core\ComponentModel`
  - `dotnet-scaffold.csproj` references these files using Compile includes.
- moved `DotnetCliRunner.cs` and `DotnetCommands.cs` to `Microsoft.DotNet.Scaffolding.Core\CliHelpers`
- moved `IFileSystem` and `IEnvironmentService` to `Microsoft.DotNet.Scaffolding.Core\Services`
- moved `AddPackagesStep.cs` to `Microsoft.DotNet.Scaffolding.Core\Steps`
- removed unused services from `Microsoft.DotNet.Scaffolding.Services`
- removed unused project references from all dotnet-scaffolding projects


src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/CliHelpers/DotnetCliRunner.cs
```diff
@@ -2,12 +2,12 @@
...
 using System.Diagnostics;
 
-namespace Microsoft.DotNet.Scaffolding.Helpers.General;
+namespace Microsoft.DotNet.Scaffolding.Core;
 
 /// <summary>
 /// To run 'dotnet' or any processes and capture consequent stdout and stderr (using 'ExecuteAndCaptureOutput'
 /// </summary>
-internal class DotnetCliRunner
+public class DotnetCliRunner
```
tlmii, line 10: Do you intend to move this to the .Internal assembly in a second refactoring PR? I'm not sure about having it be part of the public API here so wanted to check.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Services/EnvironmentService.cs
```diff
@@ -1,12 +1,11 @@
...
 using System.Runtime.InteropServices;
 
-namespace Microsoft.DotNet.Scaffolding.Helpers.Services.Environment;
-
+namespace Microsoft.DotNet.Scaffolding.Core.Services;
 /// <summary>
 /// Wrapper over System.Environment abstraction for unit testing.
 /// </summary>
-internal class EnvironmentService : IEnvironmentService
+public class EnvironmentService : IEnvironmentService
```
tlmii, line 8: These implementations really have to be public because dotnet-scaffold uses them (we could make them part of the default services in Core otherwise). But that means dotnet-scaffold currently has a dependency on Core (dotnet-scaffold -> Helpers -> TextTemplating -> Core) which will be broken with further refactoring (and because we didn't want it to have it. Then where would these live? Maybe shared files (they'd still have binary compat via the interface so I think it'd be ok?)? Not sure what exactly to do there.

src/dotnet-scaffolding/dotnet-scaffold/AppBuilder/ScaffoldCommandAppBuilder.cs
```diff
@@ -1,8 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
-
 using System.Reflection;
+using Microsoft.DotNet.Scaffolding.Core.Services;
 using Microsoft.DotNet.Scaffolding.Helpers.Services;
```
tlmii, line 4: I think this is unused now

tlmii (approved):
The changes that are in this PR look fine. The further refactoring will need to untangle the 

`dotnet-scaffold -> Helpers -> TextTemplating -> Core`

dependency that I just realized was there while reviewing this.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/CliHelpers/DotnetCliRunner.cs
deepchoudhery: Actually it should be in `Internal` and Core might reference it as a `Compile Include` since Core won't be referencing Internal

src/dotnet-scaffolding/dotnet-scaffold/AppBuilder/ScaffoldCommandAppBuilder.cs
deepchoudhery: will remove tyty

PR #2901, opened 2024-07-30
Add tool commands

tlmii:
These commands are standalone right now since they aren't plugged into anything, but it lays the foundation for being able to use them elsewhere.

These commands are implemented:
`dotnet-scaffold tool install [name]`
`dotnet-scaffold tool list`
`dotnet-scaffold tool uninstall [name]`

Install will (if required) install the tool globally and add it to the manifest. uninstall will remove it from the manifest and uninstall it globally. List prints a table that just shows the name for now, But I assume we'll add more information.

Under the hood, `IToolManager` and `IManifestService` can be used in the FirstPartyToolInitializer to do the work it needs to do there when we're ready to plug that in. Then `IManifestService` can be used to get the list of registered tools to call get-commands on (rather than just getting all dotnet tools like we do now).

The manifest file right now is _super_ simple, and intentionally so. I didn't want to lock us into much yet.


deepchoudhery (approved):
Looks good for v1, i'll try to iterate on it.

PR #2929, opened 2024-08-16
add support for 'mvccontroller-crud' and 'apicontroller-crud'

deepchoudhery:
Added `mvccontroller-crud` and `apicontroller-crud` in `dotnet-scaffold-aspnet`
- added `ApiEfController.tt` and `MvcEfController.tt` templates
- added `ValidateEfControllerStep.cs` for validation in the new scaffolders
- added `EfControllerSettings` and `EfControllerModel`
- added 'efControllerChanges.json' for the CodeChangeStep


src/dotnet-scaffolding/dotnet-scaffold-aspnet/ScaffoldSteps/ValidateEfControllerStep.cs
```diff
@@ -0,0 +1,204 @@
...
+            return null;
+        }
+        else
+        {
+            ControllerName = Path.GetFileNameWithoutExtension(ControllerName);
+        }
+
+        if (string.IsNullOrEmpty(ControllerType))
+        {
+            return null;
```
anvillan, line 110: Don't you want to log anything if `ControllerType` is null (so you return null)? You do it for the other cases, above

anvillan (approved):

src/dotnet-scaffolding/dotnet-scaffold-aspnet/ScaffoldSteps/ValidateEfControllerStep.cs
deepchoudhery: oops should be logging, will add a simple error. Since we are initializing this property, it'll be internal like message (and we always initialize this)

PR #2949, opened 2024-09-04
remove 'Microsoft.DotNet.ProjectModel' and bump NuGet.* deps, bump version

deepchoudhery:
bumping to 8.0.5


anvillan (approved):

global.json
```diff
@@ -1,21 +1,14 @@
...
+    "version": "8.0.108",
     "allowPrerelease": true,
     "rollForward": "latestMajor"
+  },
+  "tools": {
+    "dotnet": "8.0.108"
+  },
+  "msbuild-sdks": {
+    "Microsoft.DotNet.Arcade.Sdk": "8.0.0-beta.24426.2",
+    "Microsoft.DotNet.Helix.Sdk": "8.0.0-beta.24426.2"
```
tlmii, line 12: Just curious why the Helix SDK got added here now? Are we using it now when we weren't before? Just to bring up to parity with the arcade global.json (though that has NoTargets too)?

PR #2950, opened 2024-09-04
fixing ef crud controller templates

deepchoudhery:
fix apicontroller-crud and mvccontroller-crud templates for `dotnet-scaffold-aspnet`

### DIFF
--- a/src/dotnet-scaffolding/dotnet-scaffold-aspnet/Templates/EfController/ApiEfController.cs
+++ b/src/dotnet-scaffolding/dotnet-scaffold-aspnet/Templates/EfController/ApiEfController.cs
@@ -33,6 +33,9 @@ namespace Microsoft.DotNet.Tools.Scaffold.AspNet.Templates.EfController
     string dbContextName = Model.DbContextInfo.DbContextClassName;
     string entitySetName = Model.DbContextInfo.EntitySetVariableName ?? modelName;
     string modelNamespace = Model.ModelInfo.ModelNamespace;
+    string primaryKeyName = Model.ModelInfo.PrimaryKeyName;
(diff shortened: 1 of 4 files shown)


src/dotnet-scaffolding/dotnet-scaffold-aspnet/Templates/EfController/ApiEfController.cs
```diff
@@ -33,6 +33,9 @@ public virtual string TransformText()
     string dbContextName = Model.DbContextInfo.DbContextClassName;
     string entitySetName = Model.DbContextInfo.EntitySetVariableName ?? modelName;
     string modelNamespace = Model.ModelInfo.ModelNamespace;
+    string primaryKeyName = Model.ModelInfo.PrimaryKeyName;
+    string primaryKeyNameLowerInv = primaryKeyName.ToLowerInvariant();
```
tlmii, line 37: I'm a little perplexed by all the ToLowerInvariant stuff - shouldn't it really be camel casing if these are used for variable names? Is the assumption that they're all one word (e.g. `id`)? Though I note that's not a change, since the existing code already has it that way (e.g. `modelNameLowerInvariant`).

tlmii (approved):

src/dotnet-scaffolding/dotnet-scaffold-aspnet/Templates/EfController/ApiEfController.cs
deepchoudhery: yep I think I have just been assuming its one word variable names, they could end up being 'moviekey' when it should be 'movieKey'. Will fix this

PR #2953, opened 2024-09-05
add  `blazor-identity` to `dotnet-scaffold-aspnet`

deepchoudhery:
- added `blazor-identity` scaffolder to `dotnet-scaffold-aspnet`
  - templates added to `Templates\BlazorIdentity` and `Templates\Files\ApplicationUser.tt`
  - added `ValidateBlazorIdentityStep` : validation step for this scaffolder.
  - code changes in `CodeModificationConfigs\blazorIdentityChanges.json`


src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.CodeModification/CodeModificationStep.cs
```diff
@@ -96,13 +96,28 @@ private void EditCodeModifierConfig(CodeModifierConfig codeModifierConfig)
             return;
         }
 
+        //modify CodeSnippets 'CheckBlock', 'Parent', and 'Block'
```
tlmii, line 99: This whole block of code feels like something that should be encapsulated somewhere - perhaps on the `CodeModifierConfig` class itself, or on an extension method for it, or something. Just seems like a lot of detail that `CodeModificationStep` shouldn't need to know about beyond needing to call `codeModifierConfig.PerformReplacements(CodeModifierProperties)` or something like that. Would probably make it a little more reasonable to write tests for it too, since it'd be public methods on the impacted class rather than private methods in a ScaffoldStep.

eng/Versions.props
```diff
@@ -54,7 +54,7 @@
     <MonoTextTemplatingVersion>3.0.0-preview-0052-g5d0f76c785</MonoTextTemplatingVersion>
     <SpectreConsoleFlowVersion>0.5.638</SpectreConsoleFlowVersion>
     <SpectreConsoleVersion>0.47.0</SpectreConsoleVersion>
-    <MicrosoftCodeAnalysisPreviewVersion>4.11.0-2.final</MicrosoftCodeAnalysisPreviewVersion>
+    <MicrosoftCodeAnalysisPreviewVersion>4.12.0-1.final</MicrosoftCodeAnalysisPreviewVersion>
```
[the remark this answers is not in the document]
deepchoudhery: bringing it up to latest.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.CodeModification/CodeModificationStep.cs
deepchoudhery: ok yes will move this on the class `CodeModifierConfig` itself

src/dotnet-scaffolding/dotnet-scaffold-aspnet/ScaffoldSteps/ValidateBlazorIdentityStep.cs
```diff
@@ -0,0 +1,169 @@
...
+    public string? DataContext { get; set; }
+    public ValidateBlazorIdentityStep(
+        IFileSystem fileSystem,
+        ILogger<ValidateBlazorIdentityStep> logger)
+    {
+        _fileSystem = fileSystem;
+        _logger = logger;
+    }
+
+    public override async Task<bool> ExecuteAsync(ScaffolderContext context, CancellationToken cancellationToken = default)
```
tlmii, line 33: There's _some_ validation in here (in the aptly named ValidateBlazorIdentitySettings call on the first line in particular), but really most of it seems like setup (filling in codeModifierProperties, filling in dbContextProperties, adding everything to context.Properties, etc. 

Are all of these things that are used by multiple steps afterwards, thus it makes sense to have one step that pre-calculates things? I guess having that in a Validation step makes sense... though I wonder if we should have a better (more "intentional") mechanism for that. Kind of the same way we were thinking about with validation overall.

src/dotnet-scaffolding/dotnet-scaffold-aspnet/Templates/BlazorIdentity/IdentityComponentsEndpointRouteBuilderExtensions.Interfaces.cs
```diff
@@ -0,0 +1,8 @@
+using Microsoft.DotNet.Scaffolding.TextTemplating;
+
+namespace Microsoft.DotNet.Tools.Scaffold.AspNet.Templates.BlazorIdentity;
+
+public partial class IdentityComponentsEndpointRouteBuilderExtensions : ITextTransformation
```
tlmii, line 5: Can you give me the two sentence overview of what this is (and other similar classes below are) for? I see that the other side of the partial declaration comes from the generated code from the .tt files, but I'm not sure I understand the purpose of this side of it.

src/dotnet-scaffolding/dotnet-scaffold-aspnet/Program.cs
```diff
@@ -254,19 +275,20 @@ static void ConfigureServices(IServiceCollection services)
 
     static void ConfigureSteps(IServiceCollection services)
     {
-        services.AddTransient<CodeModificationStep>();
         services.AddTransient<ValidateBlazorCrudStep>();
         services.AddTransient<ValidateMinimalApiStep>();
+        services.AddTransient<ValidateBlazorIdentityStep>();
```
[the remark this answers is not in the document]
deepchoudhery: I can do this now even, would get the current executing assembly, find types inheriting from ScafffoldStep and register them

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.CodeModification/CodeModification/ProjectModifier.cs
```diff
@@ -122,6 +128,29 @@ public async Task<bool> RunAsync()
...
+            return null;
+        }
+
+        var replacements = file.Replacements?.Where(cc => ProjectModifierHelper.FilterOptions(cc.Options, toolOptions));
+        if (replacements is null || !replacements.Any())
+        {
+            return document;
+        }
+
+        //should be a safe cast as we are starting with a 'Document' type
```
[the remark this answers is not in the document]
deepchoudhery: yep this seems way easier, I was just rushing this change so didn't think it through

src/dotnet-scaffolding/dotnet-scaffold-aspnet/Program.cs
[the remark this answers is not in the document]
tlmii: I don't know if you meant now as in this PR or now as in just sometime before GA but I'd at least make it a separate PR. We can discuss whether it should be automatic or something you call (e.g. `services.RegisterSteps()` or similar) and then modify it everywhere that could use it. I also wonder if it'd be better as a source generator or with reflection. Speed isn't too important, so either one is probably fine. Though anything that avoids reflection makes me curious.

src/dotnet-scaffolding/dotnet-scaffold-aspnet/ScaffoldSteps/ValidateBlazorIdentityStep.cs
deepchoudhery: yep I would say majority of the validation of these scaffolders is the same so we should be able to pull that out and have scaffolder specific validation after. Adding a TODO for now

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.CodeModification/CodeModification/ProjectModifier.cs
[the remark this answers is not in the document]
deepchoudhery: also switched to `Task<T?> ApplyTextReplacements<T>` instead of having multiple definitions of this method for both Document types. Looks and works good.

src/dotnet-scaffolding/dotnet-scaffold-aspnet/Templates/BlazorIdentity/IdentityComponentsEndpointRouteBuilderExtensions.Interfaces.cs
deepchoudhery: yep I never really explained that mb We defined the `ITextTransformation` class to help with separation of concerns and define a simple interface we can interact with the auto-generated templated code and add on top of it (if needed) without modifying these auto-generated files. The `.Interfaces.cs` files help with the interface declaration without modifying the auto-generated template file. There might be a better way we can come up with eventually but this seems ok enough for now.

tlmii (approved):

PR #2957, opened 2024-09-06
add/edit paths where we look for dotnet sdk.

deepchoudhery:
incorrect paths were being checked for both mac and linux. For mac, the ARM path was not being checked. For linux, the path being checked was wrong.
- added the correct linux default path.
- added the correct path to look in macOS for the ARM .NET SDK. Check for this before an x64 version of the sdk because macOS is ARM based now (and x64 will eventually be deprecated)


tlmii (approved):
I find it odd that linux can have so many possible "standard" install locations, but that's just something we have to deal with, not something we can change.

deepchoudhery:
> I find it odd that linux can have so many possible "standard" install locations, but that's just something we have to deal with, not something we can change.

I did not dig deep enough potentially so will get this tested more and update this after

tlmii (approved):

PR #2975, opened 2024-09-20
installing and prioritizing local tools for dotnet-scaffold

deepchoudhery:
Currently dotnet-scaffold only deals with global tools"
- It will install the 1st party tools (`dotnet-scaffold-aspire` and `dotnet-scaffold-aspnet`) globally
- Also populates the list of commands and category by iterating over all global tools. 

Changing this to installing 1st party tools as 'local' tools:
- Allow repos to control the version of these 1st party tools 
- These 1st party tools are only relevant in the context of the repo/project so global tool list will stay clean
- When populating the list of commands, search local tools first, and then global tools. For duplicates, only track and invoke the local one.

changes description:
- most changes are in `DotNetToolService.GetCommands`
- for invoking local tools, we have to invoke the `dotnet` command, for eg. `dotnet dotnet-scaffold-aspnet` instead of just `dotnet-scaffold-aspnet`.
- minor changes to `IDotNetToolService.InstallDotNetTool` amd `IDotNetToolService.UninstallDotNetTool` to account for local scenarios.

adding some information on `global` vs. `local` tools. We should be clear on the following info understand installing and using these tools in both contexts.

| Category| global | local |
| --- | --- |--- |
| Invocation (eg. 'sample-tool'| use the command name, `sample-tool`. If the command name has `dotnet-` prefix, it can invoked with `dotnet ` instead as well. | prefix command name with 'dotnet' or 'dotnet tool run', `dotnet sample-tool` or `dotnet tool run sample-tool` from the installation directory |
| Location on disk | On Linux and macOS it's `$HOME/.dotnet/tools`, and on Windows it's `%USERPROFILE%\.dotnet\tools` | Installed in the NuGet global directory, and shim files are created in `$HOME/.dotnet/toolResolverCache` to point to the tools' location |
- upon a local tools installation, a manifest is either created or an existing one is used, the default one being at `CURRENT_PATH\.config\dotnet-tools.json`. This config consists of all the local tool versions and command names.

### DIFF
--- a/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/ComponentModel/DotnetToolInfo.cs
+++ b/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/ComponentModel/DotnetToolInfo.cs
@@ -3,14 +3,14 @@
 namespace Microsoft.DotNet.Scaffolding.Core.ComponentModel;
 
 /// <summary>
-/// Info from 'dotnet tool list -g'
+/// Info from 'dotnet tool list' and 'dotnet tool list -g'
 /// </summary>
 internal class DotNetToolInfo
 {
     public string PackageName { get; set; } = default!;
     public string Version { get; set; } = default!;
     public string Command { get; set; } = default!;
-
+    public bool IsGlobalTool { get; set; } = false;
     public string ToDisplayString()
     {
         return $"{Command} ({PackageName} v{Version})";
--- a/src/dotnet-scaffolding/dotnet-scaffold-aspnet/ScaffoldSteps/EmptyControllerScaffolderStep.cs
+++ b/src/dotnet-scaffolding/dotnet-scaffold-aspnet/ScaffoldSteps/EmptyControllerScaffolderStep.cs
@@ -32,7 +32,7 @@ internal class EmptyControllerScaffolderStep : ScaffoldStep
         var result = false;
         if (stepSettings is not null)
         {
-            return Task.FromResult(InvokeDotnetNew(stepSettings));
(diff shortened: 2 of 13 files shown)


tlmii (approved):
One thing we'll want to consider as we (fully) implement the `dotnet-scaffold tool install` experience is how to handle tools specified in an existing repo manifest when they're not yet specified in our manifest.  Do we tell users they should call dotnet tool restore first before doing our stuff? Do we call it ourselves if we see one in the directory hierarchy? The current pattern of "see if the tool is installed, install it if not, then add it to our manifest" just doesn't fit exactly right with the tool manifest (or, at minimum, there are scenarios we need to test with it, like what happens if you call a tool specified in a tool manifest that hasn't been restored? what's the error message?)

Anyway, this PR, for the big chunk it does, looks good to me.

deepchoudhery:
> One thing we'll want to consider as we (fully) implement the `dotnet-scaffold tool install` experience is how to handle tools specified in an existing repo manifest when they're not yet specified in our manifest. Do we tell users they should call dotnet tool restore first before doing our stuff? Do we call it ourselves if we see one in the directory hierarchy? The current pattern of "see if the tool is installed, install it if not, then add it to our manifest" just doesn't fit exactly right with the tool manifest (or, at minimum, there are scenarios we need to test with it, like what happens if you call a tool specified in a tool manifest that hasn't been restored? what's the error message?)
> 
> Anyway, this PR, for the big chunk it does, looks good to me.

yep we still don't really use our manifest for gathering the scaffolders yet and still rely on the full dotnet tool list, all this work would be done in that PR I imagine, soon.

PR #2985, opened 2024-10-01
Update transitive dependency for release/8.0

deepchoudhery:
also bumped version to 8.0.6

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -25,5 +25,4 @@
   <ItemGroup>
     <None Include="$(RepoRoot)NOTICE.txt" Pack="true" PackagePath="" />
   </ItemGroup>
-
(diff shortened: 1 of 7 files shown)


tlmii (approved):

tlmii (commented):
Actually same question as in 6.0 branch... Since we dont have cpm here.

Directory.Build.props
```diff
@@ -25,4 +25,5 @@
   <ItemGroup>
     <None Include="$(RepoRoot)NOTICE.txt" Pack="true" PackagePath="" />
   </ItemGroup>
+
```
tlmii, line 28: nit: leftover whitespace from the old change

tlmii (approved):

PR #2993, opened 2024-10-03
add `identity` scaffolder to `dotnet-scaffold-aspnet`

deepchoudhery:
- added `identity` scaffolder to `dotnet-scaffold-aspnet`
  - templates added to `Templates\Identity` and also using existing `Templates\Files\ApplicationUser.tt`
  - rename `ValidateBlazorIdentityStep` to `ValidateIdentityStep`: validation step for both this scaffolder and the blazor-identity scaffolder.
  - added `IdentityHelper` for this scaffolder specific helpers.
  - code changes in `CodeModificationConfigs\identityChanges.json`

minor changes : 
- renamed `BlazorIdentityModel` to `IdentityModel` : both scaffolders need exactly the same properties.
- renamed `BlazorIdentitySettings` to `IdentitySettings` : same settings needed but added a `BlazorScenario` property. Allows for differentiation in some namespace and output path creation logic.
- minor changes to all `blazor-identity` templates due to the class name changes above, no logic/template code is affected.
- both identity scaffolders will live under the `Identity` category. (adding support for multiple categories later)


src/dotnet-scaffolding/dotnet-scaffold-aspnet/Helpers/BlazorIdentityHelper.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.DotNet.Tools.Scaffold.AspNet.Helpers;
 
 internal static class BlazorIdentityHelper
```
tlmii, line 11: Really feels like BlazorIdentityHelper and IdentityHelper should share a lot of code. They're so similar.

tlmii (approved):
Without focusing too much on the templates themselves, this looks good to me. Had two comments, neither of which have to be addressed in this PR but both of which would be good to see done (or some variant) eventually.

src/dotnet-scaffolding/dotnet-scaffold-aspnet/Extensions/IdentityScaffolderBuilderExtensions.cs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    public static IScaffoldBuilder WithIdentityCodeChangeStep(this IScaffoldBuilder builder)
+    {
+        builder = builder.WithStep<CodeModificationStep>(config =>
+        {
+            var step = config.Step;
+            var codeModificationFilePath = GlobalToolFileFinder.FindCodeModificationConfigFile("identityChanges.json", System.Reflection.Assembly.GetExecutingAssembly());
+            //get needed properties and cast them as needed
+            config.Context.Properties.TryGetValue(nameof(IdentitySettings), out var identitySettingsObj);
```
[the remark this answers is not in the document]
deepchoudhery: yep definitely need this, will add a TODO

src/dotnet-scaffolding/dotnet-scaffold-aspnet/Helpers/BlazorIdentityHelper.cs
deepchoudhery: only one method in common but this could be one file with both identity helper methods. will add a TODO

tlmii: It looked like maybe GetTextTemplatingProperties (same name in both), GetApplicationUserTextTemplatingProperty (same name in both), BlazorIdentityTemplateTypes/IdentityTemplateTypes could be shared implementation with just different inputs (for instance if this wasn't a static class, but was perhaps a base class with two derived classes and the derived classes provided the values used by the base class implementations of those methods). But I haven't gone line by line to see if they really are close enough to make that worthwhile - was just at a glance.

PR #2995, opened 2024-10-04
fix `DotnetCliRunner`

deepchoudhery:
When using `DotNetCliRunner.ExecuteWithCallbacks` or `DotNetCliRunner.ExecuteAndCaptureOutput` either in dotnet-scaffold or elsewhere, `Process.Start()` can throw on a myriad of reasons. 
- Found this when had another dotnet global tool installed `Sarif.PatternMatcher.Cli` and this method would throw and exit when trying to start the tool's process. 
- We should catch the exception and continue execution
- `dotnet-scaffold` would fail on getting all the commands due to this failure instead of returning all found commands.

### DIFF
--- a/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/CliHelpers/DotnetCliRunner.cs
+++ b/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/CliHelpers/DotnetCliRunner.cs
@@ -49,6 +49,8 @@ internal class DotnetCliRunner
         }
         catch (Exception e)
         {
+            stdOutCallback(e.Message);
+            stdErrCallback(e.Message);
             return -1;
         }
 
@@ -80,7 +82,7 @@ internal class DotnetCliRunner
         }
         catch (Exception e)
         {
-            stdOut = null;
+            stdOut = e.Message;
             stdErr = e.Message;
             return -1;
         }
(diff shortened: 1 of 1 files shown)


src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/CliHelpers/DotnetCliRunner.cs
```diff
@@ -66,7 +76,16 @@ public int ExecuteAndCaptureOutput(out string? stdOut, out string? stdErr)
...
         process.EnableRaisingEvents = true;
 
-        process.Start();
+        try
+        {
+            process.Start();
+        }
+        catch (Exception e)
+        {
+            stdOut = e.Message;
```
tlmii, line 85: Not really sure about writing it out to both... but definitely better than writing it out to neither! (same for the callback scenario)

deepchoudhery: I only had to stdErr earlier but figured if someone is debugging, they might only check one of those. Maybe just stdOut then?

tlmii: I think we probably just need to be consistent throughout the repo. If we're going to write errors to stderr, then we should do so. If we're going to write them to stdout in red, then we should do that. I tend to lean towards the latter because little of the ecosystem seems to use stderr, but I'm not particularly stuck on one or the other.

deepchoudhery: ok makes sense, lets keep it consistent with stdErr

tlmii (approved):

PR #2998, opened 2024-10-07
minor Blazor Identity fixes

deepchoudhery:
cherry-picking #2997
- some more additional fixes, mostly missing logic as the branches were inconsistent.

### DIFF
--- a/scripts/install-aspnet-codegenerator.cmd
+++ b/scripts/install-aspnet-codegenerator.cmd
@@ -1,5 +1,5 @@
 set VERSION=9.0.1-dev
-set DEFAULT_NUPKG_PATH=%userprofile%/.nuget/packages
+set DEFAULT_NUPKG_PATH=C:\Nuget
 set SRC_DIR=%cd%
 set NUPKG=artifacts\packages\Debug\Shipping\
 call taskkill /f /im dotnet.exe
--- a/scripts/install-aspnet-codegenerator.cmd
+++ b/scripts/install-aspnet-codegenerator.cmd
@@ -1,4 +1,4 @@
-set VERSION=9.0.0-dev
+set VERSION=9.0.1-dev
 set DEFAULT_NUPKG_PATH=%userprofile%/.nuget/packages
 set SRC_DIR=%cd%
 set NUPKG=artifacts\packages\Debug\Shipping\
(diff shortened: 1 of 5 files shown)


scripts/install-aspnet-codegenerator.cmd
```diff
@@ -1,5 +1,5 @@
-set VERSION=9.0.0-dev
-set DEFAULT_NUPKG_PATH=%userprofile%/.nuget/packages
+set VERSION=9.0.1-dev
+set DEFAULT_NUPKG_PATH=C:\Nuget
```
tlmii, line 2: This is probably not what we want, right? (I note it's what was already in the branch so it wasn't a change there)

scripts/install-aspnet-codegenerator.cmd
```diff
@@ -1,5 +1,5 @@
-set VERSION=9.0.0-dev
-set DEFAULT_NUPKG_PATH=%userprofile%/.nuget/packages
+set VERSION=9.0.1-dev
```
tlmii, line 1: Should this be bumped to 9.0.1-dev now if we haven't shipped 9.0.0 yet? Its just for testing I guess so doesn't matter much, just curious the intent.

scripts/install-aspnet-codegenerator.cmd
deepchoudhery: on my devbox, the `USERPROFILE` path does not work, I'll just add it to search for both paths. Or check if there is a default nuget path environment variable. Will fix this in the next PR

scripts/install-aspnet-codegenerator.cmd
deepchoudhery: yep left over from testing sorry

scripts/install-aspnet-codegenerator.cmd
tlmii: If you mean an actual devbox and not just your dev box, devbox overrides the nuget package location by setting the NUGET_PACKAGES environment variable, which overrides the default `%userprofile%\.nuget\packages` location (for me it points to `q:\.tools\.nuget\packages\`).

tlmii (approved):

PR #3019, opened 2024-10-15
setting up telemetry for `dotnet-scaffold`, `dotnet-scaffold-aspire`, and `dotnet-scaffold-aspnet`

deepchoudhery:
- added `ITelemetryService`
  - checks if telemetry is enabled
  - initializes the `TelemetryClient` and `TelemetryCommonProperties`
  - use `TelemetryClient.TrackEvent` to send events to app insights instance.
- added `IFirstTimeUseNoticeSentinel`
  - checks and adds (if needed) a tmp file to remember if telemetry disclosure text was already displayed to the user console.
- added `AddTelemetry` extension for adding telemetry to the first party tool service providers
- added `IEnvironmentService.GetMacAddress()` and `IEnvironmentService.IsDockerContainer`

Telemetry scenarios : 
`dotnet-scaffold`
1. tool is invoked --> check if sentinel file exists (no) --> sentinel file is created --> first time disclosure text is displayed (wait for user to continue) --> telemetry is disabled for this instance
2. tool is invoked --> check if sentinel file exists (yes) --> telemetry is enabled for this instance

`dotnet-scaffold-aspnet`/`dotnet-scaffold-aspire`
1. tool is invoked --> check if sentinel file exists (no) --> sentinel file is created --> first time disclosure text is displayed --> telemetry is disabled for this instance
2. tool is invoked --> check if sentinel file exists (yes) --> no disclosure text is displayed --> telemetry is enabled for this instance.
3. tool is invoked by `dotnet-scaffold` (first time) --> first time disclosure check is skipped --> telemetry is enabled for this instance


src/dotnet-scaffolding/dotnet-scaffold-aspire/Program.cs
```diff
@@ -24,6 +24,7 @@
        .WithOption(appHostProjectOption)
        .WithOption(projectOption)
        .WithOption(prereleaseOption)
+       .WithStep<FirstTimeTelemetrySetupStep>()
```
tlmii, line 27: I think I see a couple of problems with this approach, but let me know if I'm missing something or overthinking it.

One, having to add this to each scaffolder makes it kind of easy to miss one. If a new scaffolder is added, but the person writing it doesn't remember to add this step, then someone could run the tool without the first notice experience.

Two, I wonder about other scenarios where we need to run the first run experience. what about `dotnet-scaffold-aspire --version`? Or `dotnet-scaffold-aspire --help`? If we collect telemetry for them, we'd need to display the banner. Right now we don't, so we could avoid it I guess. And I note that `dotnet` itself doesn't display it for things like `dotnet --info` and `dotnet --version`. But I'm guessing that's because they don't collect telemetry for them. 

I wonder if it would be simple enough to call something down before the runner.RunAsync (after the Build, so that the services are created) so that we guarantee it always executes no matter what the call is?

src/dotnet-scaffolding/dotnet-scaffold-aspire/ScaffoldSteps/FirstTimeTelemetrySetupStep.cs
```diff
@@ -0,0 +1,44 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+using Microsoft.DotNet.Scaffolding.Core.Scaffolders;
+using Microsoft.DotNet.Scaffolding.Core.Steps;
+using Microsoft.DotNet.Scaffolding.Internal.Services;
+using Microsoft.DotNet.Scaffolding.Internal.Telemetry;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DotNet.Tools.Scaffold.Aspire.ScaffoldSteps;
+
+internal class FirstTimeTelemetrySetupStep : ScaffoldStep
```
tlmii, line 11: Could this (or whatever replaces it if we change it based on my above comment) be shared so you don't need it in each first party scaffolder?

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/Telemetry/TelemetryConstants.cs
```diff
@@ -0,0 +1,13 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.DotNet.Scaffolding.Internal.Telemetry;
+
+internal static class TelemetryConstants
+{
+    public static readonly string TELEMETRY_OPTOUT = "DOTNET_SCAFFOLD_TELEMETRY_OPTOUT";
```
tlmii, line 8: Do we need to discuss whether there should be a single opt out for all first party scaffolders and the main tool, or if you should be able to opt out individual tools individually? The first time use sentinel is done per-tool. But currently the opt out and the skip are done globally.

src/dotnet-scaffolding/dotnet-scaffold-aspire/ScaffoldSteps/ValidateOptionsStep.cs
```diff
@@ -19,7 +20,8 @@ public ValidateOptionsStep(ILogger<ValidateOptionsStep> logger)
 
     public override Task<bool> ExecuteAsync(ScaffolderContext context, CancellationToken cancellationToken = default)
     {
-        if (!ValidateMethod(context, _logger))
+        var validationResult = ValidateMethod(context, _logger);
+        if (!validationResult)
```
tlmii, line 24: Just for making debugging easier?

src/dotnet-scaffolding/dotnet-scaffold-aspnet/Program.cs
```diff
@@ -31,6 +31,7 @@ public static void Main(string[] args)
             .WithDescription("Add an empty razor component to a given project")
             .WithOption(projectOption)
             .WithOption(fileNameOption)
+            .WithStep<FirstTimeTelemetrySetupStep>()
```
tlmii, line 34: My comment above about needing to add this to all scaffolders in dotnet-scaffold-aspire seems even more relevant here with how many there are...

src/dotnet-scaffolding/dotnet-scaffold/Telemetry/CommandExecuteTelemetryEvent.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Microsoft.DotNet.Tools.Scaffold.Telemetry;
+
+internal class CommandExecuteTelemetryEvent : TelemetryEventBase
+{
+    private const string TelemetryEventName = "CommandExecute";
+    public CommandExecuteTelemetryEvent(DotNetToolInfo dotnetToolInfo, CommandInfo commandInfo, int? exitCode) : base(TelemetryEventName)
+    {
+        SetProperty("PackageName", dotnetToolInfo.PackageName);
+        SetProperty("Version", dotnetToolInfo.Version);
+        SetProperty("ToolName", dotnetToolInfo.Command);
```
tlmii, line 15: So many of these things (ToolName, CommandName, CommandCategory) become potentially PII when we allow extensibility. We'll probably need to talk to the privacy review folks about the right thing to do (always hash and have a well-known hash list? Flag first-party and only hash non-first party?).

src/dotnet-scaffolding/dotnet-scaffold-aspire/ScaffoldSteps/ValidateOptionsStep.cs
deepchoudhery: I was recording telemetry for this before so to make that easier, will just leave it like this since that's coming soon

scripts/install-scaffold.cmd
```diff
@@ -1,5 +1,5 @@
 set VERSION=9.0.0-dev
-set DEFAULT_NUPKG_PATH=%userprofile%/.nuget/packages
+set DEFAULT_NUPKG_PATH=C:\Nuget
```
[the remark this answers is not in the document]
deepchoudhery: yepyep, will revert this for now

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/CliHelpers/DotnetCliRunner.cs
```diff
@@ -117,5 +120,12 @@ private DotnetCliRunner(string commandName, IEnumerable<string> args)
         _psi.Environment.Remove("MSBuildSDKsPath");
         _psi.Environment.Remove("MSBuildExtensionsPath");
         _psi.Environment.Remove("MSBUILD_EXE_PATH");
+#if DOTNET_SCAFFOLD_INTERNAL
+        // Specify that this is being launched by dotnet-scaffold, so that we avoid
+        // the first run banner, etc.
+        // TODO: Allow extra Env Vars to be passed into the runner, so this can be done
```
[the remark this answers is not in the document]
deepchoudhery: yepyep sg

src/dotnet-scaffolding/dotnet-scaffold-aspire/Program.cs
deepchoudhery: yep I ran into the help scenario as well and thought it was fine to not show banner or collect telemetry then. I do prefer just doing it before RunAsync, and the adding the steps is cumbersome/easy to miss. Will switch to that, tyty

src/dotnet-scaffolding/dotnet-scaffold-aspnet/Program.cs
deepchoudhery: yep removing.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/Telemetry/TelemetryConstants.cs
deepchoudhery: single opt-out for both is fine I think.

src/dotnet-scaffolding/dotnet-scaffold-aspire/Program.cs
deepchoudhery: addressed

src/dotnet-scaffolding/dotnet-scaffold/Telemetry/CommandExecuteTelemetryEvent.cs
deepchoudhery: should I leave these in rn and we can address this in a privacy review?

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/Services/TelemetryService.cs
```diff
@@ -0,0 +1,149 @@
...
+    {
+        if (_trackEventTask is not null)
+        {
+            _trackEventTask = _trackEventTask.ContinueWith(
+                x => TrackEventTask(eventName, properties, measurements),
+                TaskScheduler.Default
+            );
+
+            //wait on the task to finish
+            Flush();
```
tlmii, line 55: We don't want to wait on the task here - that would negate using the task because it would block every call to TrackEvent. We would need to expose it via the interface and call it from somewhere after all the telemetry had been sent. In my testing for just dotnet-scaffold itself I wired it into ScaffoldCommand.ExecuteAsync before returning the result. But there is probably a better location (and possibly one that is more consistent between the tools since that won't exist in the other two).

src/dotnet-scaffolding/dotnet-scaffold/Telemetry/CommandExecuteTelemetryEvent.cs
tlmii: Leave it for now - let's try to talk to Phil tomorrow at least. I'd be hesitant to ship even a preview version wirhout the right answer because once the telemetry gets categorized on the backend it's very hard to change.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/Services/TelemetryService.cs
deepchoudhery: ah i see, will do

deepchoudhery: added `IFirstPartyToolTelemetryWrapper`, since we can't add this logic to the runner, the cleanest way I thought of for doing this work. Maybe not the best name, would like suggestions for that.

PR #3029, opened 2024-10-17
calling 'dotnet tool restore' before invoking local tools

deepchoudhery:
when bumping .NET SDK versions on a machine (.NET 8 --> 9 lets  say), local tools are not available to call until a 'dotnet tool restore' is called. Also a fairly mindful operation incase local tool files are messed with, this will help them be repaired before invocation.


tlmii (approved):
Curious about the perf impact of doing this every time. Wonder if there's some way we can detect the state. But for now it seems fine.

deepchoudhery:
> Curious about the perf impact of doing this every time. Wonder if there's some way we can detect the state. But for now it seems fine.

fair yeah, could be better to just restore our first party tools specifically (if that capability is there)

PR #3035, opened 2024-10-21
supporting scaffolders having multiple categories

deepchoudhery:
previously, scaffolders only were allowed 1 'Category'. Multiple are supported now : 
- changed `Microsoft.DotNet.Scaffolding.Core.Scaffolders.IScaffolder.Category` --> `Microsoft.DotNet.Scaffolding.Core.Scaffolders.IScaffolder.Categories`
- `IScaffoldBuilder.WithCategory` is adding to a `HashSet` now, by default including an `All` category.
- `Microsoft.DotNet.Scaffolding.Core.ComponentModel.DisplayCategory` --> `DisplayCategories` to account for the changes above
- `blazor-identity` now under multiple categories : `Blazor` and `Identity`
- changed telemetry tracking : 
  - previously, only tracking `CommandCategory` for the category for the picked scaffolder.
  - with above changes, tracking `ChosenCategory` for the interactively picked category, and `AllCommandCategories` for all applicable categories for the picked scaffolder. @sayedihashimi would like feedback on these names pls.

### DIFF
--- a/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Scaffolders/Scaffolder.cs
+++ b/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Scaffolders/Scaffolder.cs
@@ -16,7 +16,7 @@ public class Scaffolder : IScaffolder
     private readonly List<ScaffoldStepPreparer> _preparers;
     public string Name => _name;
     public string DisplayName => _displayName;
-    public List<string> Categories => _categories;
+    public IEnumerable<string> Categories => _categories;
     public string? Description => _description;
     public IEnumerable<ScaffolderOption> Options => _options;
 
--- a/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Scaffolders/Scaffolder.cs
+++ b/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Scaffolders/Scaffolder.cs
@@ -9,22 +9,22 @@ public class Scaffolder : IScaffolder
 {
     private readonly string _name;
     private readonly string _displayName;
-    private readonly string _category;
+    private readonly List<string> _categories;
     private readonly string? _description;
     private readonly List<ScaffolderOption> _options;
     private readonly List<ScaffoldStep> _steps;
     private readonly List<ScaffoldStepPreparer> _preparers;
     public string Name => _name;
     public string DisplayName => _displayName;
-    public string Category => _category;
+    public List<string> Categories => _categories;
     public string? Description => _description;
     public IEnumerable<ScaffolderOption> Options => _options;
 
-    internal Scaffolder(string name, string displayName, string category, string? description, List<ScaffolderOption> options, List<ScaffoldStep> steps, List<ScaffoldStepPreparer> preparers)
+    internal Scaffolder(string name, string displayName, List<string> categories, string? description, List<ScaffolderOption> options, List<ScaffoldStep> steps, List<ScaffoldStepPreparer> preparers)
     {
         _name = name;
         _displayName = displayName;
-        _category = category;
+        _categories = categories;
         _description = description;
         _options = options;
         _steps = steps;
--- a/src/dotnet-scaffolding/dotnet-scaffold/Flow/Steps/CategoryPickerFlowStep.cs
+++ b/src/dotnet-scaffolding/dotnet-scaffold/Flow/Steps/CategoryPickerFlowStep.cs
@@ -27,7 +27,8 @@ namespace Microsoft.DotNet.Tools.Scaffold.Flow.Steps
 
         public ValueTask ResetAsync(IFlowContext context, CancellationToken cancellationToken)
         {
-            context.Unset(FlowContextProperties.ScaffoldingCategory);
+            context.Unset(FlowContextProperties.ScaffoldingCategories);
+            context.Unset(FlowContextProperties.ChosenCategory);
             context.Unset(FlowContextProperties.ComponentName);
             context.Unset(FlowContextProperties.ComponentObj);
             context.Unset(FlowContextProperties.CommandName);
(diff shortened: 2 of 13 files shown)


src/dotnet-scaffolding/dotnet-scaffold/Flow/Steps/CategoryPickerFlowStep.cs
```diff
@@ -92,15 +93,24 @@ public ValueTask<FlowStepResult> ValidateUserInputAsync(IFlowContext context, Ca
...
 
-        private void SelectCategory(IFlowContext context, string categoryName)
+        private void SelectCategories(IFlowContext context, List<string> categories)
         {
             context.Set(new FlowProperty(
-                FlowContextProperties.ScaffoldingCategory,
-                categoryName,
+                FlowContextProperties.ScaffoldingCategories,
+                categories,
+                "Scaffolding Categories",
```
sayedihashimi, line 105: Where does this line show up in the UI? I'm wondering if we need 'Scaffolding' here. The user is already in `dotnet scaffold` so it may be redundant.

deepchoudhery: does not show up in the UI?

deepchoudhery: currently there is no display of the selected options.

sayedihashimi: ok thanks

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Scaffolders/Scaffolder.cs
```diff
@@ -9,22 +9,22 @@ public class Scaffolder : IScaffolder
...
-    private readonly string _category;
+    private readonly List<string> _categories;
     private readonly string? _description;
     private readonly List<ScaffolderOption> _options;
     private readonly List<ScaffoldStep> _steps;
     private readonly List<ScaffoldStepPreparer> _preparers;
     public string Name => _name;
     public string DisplayName => _displayName;
-    public string Category => _category;
+    public List<string> Categories => _categories;
```
tlmii, line 19: For a public API I would generally prefer this to be IEnumerable or ReadOnlyCollection or something like that rather than List (see [this](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections), though its been a while since I read it)

src/dotnet-scaffolding/dotnet-scaffold/Telemetry/CommandExecuteTelemetryEvent.cs
```diff
@@ -8,14 +9,19 @@ namespace Microsoft.DotNet.Tools.Scaffold.Telemetry;
...
-    public CommandExecuteTelemetryEvent(DotNetToolInfo dotnetToolInfo, CommandInfo commandInfo, int? exitCode) : base(TelemetryEventName)
+    public CommandExecuteTelemetryEvent(DotNetToolInfo dotnetToolInfo, CommandInfo commandInfo, int? exitCode, string? chosenCategory = null) : base(TelemetryEventName)
     {
         SetProperty("PackageName", dotnetToolInfo.PackageName);
         SetProperty("Version", dotnetToolInfo.Version);
         SetProperty("ToolName", dotnetToolInfo.Command);
         SetProperty("ToolLevel", dotnetToolInfo.IsGlobalTool ? TelemetryConstants.GlobalTool : TelemetryConstants.LocalTool);
         SetProperty("CommandName", commandInfo.Name);
-        SetProperty("CommandCategory", commandInfo.DisplayCategory);
+        SetProperty("AllCommandCategories", string.Join(",", commandInfo.DisplayCategories));
```
tlmii, line 19: There's a max length for telemetry properties. Last I checked it was 8192? Probably not an issue here... but something for us to watch out for as we concatenate things like this.

src/dotnet-scaffolding/dotnet-scaffold/Telemetry/CommandExecuteTelemetryEvent.cs
```diff
@@ -8,14 +9,19 @@ namespace Microsoft.DotNet.Tools.Scaffold.Telemetry;
...
+        if (!string.IsNullOrEmpty(chosenCategory))
+        {
+            SetProperty("ChosenCategory", chosenCategory);
```
tlmii, line 22: Just noting generally that from our discussion earlier we'll want to go back and make these two new properties hashed.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Scaffolders/Scaffolder.cs
deepchoudhery: yup no List<T> is public APIs, will switch.

src/dotnet-scaffolding/dotnet-scaffold/Telemetry/CommandExecuteTelemetryEvent.cs
deepchoudhery: good to know, for this option we should be plenty ok. thats hundreds of categories (including whitespace, and commas)

src/dotnet-scaffolding/dotnet-scaffold/Telemetry/CommandExecuteTelemetryEvent.cs
deepchoudhery: Is [this](https://github.com/dotnet/Scaffolding/blob/b4afd5d143d15dfd6dbb1a8efd0465463e9b1d5c/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/Extensions/StringExtensions.cs#L152) enough for hashing?

tlmii: Yes, that should be good as far as I'm aware

tlmii (approved):

PR #3043, opened 2024-10-25
add telemetry events for first party `dotnet-scaffold-*` tools.

deepchoudhery:
Mostly want to track which steps are used and their success rate. Since I did not want to keep all telemetry code and references in this project, wrapped all the used steps as in to add the telemetry service and track the result of said steps.
`dotnet-scaffold-aspire` : 
- added `WrappedAddPackageStep`,
- simply just get the result of the base step and track it
- using Wrapped steps in all the `BuilderExtensions`

`dotnet-scaffold-aspnet` : 
- added `WrappedAddPackageStep`, `WrappedCodeModificationStep`, and `WrappedTextTemplatingStep`.
- using Wrapped steps in all the `BuilderExtensions`
- tracking all the `Validate` step telemetry using the same telemetry event, `ValidateScaffolderTelemetryEvent`. Another simple event tracking of success/failure.

### DIFF
--- a/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/Services/TelemetryService.cs
+++ b/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/Services/TelemetryService.cs
@@ -6,6 +6,7 @@ using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.Channel;
 using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.DotNet.PlatformAbstractions;
+using Microsoft.DotNet.Scaffolding.Internal.Extensions;
 using Microsoft.DotNet.Scaffolding.Internal.Telemetry;
 using Microsoft.Extensions.Logging;
 
@@ -128,8 +129,16 @@ internal class TelemetryService : ITelemetryService, IDisposable
             var eventProperties = new Dictionary<string, string>(_commonProperties);
             foreach (KeyValuePair<string, string> property in properties)
             {
-                eventProperties[property.Key] = property.Value;
+                string eventName = property.Key;
+                bool pii = eventName.StartsWith(TelemetryConstants.PII);
+                if (pii)
+                {
+                    eventName = eventName.Substring(TelemetryConstants.PII.Length);
+                }
+
+                eventProperties[eventName] = pii ? property.Value.ToString()!.Hash() : property.Value.ToString()!;
             }
+
             return eventProperties;
         }
         else
--- a/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/Services/TelemetryService.cs
+++ b/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/Services/TelemetryService.cs
@@ -136,7 +136,7 @@ internal class TelemetryService : ITelemetryService, IDisposable
                     eventName = eventName.Substring(TelemetryConstants.PII.Length);
                 }
 
-                eventProperties[eventName] = pii ? property.Value.ToString()!.Hash() : property.Value.ToString()!;
+                eventProperties[eventName] = pii ? property.Value.Hash() : property.Value.ToString()!;
             }
 
             return eventProperties;
--- a/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/Telemetry/TelemetryEventBase.cs
+++ b/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/Telemetry/TelemetryEventBase.cs
@@ -1,19 +1,51 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using Microsoft.DotNet.Scaffolding.Internal.Extensions;
+
 namespace Microsoft.DotNet.Scaffolding.Internal.Telemetry;
 
 internal abstract class TelemetryEventBase(string name)
 {
     private readonly Dictionary<string, string> _properties = [];
     private readonly Dictionary<string, double> _measurements = [];
-
+    
     public string Name => name;
     public IReadOnlyDictionary<string, string> Properties => _properties;
     public IReadOnlyDictionary<string, double> Measurements => _measurements;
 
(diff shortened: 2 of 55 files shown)


src/dotnet-scaffolding/dotnet-scaffold-aspnet/ScaffoldSteps/ValidateRazorPagesStep.cs
```diff
@@ -40,6 +45,7 @@ public override async Task<bool> ExecuteAsync(ScaffolderContext context, Cancell
         var codeModifierProperties = new Dictionary<string, string>();
         if (razorPagesSettings is null)
         {
+            _telemetryService.TrackEvent(new ValidateScaffolderTelemetryEvent(nameof(ValidateIdentityStep), context.Scaffolder.DisplayName, result: false));
```
tlmii, line 48: Should all of these (other files too) be ValidateIdentityStep? Shouldn't it be ValidateRazorPagesStep in this file, etc?

src/dotnet-scaffolding/dotnet-scaffold/Telemetry/CommandExecuteTelemetryEvent.cs
```diff
@@ -15,7 +15,7 @@ public CommandExecuteTelemetryEvent(DotNetToolInfo dotnetToolInfo, CommandInfo c
         SetProperty("Version", dotnetToolInfo.Version);
         SetProperty("ToolName", dotnetToolInfo.Command);
         SetProperty("ToolLevel", dotnetToolInfo.IsGlobalTool ? TelemetryConstants.GlobalTool : TelemetryConstants.LocalTool);
-        SetProperty("CommandName", commandInfo.Name);
+        SetProperty("CommandName", commandInfo.Name.Hash());
         SetProperty("AllCommandCategories", string.Join(",", commandInfo.DisplayCategories).Hash());
```
tlmii, line 19: Not new here, but we probably want to Hash each category, not the comma-delimeted list. Otherwise our well-known values list becomes really complicated.

src/dotnet-scaffolding/dotnet-scaffold/Telemetry/CommandExecuteTelemetryEvent.cs
```diff
@@ -15,7 +15,7 @@ public CommandExecuteTelemetryEvent(DotNetToolInfo dotnetToolInfo, CommandInfo c
         SetProperty("Version", dotnetToolInfo.Version);
         SetProperty("ToolName", dotnetToolInfo.Command);
         SetProperty("ToolLevel", dotnetToolInfo.IsGlobalTool ? TelemetryConstants.GlobalTool : TelemetryConstants.LocalTool);
-        SetProperty("CommandName", commandInfo.Name);
+        SetProperty("CommandName", commandInfo.Name.Hash());
```
tlmii, line 18: I wonder if we should do something like UA has where we have an overload to SetProperty that takes bool isPii (or make it have a default false and no overload I guess) and it handles the hashing under the hood. So we don't need all the events to know about how we do the hashing, just in case it needs to change. That of course complicates my other comment about AllCommandCategories, but maybe that's the rare case where we can't (or we support a SetProperty overload that takes an IEnumerable<string>).

src/dotnet-scaffolding/dotnet-scaffold-aspnet/ScaffoldSteps/ValidateRazorPagesStep.cs
deepchoudhery: oops yep will fix these

src/dotnet-scaffolding/dotnet-scaffold/Telemetry/CommandExecuteTelemetryEvent.cs
deepchoudhery: I like this idea, will take a quick peek at UA and having the overload for IEnumerable sounds good.

src/dotnet-scaffolding/dotnet-scaffold/Telemetry/CommandExecuteTelemetryEvent.cs
deepchoudhery: yep adding an extension for that, will work differently than the isPII scenario, verified

src/dotnet-scaffolding/dotnet-scaffold/Telemetry/CommandExecuteTelemetryEvent.cs
deepchoudhery: done, verified

tlmii (approved):
One nit and one tweak, but otherwise approved

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/Services/TelemetryService.cs
```diff
@@ -128,8 +129,16 @@ private Dictionary<string, string> GetEventProperties(IReadOnlyDictionary<string
...
             {
-                eventProperties[property.Key] = property.Value;
+                string eventName = property.Key;
+                bool pii = eventName.StartsWith(TelemetryConstants.PII);
+                if (pii)
+                {
+                    eventName = eventName.Substring(TelemetryConstants.PII.Length);
+                }
+
+                eventProperties[eventName] = pii ? property.Value.ToString()!.Hash() : property.Value.ToString()!;
```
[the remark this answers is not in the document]
deepchoudhery: oops yes, was trying to get measurements as well and decided against it, leftover from that.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/Telemetry/TelemetryEventBase.cs
```diff
@@ -1,19 +1,51 @@
...
+using Microsoft.DotNet.Scaffolding.Internal.Extensions;
+
 namespace Microsoft.DotNet.Scaffolding.Internal.Telemetry;
 
 internal abstract class TelemetryEventBase(string name)
 {
     private readonly Dictionary<string, string> _properties = [];
     private readonly Dictionary<string, double> _measurements = [];
-
+
```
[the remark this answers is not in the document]
deepchoudhery: fixed ty

PR #3060, opened 2024-10-30
check dotnet-scaffold version, account for prerelease

deepchoudhery:
if prerelease, install prerelease first party-packages. If GA/release packages, install GA/release first party packages.
- checking for prerelease using the keywords often used for preview .NET packages.

### DIFF
--- a/src/dotnet-scaffolding/dotnet-scaffold/AppBuilder/ScaffoldCommandAppBuilder.cs
+++ b/src/dotnet-scaffolding/dotnet-scaffold/AppBuilder/ScaffoldCommandAppBuilder.cs
@@ -1,8 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
-using System.Reflection;
 using Microsoft.DotNet.Scaffolding.Internal.Services;
 using Microsoft.DotNet.Tools.Scaffold.AppBuilder;
 using Microsoft.DotNet.Tools.Scaffold.Command;
+using Microsoft.DotNet.Tools.Scaffold.Helpers;
 using Microsoft.DotNet.Tools.Scaffold.Services;
 using Microsoft.Extensions.DependencyInjection;
 using Spectre.Console.Cli;
@@ -13,7 +13,7 @@ internal class ScaffoldCommandAppBuilder(string[] args)
 {
     private readonly string[] _args = args;
     //try to update this every release
-    private readonly string _backupDotNetScaffoldVersion = "9.0.0-dev";
+    private readonly string _backupDotNetScaffoldVersion = "9.0.0";
 
     public ScaffoldCommandApp Build()
     {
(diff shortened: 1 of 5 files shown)


src/dotnet-scaffolding/dotnet-scaffold/Helpers/ToolHelper.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using System.Diagnostics;
+using System.Reflection;
+
+namespace Microsoft.DotNet.Tools.Scaffold.Helpers;
+
+internal static class ToolHelper
+{
+    // Define keywords for prerelease versions
+    internal static string[] PrereleaseKeywords = { "preview", "rc", "rtm", "alpha", "beta", "dev" };
```
tlmii, line 12: _Probably_ not worth bringing in NuGet.Versioning package just for this, but if we did (and I don't think it has any transitive dependencies, so not that big of a deal), SemanticVersion class has a [IsPrerelease](https://github.com/NuGet/NuGet.Client/blob/dddfdd8617e80ec5a2ea0dd94a094b243f355d09/src/NuGet.Core/NuGet.Versioning/SemanticVersion.cs#L213) property.

src/dotnet-scaffolding/dotnet-scaffold/AppBuilder/ScaffoldCommandAppBuilder.cs
```diff
@@ -23,7 +23,7 @@ public ScaffoldCommandApp Build()
         {
             config
                 .SetApplicationName("dotnet-scaffold")
-                .SetApplicationVersion(GetToolVersion())
+                .SetApplicationVersion(ToolHelper.GetToolVersion() ?? _backupDotNetScaffoldVersion)
```
tlmii, line 26: Since we theoretically control whether the informational version is present in the assembly or not, what's the purpose of the backup version?

tlmii (approved):

src/dotnet-scaffolding/dotnet-scaffold/Helpers/ToolHelper.cs
```diff
@@ -0,0 +1,31 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics;
```
tlmii, line 4: I think this using is unnecessary

src/dotnet-scaffolding/dotnet-scaffold/Helpers/ToolHelper.cs
deepchoudhery: not worth it IMO, have used it in the past and it can bug out.

src/dotnet-scaffolding/dotnet-scaffold/AppBuilder/ScaffoldCommandAppBuilder.cs
deepchoudhery: The calls to get the assembly informational version is nullable (so is ToolHelper.GetToolVersion() consequently) so its a backup to that. It's something to update every release (slightly annoying) but I can figure out a way to get it from Versions.props if that helps?

PR #3087, opened 2024-11-20
initial logging changes

deepchoudhery:
draft PR for now, want to get some feedback before adding support for trace logging, Serilog, file logging etc.


eng/Versions.props
```diff
@@ -11,49 +11,51 @@
...
-    <MicrosoftEntityFrameworkCoreDesignPackageVersion>9.0.0-rc.2.24474.1</MicrosoftEntityFrameworkCoreDesignPackageVersion>
+    <MicrosoftEntityFrameworkCoreDesignPackageVersion>9.0.0</MicrosoftEntityFrameworkCoreDesignPackageVersion>
     <!-- Microsoft.EntityFrameworkCore -->
-    <MicrosoftEntityFrameworkCorePackageVersion>9.0.0-rc.2.24474.1</MicrosoftEntityFrameworkCorePackageVersion>
+    <MicrosoftEntityFrameworkCorePackageVersion>9.0.0</MicrosoftEntityFrameworkCorePackageVersion>
     <!-- Microsoft.EntityFrameworkCore.SqlServer -->
-    <MicrosoftEntityFrameworkCoreSqlServerPackageVersion>9.0.0-rc.2.24474.1</MicrosoftEntityFrameworkCoreSqlServerPackageVersion>
+    <MicrosoftEntityFrameworkCoreSqlServerPackageVersion>9.0.0</MicrosoftEntityFrameworkCoreSqlServerPackageVersion>
     <!-- Microsoft.Extensions.DependencyInjection -->
-    <MicrosoftExtensionsDependencyInjectionPackageVersion>9.0.0-rc.2.24473.5</MicrosoftExtensionsDependencyInjectionPackageVersion>
```
tlmii, line 20: We should split out all these version updates to a PR of its own

tlmii:
@deepchoudhery Could you add some examples of how the logging options would be used by the end user in the PR description? They seem pretty straightforward, but I want to make sure.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunnerBuilder.cs
```diff
@@ -72,13 +74,41 @@ private void AddDefaultServices()
...
+            loggerConfig.MinimumLevel.Verbose();
+        }
+
+        if (isLogToFileEnabled)
+        {
+            var currentDirectory = Directory.GetCurrentDirectory();
+            if (!string.IsNullOrEmpty(currentDirectory))
+            {
+                var logPath = Path.Combine(currentDirectory, _defaultLogFolder, "dotnet-scaffold.txt");
+                loggerConfig.WriteTo.File(logPath, rollingInterval: RollingInterval.Minute);
```
tlmii, line 94: Does RollingInterval.Minute mean you get a new log file every minute? If so, is that by wall clock or runtime? Like if you start running the scaffolder at 1:00:55 and it runs for 10 seconds to 1:01:05, do you end up with two log files for the run? Or just one?

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Logging/AnsiConsoleSink.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Microsoft.DotNet.Scaffolding.Core.Logging;
+
+public class AnsiConsoleSink : ILogEventSink
+{
+    public void Emit(LogEvent logEvent)
+    {
+        Debugger.Launch();
+        //should be "enabled", "disabled", "", or null. Empty or null will indicate that a tool is not launched by dotnet-scaffold.
+        bool isRedirectedToDotnetScaffold = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("DOTNET_SCAFFOLD_TELEMETRY_STATE"));
```
tlmii, line 16: Tying this to telemetry doesn't feel great. I get the logic of it. But I sort of wonder if we shouldn't add something more explicit to say that we were launched from dotnet-scaffold. I'm not sure what exactly... Another environment variable?

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunnerBuilder.cs
deepchoudhery: testing all these scenarios out today, will follow up soon.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Logging/AnsiConsoleSink.cs
deepchoudhery: yep another environment variable should be ok here.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunnerBuilder.cs
deepchoudhery: yes, so we get 2 log files on the minute switch, and multiple executions do not create duplicate files, it adds to the same file. adding more info in the description. I think I prefer logs every minute vs hourly. Any thoughts on this?

tlmii: I think the most common scenario we're going to want these log files from is when we ask customers "send us the logs from your most recent run". From my perspective, it would be nice if that was one file that contained only the logs from that current run (or, alternatively, one log file per scaffolder or similar distinction). So we don't have to ask if it was the run at 4am or the run at 4:30am (or 4:00:00am vs 4:00:30am for minute rolling), and we don't have to ask to make sure they send all the files because it might have gotten split across multiple and then we're left piecing them back together.

I see the rolling interval from serilog being more interesting for log files that are being regularly generated - a web app that's running all the time and it dumps logs every few seconds all day or whatever. The scaffolder runs are discrete runs that shouldn't last more than a couple minutes at most, though it'd be interesting to have all of those minutes together in one file.

I wonder if we should consider a different approach of just picking a filename based on the execution time of the tool and not rolling it. Whatever goes into that log file goes into it. So we get PATH\dotnet-scaffold_20240612125152.log for that entire run. 

Thoughts?

deepchoudhery: yeah that makes a lot of sense, one file per execution/run. also do we prefer .log extension over .txt?

deepchoudhery: nvm, .NET seems to use .log convention so will do that

deepchoudhery: addressed, updating the PR description

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Roslyn/Services/MsBuildInitializer.cs
```diff
@@ -45,21 +47,27 @@ private void RegisterMsbuild()
 
             if (!sdkPaths.Any())
             {
-                _logger.LogInformation($"Could not find a .NET SDK at the default locations.");
+                _logger.LogDebug($"Could not find a .NET SDK at the default locations.");
```
tlmii, line 50: Should we make this message and the one above it different so we know which check failed based on the message?

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunnerBuilder.cs
```diff
@@ -72,13 +75,42 @@ private void AddDefaultServices()
 
     private void AddCoreServices()
     {
-        Services.AddLogging();
-        Logging.AddCleanConsoleFormatter();
+        //get some logging properties from environment variables
+        var isVerboseEnabled = Environment.GetEnvironmentVariable(ScaffolderConstants.ENABLE_VERBOSE_LOGGING)?.Equals("true", StringComparison.OrdinalIgnoreCase) == true;
```
tlmii, line 79: For clarity, I'd split all this logging configuration stuff out into a separate method that returns the LoggerConfiguration object.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunnerBuilder.cs
```diff
@@ -72,13 +75,42 @@ private void AddDefaultServices()
...
+            var currentDirectory = Directory.GetCurrentDirectory();
+            if (!string.IsNullOrEmpty(currentDirectory))
+            {
+                var filePath = $"dotnet-scaffold-{DateTime.UtcNow:yyyy-MM-dd_HH-mm}.log";
+                var logPath = StringUtil.GetUniqueFilePath(Path.Combine(currentDirectory, _defaultLogFolder, filePath));                
+                loggerConfig.WriteTo.File(logPath, rollingInterval: RollingInterval.Infinite);
+            }
+        }
+
+        Log.Logger = loggerConfig.CreateLogger();
```
tlmii, line 100: Out of curiosity, did you try the configuration constructs mentioned here? https://github.com/serilog/serilog-extensions-hosting ? I ask just because I like the inline variant of it that allows for not touching the static `Log.Logger`. Though i'm not entirely sure the end result is the same - its technically a different package (`Serilog.Extensions.Hosting` vs `Serilog.Extensions.Logging`). Though the note on the [Logging readme](https://github.com/serilog/serilog-extensions-logging?tab=readme-ov-file) says 

> Non-web .NET Core Instructions
Non-web .NET Core applications should prefer [Serilog.Extensions.Hosting](https://github.com/serilog/serilog-extensions-hosting) and UseSerilog() instead.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunnerBuilder.cs
tlmii, line 79: Since we do this `[string]?.Equals("true", StringComparison.OrdinalIgnoreCase) == true` check at least three times in this PR (maybe other places too?) maybe we should create another extension method like the `GetEnvironmentVariableAsBool` one we have? Looks like maybe none of these are using IEnvironmentService - if they can't be changed, I guess they could be extension methods on strings. Or maybe just a helper function so as not to pollute the string intellisense...

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunnerBuilder.cs
```diff
@@ -72,13 +75,42 @@ private void AddDefaultServices()
...
+        {
+            loggerConfig.MinimumLevel.Verbose();
+        }
+
+        if (isLogToFileEnabled)
+        {
+            var currentDirectory = Directory.GetCurrentDirectory();
+            if (!string.IsNullOrEmpty(currentDirectory))
+            {
+                var filePath = $"dotnet-scaffold-{DateTime.UtcNow:yyyy-MM-dd_HH-mm}.log";
```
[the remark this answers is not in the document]
deepchoudhery: ok sounds good, will switch.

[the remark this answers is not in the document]
sayedihashimi: Agreed that it would be better to have a single folder to view all log files.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Logging/AnsiConsoleSink.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+    }
+
+    private string FormatMessage(LogEventLevel level, string message)
+    {
+        return level switch
+        {
+            LogEventLevel.Verbose => $"[gray]{message}[/]",
+            LogEventLevel.Debug => $"[gray]{message}[/]",
+            LogEventLevel.Information => $"[green]{message}[/]",
```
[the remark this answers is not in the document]
sayedihashimi: Agreed, we should only colorize for specific reasons.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunnerBuilder.cs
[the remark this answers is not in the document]
deepchoudhery: maybe in the .dotnet-scaffold folder then? `%USERPROFILE%\.dotnet-scaffold\.logs\`?

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Logging/AnsiConsoleSink.cs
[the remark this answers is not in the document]
deepchoudhery: what color do we want the normal messages? plain white?

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunnerBuilder.cs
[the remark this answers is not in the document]
tlmii: That's probably fine. %TEMP% is often used to make sure it can get cleared out automatically at some point. But since we don't do any automatic logging (its all opt-in) its probably not necessary.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Microsoft.DotNet.Scaffolding.Core.csproj
```diff
@@ -1,13 +1,18 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFrameworks>$(BuildTfm)</TargetFrameworks>
     <PackageTags>dotnet;scaffold;</PackageTags>
     <PackageReadmeFile>README.md</PackageReadmeFile>
+    <NoWarn>$(NoWarn);8002</NoWarn>
```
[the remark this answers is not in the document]
deepchoudhery: used to be for packing unsigned dlls (Spectre.Console, Serilog in this case) but its actually needed anymore with .NET 9 good catch tyty, removing.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Logging/AnsiConsoleSink.cs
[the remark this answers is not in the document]
tlmii: More specifically I think it should just be the default color? I don't know what specifying white would do if they have a terminal with a white background. But yeah, that's the way msbuild output is, for instance. There's callouts of green and red for success and failure. Maybe they do the gray for verbose like you have here too? Can't remember. But overall its just white.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunnerBuilder.cs
deepchoudhery: yep an extension method would do, i'll try to see if adding IEnvironmentService makes sense first.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunnerBuilder.cs
deepchoudhery: oh yeah i missed this I think. I'll play with it

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunnerBuilder.cs
deepchoudhery: sounds good

deepchoudhery: using a helper, fixed

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Logging/AnsiConsoleSink.cs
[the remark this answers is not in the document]
deepchoudhery: there is a default color (or leaving it blank) and it will the default terminal color, doing that.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Roslyn/Services/MsBuildInitializer.cs
deepchoudhery: yes will do, makes sense.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunnerBuilder.cs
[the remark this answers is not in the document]
deepchoudhery: logging to %USERPROFILE%\.dotnet-scaffold\.logs\ for now then, fixed.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunnerBuilder.cs
deepchoudhery: actually didn't need the other package at all, builder.AddSerilog has an overload to add the created LoggerConfiguration, looks like `builder.AddSerilog(loggerConfig.CreateLogger(), dispose: true);` now. Tested and everything is good.

tlmii (approved):

PR #3092, opened 2024-11-26
Update transitive dependencies

tlmii:
Updating a few transitive dependencies to fix some alerts. The 9.0 versions of these packages still ship net8.0 binaries so we should be ok to update to them.


deepchoudhery (approved):
sounds good

PR #3113, opened 2024-12-10
fix AccountLayout, also MSBuildProjectService for blazor projects.

deepchoudhery:
AccountLayout had an incorrect `@layout` reference. Fixing that reference.
- adding a try catch to MSBuildProjectService.Initialize. Loading MSBuild projects fails for blazor projects with no real solution afaik so will dig into this later.  The MSBuildProjectService is nullable friendly so no issues.
- fixed `@layout` reference to `IdentityModel.IdentityLayoutNamespace` and initializing this property correctly.
- updated version from 9.0.0 to 9.0.1.

### DIFF
--- a/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Roslyn/Services/MSBuildProjectService.cs
+++ b/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Roslyn/Services/MSBuildProjectService.cs
@@ -9,7 +9,7 @@ namespace Microsoft.DotNet.Scaffolding.Roslyn.Services;
 public class MSBuildProjectService : IMSBuildProjectService
 {
     private readonly string _projectPath;
-    private Build.Evaluation.Project? _project;
+    private Project? _project;
     private bool _initialized;
     private readonly object _initLock = new();
     public MSBuildProjectService(string projectPath)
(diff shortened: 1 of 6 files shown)


src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Roslyn/Services/MSBuildProjectService.cs
```diff
@@ -53,13 +53,12 @@ private void Initialize(bool refresh = false)
                 return;
             }
 
-            //try loading MSBuild project the faster way.
-            _project = new Project(_projectPath, null, null, new Build.Evaluation.ProjectCollection(), ProjectLoadSettings.IgnoreMissingImports);
-            //if fails, use the longer way using GLobalProjectCollection.
-            if (_project is null)
+            try
             {
-                _project = ProjectCollection.GlobalProjectCollection.LoadProject(_projectPath);
```
tlmii, line 61: This backup is no longer needed?

deepchoudhery: yep this needs to be reworked. The backup wasn't particularly working so need a better way to go about this.

PR #3118, opened 2024-12-13
fix StringUtil.ToPath and some null checks

deepchoudhery:
Blazor Identity and CRUD scenario was broken in projects where project names had '.' characters in them. Due to faulty StringUtil.ToPath method
- fixed the method to account for such project names (both dotnet-aspnet-codegenerator and new dotnet-scaffold)
- added a test case, fixed one as well
- added a missing null check in ProjectExtensions.GetDocumentFromName

### DIFF
--- a/src/Shared/Microsoft.DotNet.Scaffolding.Shared/General/StringUtil.cs
+++ b/src/Shared/Microsoft.DotNet.Scaffolding.Shared/General/StringUtil.cs
@@ -60,7 +60,7 @@ namespace Microsoft.DotNet.Scaffolding.Shared
                 }
 
                 // Remove redundant project root namespace folder from basePath if present
-                if (basePath.EndsWith(Path.Combine(projectRootNamespace, "") + Path.DirectorySeparatorChar, StringComparison.Ordinal))
+                if (basePath.EndsWith(Path.Combine(projectRootNamespace, string.Empty) + Path.DirectorySeparatorChar, StringComparison.Ordinal))
                 {
                     basePath = basePath.Substring(0, basePath.Length - projectRootNamespace.Length - 1);
                 }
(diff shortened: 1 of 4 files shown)


src/Shared/Microsoft.DotNet.Scaffolding.Shared/General/StringUtil.cs
```diff
@@ -40,23 +40,49 @@ public static bool ContainsIgnoreCase(this string input, string value)
...
+                    basePath = Path.GetDirectoryName(basePath);
                 }
-                //invalid path
-                catch (Exception ex) when (ex is ArgumentException || ex is PathTooLongException || ex is NotSupportedException)
-                {}
-            }
 
-            return path;
+                // Ensure basePath ends with a directory separator
+                if (!basePath.EndsWith(Path.DirectorySeparatorChar.ToString(), StringComparison.Ordinal))
```
tlmii, line 57: Not saying you should do it now, but all three of these kinds of checks (the three if blocks starting here) all seem like things that could be pretty common for scaffolders to need to do that could make for useful helper methods. Obviously not specific to the namespace and periods and whatnot, but like EnsureEndingDirectorySeparator(), TryRemoveFromEnd(), TryFromFromStart() or whatever (not suggesting those names, just giving those to signal intent)

tlmii: (this comment really more applies to the one in dotnet-scaffold, didn't realize which one I was looking at at the time)

tlmii (approved):

src/Shared/Microsoft.DotNet.Scaffolding.Shared/General/StringUtil.cs
deepchoudhery: ye those will be handy, will add a TODO

PR #3162, opened 2025-06-17
Blazor CRUD uses `NotFound()` method instead of navigation

### LINKED ISSUE #3168: Update the Blazor CRUD scaffolder for .NET 10 to use NavigationManager.NotFound()
In .NET 10 we're adding proper support to Blazor for generating Not Found responses. We should update the Blazor CRUD scaffolder to call `NavigationManager.NotFound()` instead of the current hack of redirecting to a nonexistent page using `NavigationManager.NavigateTo("notfound")`.

ilonatommy:
This PR is updating Blazor CRUD scaffolder to use `NavigationManager.NotFound()` instead of `NavigationManager.NavigateTo("notfound")`. I am not sure how to apply all the necessary changes. I will list all the actions needed and would be grateful for help in completing it, cc @deepchoudhery, @tlmii.

Changes:
- [x] Replace`NavigateTo(\"notfound\")` with `NavigationManager.NotFound()` Blazor CRUD.
- [x] Add `NotFound.razor` page to the project that could be same as https://github.com/dotnet/aspnetcore/blob/9a6b2eb7ee0eaffcdd3a361b06899b4fe735a4d6/src/ProjectTemplates/Web.ProjectTemplates/content/BlazorWeb-CSharp/BlazorWeb-CSharp/Components/Pages/NotFound.razor
- [x] Add `NotFoundPage` parameter to `Router`, like here:
https://github.com/dotnet/aspnetcore/blob/9a6b2eb7ee0eaffcdd3a361b06899b4fe735a4d6/src/ProjectTemplates/Web.ProjectTemplates/content/BlazorWeb-CSharp/BlazorWeb-CSharp/Components/Routes.razor#L7
- [x] Add re-execution middleware with the route to NotFound.razor page in Program.cs, like here: https://github.com/dotnet/aspnetcore/blob/a54b996f94a70bb0295fff8569b0f09e9f27b8bb/src/ProjectTemplates/Web.ProjectTemplates/content/BlazorWeb-CSharp/BlazorWebCSharp.1/Program.cs#L103.
- [x] Test the result:
`dotnet new blazor` [server.diff.txt](https://github.com/user-attachments/files/23626712/server.diff.txt)
`dotnet new blazor -int WebAssembly` [wasm.diff.txt](https://github.com/user-attachments/files/23626686/wasm.diff.txt)

Fixes #3168

### DIFF
--- a/src/dotnet-scaffolding/dotnet-scaffold/AspNet/CodeModificationConfigs/blazorWebCrudChanges.json
+++ b/src/dotnet-scaffolding/dotnet-scaffold/AspNet/CodeModificationConfigs/blazorWebCrudChanges.json
@@ -67,18 +67,8 @@
           "CheckBlock": "NotFoundPage=\"typeof(Pages.NotFound)\""
         },
         {
-          "ReplaceSnippet": [ "<Router AppAssembly=\"typeof(Program).Assembly\" AdditionalAssemblies=\"new[] { typeof($(AutoGenProjectName).Client._Imports).Assembly }\">" ],
-          "Block": "<Router AppAssembly=\"typeof(Program).Assembly\" AdditionalAssemblies=\"new[] { typeof($(AutoGenProjectName).Client._Imports).Assembly }\" NotFoundPage=\"typeof(Pages.NotFound)\">",
-          "CheckBlock": "NotFoundPage=\"typeof(Pages.NotFound)\""
-        }
-      ]
-    },
-    {
-      "FileName": "App.razor",
-      "Replacements": [
-        {
-          "ReplaceSnippet": [ "<Router AppAssembly=\"@typeof(App).Assembly\">" ],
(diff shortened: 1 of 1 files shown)


danroth27:
Note that these changes are specific to .NET 10. For .NET 9 and earlier we'll need to continue to generate the existing code for handling not found responses.

deepchoudhery (approved):
LGTM

danroth27:
@javiercn It looks like you had to abandon #3178, so I assume this PR is still where we should focus our attention, correct?

danroth27:
Or just test on a Blazor Web App originally created with .NET 9 and then retargeted to .NET 10.

danroth27:
Note that we also need to version these scaffolding updates so that the .NET 9 version of the scaffolder is preserved. This is different from the old scaffolding system where each scaffolder version shipped as a separate package from a separate branch. With `dotnet scaffold` it seems that all supported versions of the scaffolder need to ship together so that you can scaffold for both .NET 9 & 10 based projects.

phenning (approved):

PR #3186, opened 2025-09-04
Just use simple projects in MSBuildProjectServiceHelperTests

phenning:
- we should just use simple projects without SDK reference in order to make tests more reliable.
- this avoids having a dependency on the SDK for tests which do not need anything from the SDK

We are making this change due to the tests failing from breaking SDK changes which would impact the virtual project here. If we wanted a complete end to end, we could use the SDK locator to properly load the SDK, or inject the imports with the complete paths into the csprojectContent


phenning:
This will allow us to unblock the arcade update.

PR #3236, opened 2025-09-17
dev 10 is rc 1

haileymck:
keep dev10 rc1 until .net 10 is rc 2


haileymck:
update I found the rc1

Issue #3242, opened 2025-09-23
Improve dotnet-scaffold non-interactive experience


haileymck:
Created to group and track the work around improving the dotnet-scaffold non-interactive experience.

PR #3247, opened 2025-09-23
Run Spectre.Console experience not using --non-interactive

### LINKED ISSUE #3243: noninteractive dotnet-scaffold tool should run with System.CommandLine

### LINKED ISSUE #3244: interactive dotnet-scaffold should run with Spectre.Console

haileymck:
As part of the greater work in #3242 

fixes #3244 
fixes #3243 

System.CommandLine does not have a robust interactive CLI experience, but Spectre.Console does. Currently, the interactive `dotnet scaffold` tool uses Spectre.Console. Here, all the existing code is leveraged, but the `dotnet scaffold` tool is being invoked with System.CommandLine primarily. If the user wants to use the interactive experience, it will route them through the Spectre.Console experience, and Systme.CommandLine if  `--non-interactive` is specified. 

As part of the non-interactive enhancement work, future work will be building out the `dotnet scaffold` when `--non-interactive` is specified. 

image

image


src/dotnet-scaffolding/dotnet-scaffold/dotnet-scaffold.csproj
```diff
@@ -11,30 +11,20 @@
     <RootNamespace>Microsoft.DotNet.Tools.Scaffold</RootNamespace>
     <NoWarn>$(NoWarn);8002</NoWarn>
   </PropertyGroup>
-  <ItemGroup>
-    <!--
-      TODO: Find a better place for these files to live that is shared between .Core and dotnet-scaffold. Linking them like this is temporary
```
phenning, line 16: I don't have much context here on why we didn't think that dotnet-scaffold would not have this dependency. 

Does this have an impact on deployment or anything else?

haileymck: not sure the context behind this either. it sounds like they were trying to not have a dependency on `Microsoft.DotNet.Scaffolding.Core`, but there was.... can't to seem to figure out why they were trying to avoid it. Seems like just abstraction.

src/dotnet-scaffolding/dotnet-scaffold/Command/CliOptions.cs
```diff
@@ -0,0 +1,10 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.DotNet.Tools.Scaffold.Command
+{
+    internal class CliOptions
```
drewnoakes: nit: `static` class?

src/dotnet-scaffolding/dotnet-scaffold/Command/CliOptions.cs
```diff
@@ -0,0 +1,10 @@
...
+    {
+        internal static string NonInteractiveCliOption = "--non-interactive";
```
drewnoakes: nit: `readonly` or `const` field?

src/dotnet-scaffolding/dotnet-scaffold/Program.cs
```diff
@@ -1,7 +1,45 @@
...
+    }
+    else
+    {
+        var builder = new ScaffoldCommandAppBuilder(context.ParseResult.Tokens.Select(t => t.Value).ToArray());
+        var app = builder.Build();
+        await app.RunAsync();
+    }
+});
+
+runner.RunAsync(args).Wait();
```
drewnoakes: `await` rather than `.Wait()`? While it may not be a problem in this case, we generally avoid calling `Wait` on tasks as it can lead to deadlock issues.

src/dotnet-scaffolding/dotnet-scaffold/Program.cs
```diff
@@ -1,7 +1,45 @@
...
+Option nonInteractiveOption = nonInteractiveScaffoldOption.ToCliOption();
+
+var runner = builder.Build();
+
+// add handler for routing the interactive tool through the Spectre.Console experience, all others
+// are routed through System.CommandLine experience
+builder.AddHandler(async (context) =>
+{
+    var nonInteractive = context.ParseResult.GetValue(nonInteractiveOption);
+    if (nonInteractive is bool isNonInteractiveInvoked && isNonInteractiveInvoked)
```
drewnoakes: nit: can simplify a bit

```suggestion
    if (nonInteractive is true)
```

src/dotnet-scaffolding/dotnet-scaffold/Program.cs
```diff
@@ -1,7 +1,45 @@
...
+    {
+        var builder = new ScaffoldCommandAppBuilder(context.ParseResult.Tokens.Select(t => t.Value).ToArray());
+        var app = builder.Build();
+        await app.RunAsync();
+    }
+});
+
+runner.RunAsync(args).Wait();
+
+static void CreateOptions(out ScaffolderOption<bool> nonInteractiveOption)
```
drewnoakes: Why use an `out` param over a `return` here? Both work, but the out param feels like it's hinting at something more complex.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/IScaffoldRunner.cs
```diff
@@ -15,8 +16,15 @@ public interface IScaffoldRunner
     /// </summary>
     IEnumerable<IScaffolder>? Scaffolders { get; set; }
 
+    IEnumerable<ScaffolderOption>? Options { get; set; }
```
drewnoakes, line 19: Love to see doc comments on interfaces and their members. Especially public ones.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunnerBuilder.cs
```diff
@@ -77,6 +83,35 @@ public IScaffoldBuilder AddScaffolder(string name)
...
+    /// </summary>
+    public void AddOption(ScaffolderOption option)
+    {
+        if (_options is null)
+        {
+            _options = [option];
+        }
+        else
+        {
+            _options = _options.Append(option);
```
drewnoakes: This pattern works (it's close to how lists work in some functional languages, like Lisp), but it'll create a bunch more stuff in memory than just using a simple `List<ScaffolderOption>` and calling `Add`.

Each call of `Enumerable.Append` creates a new object that captures the new item and the previous enumerable. When it comes to enumerating, it ends up creating an enumerator for each item in the sequence.

In contrast, adding to a `List<>` is often "free" (because lists have extra capacity) and then enumerating them involves only a single enumerator allocation.

Possibly not a big deal, but it's good to be in the habit of writing code that has minimal overhead.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/ComponentModel/CliTypes.cs
```diff
@@ -5,7 +5,7 @@ namespace Microsoft.DotNet.Scaffolding.Core.ComponentModel;
 /// <summary>
 /// Defines the types supported by the CLI for parameters.
 /// </summary>
-internal enum CliTypes
+public enum CliTypes
```
drewnoakes: There are several things being made public in this PR. Once we make something public, it's _much_ harder to change it later, which can be quite painful. We don't ever want to break public API.

If we need to share expose these types to another assembly in the same repository, consider just using `InternalsVisibleTo` instead, and keeping the types as `internal`. That provides the access that you need, while also allowing us to change the API in future without worrying about breaking any consumers.

src/dotnet-scaffolding/dotnet-scaffold/Program.cs
haileymck: fixing this here. I opted for Wait() because that is common in this repo, filing #3249 to fix this elsewhere.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/IScaffoldRunner.cs
haileymck: trying to make comments in this repo robust :)

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunnerBuilder.cs
haileymck: ok good to know the difference between Enumerable and List here. updating to List here

src/dotnet-scaffolding/dotnet-scaffold/Program.cs
haileymck: I did this to maintain the style with the other Options in apsnet and aspire,,, but I dont think I will add further options here, updating to `return`

PR #3253, opened 2025-09-24
Use await instead of wait()

### LINKED ISSUE #3249: Replace Wait() with await

haileymck:
via comments in previous pr #3247 

fixes #3249 

There are more `wait()` and `WaitForExit()` to be replaces but the diff was getting huge... the others will be in follow up PR

### DIFF
--- a/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/Telemetry/TelemetryCommonProperties.cs
+++ b/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/Telemetry/TelemetryCommonProperties.cs
@@ -30,10 +30,8 @@ internal class TelemetryCommonProperties
     private const string DockerContainer = "Docker Container";
     private const string MacAddressHash = "Mac Address Hash";
     private const string KernelVersion = "Kernel Version";
-    private const string MacAddressHashCacheKey = "MachineId";
-    private const string IsDockerContainerCacheKey = "IsDockerContainer";
 
-    public Dictionary<string, string> GetTelemetryCommonProperties()
+    public async Task<Dictionary<string, string>> GetTelemetryCommonPropertiesAsync()
     {
         return new Dictionary<string, string>
             {
@@ -42,14 +40,14 @@ internal class TelemetryCommonProperties
                 {RuntimeId, RuntimeEnvironment.GetRuntimeIdentifier()},
                 {ProductVersion, _productVersion},
                 {DockerContainer, IsDockerContainer()},
-                {MacAddressHash, GetMacAddress() },
+                {MacAddressHash, await GetMacAddressAsync() },
                 {KernelVersion, GetKernelVersion()}
             };
     }
 
-    private string GetMacAddress()
+    private async Task<string> GetMacAddressAsync()
     {
-        var macAddress = _environmentService.GetMacAddress();
+        string macAddress = await _environmentService.GetMacAddressAsync() ?? string.Empty;
         return macAddress != null ? macAddress.Hash() : Guid.NewGuid().ToString();
     }
 
--- a/src/dotnet-scaffolding/dotnet-scaffold-aspnet/Program.cs
+++ b/src/dotnet-scaffolding/dotnet-scaffold-aspnet/Program.cs
@@ -1,14 +1,11 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System.Text.Json;
 using Microsoft.DotNet.Scaffolding.CodeModification;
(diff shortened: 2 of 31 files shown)


src/dotnet-scaffolding/dotnet-scaffold/Command/ToolUninstallCommand.cs
```diff
@@ -27,7 +27,7 @@ internal class ToolUninstallCommand(IToolManager toolManager) : Command<ToolUnin
     public override int Execute([NotNull] CommandContext context, [NotNull] ToolUninstallSettings settings)
     {
         // Remove the tool using the provided settings.
-        _toolManager.RemoveTool(settings.PackageName, settings.Global);
+        _toolManager.RemoveToolAsync(settings.PackageName, settings.Global).Wait();
```
haileymck, line 30: not sure how to avoid `Wait()` here because the function signature needs to match for `Spectre.Console`

src/dotnet-scaffolding/dotnet-scaffold/Command/ToolInstallCommand.cs
```diff
@@ -27,7 +27,7 @@ internal class ToolInstallCommand(IToolManager toolManager) : Command<ToolInstal
     public override int Execute([NotNull] CommandContext context, [NotNull] ToolInstallSettings settings)
     {
         // Add the tool using the provided settings.
-        _toolManager.AddTool(settings.PackageName, settings.AddSources, settings.ConfigFile, settings.Prerelease, settings.Version, settings.Global);
+        _toolManager.AddToolAsync(settings.PackageName, settings.AddSources, settings.ConfigFile, settings.Prerelease, settings.Version, settings.Global).Wait();
```
haileymck, line 30: not sure how to avoid `Wait()` here because the function signature needs to match for `Spectre.Console`

src/dotnet-scaffolding/dotnet-scaffold-aspnet/Program.cs
```diff
@@ -294,34 +290,7 @@ public static void Main(string[] args)
...
-                .WithDetectBlazorWasmStep()
-                .WithUpdateAppSettingsStep()
-                .WithUpdateAppAuthorizationStep()
-                .WithEntraAddPackagesStep()
-                .WithEntraBlazorWasmAddPackagesStep()
-                .WithEntraIdCodeChangeStep()
-                .WithEntraIdBlazorWasmCodeChangeStep()
-                .WithEntraIdTextTemplatingStep();
-        }
+        BuildEntraIdAzureScaffolderAsync(builder, projectOption).Wait();
```
haileymck, line 293: not sure how to avoid this `Wait()` other than moving this code out of `main` method here

haileymck:
eeks i thought I pulled main before I created this branch... going to deal with the merge conflicts tomorrow

src/dotnet-scaffolding/dotnet-scaffold-aspnet/Program.cs
drewnoakes: Main methods can be async too! ([doc](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/program-structure/main-command-line#main-return-values))

```c#
public static async Task<int> Main(string[] args)
```

You might be able to convert this to use [top-level statements](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/program-structure/top-level-statements) as well, and remove the class/method wrapper.

src/dotnet-scaffolding/dotnet-scaffold/Command/ToolInstallCommand.cs
drewnoakes: Can we use [`AsyncCommand`](https://spectreconsole.net/api/spectre.console.cli/asynccommand_1/) instead?

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/Telemetry/TelemetryCommonProperties.cs
```diff
@@ -42,14 +40,14 @@ public Dictionary<string, string> GetTelemetryCommonProperties()
...
                 {KernelVersion, GetKernelVersion()}
             };
     }
 
-    private string GetMacAddress()
+    private async Task<string> GetMacAddressAsync()
     {
-        var macAddress = _environmentService.GetMacAddress();
+        string macAddress = await _environmentService.GetMacAddressAsync() ?? string.Empty;
         return macAddress != null ? macAddress.Hash() : Guid.NewGuid().ToString();
```
drewnoakes, line 51: This changes behaviour a bit. With this new code, `macAddress` can never be `null`, so the return string will the hash of the empty string rather than a new GUID.

PR #3262, opened 2025-10-01
invoke "dotnet-scaffold aspire" for aspire commands

haileymck:
After merging the `aspire` scaffolders into `dotnet-scaffold` the `aspire` scaffolders were not appearing in the interactive mode. The way that the interactive scaffolder command is through running `get-commands` for identified scaffolding tools, but it was ignoring `dotnet-scaffold` explicitly.  Now, `aspire` commands can be run in addition to the `aspnet` commands.

Note: this is a temporary fix. I am going to rework the command identification of our own commands through `get-commands`. I want to do this after I merge `aspnet` into the `dotnet-scaffold`.


phenning (approved):

haileymck:
fixes #3242

PR #3265, opened 2025-10-02
Pin Build.Tasks.Core package version, add es-metadata.yml

phenning:


es-metadata.yml
```diff
@@ -0,0 +1,8 @@
+schemaVersion: 0.0.1
+isProduction: true
+accountableOwners:
```
haileymck, line 3: what is this file for? maybe add a comment here

phenning: This is the standard 1es inventory file that is being moved to in places of the service catalog.  if you look in other repos, including WebTools, you’ll see similar there.  I just took what got generated by the automated PR and used it here.

PR #3271, opened 2025-10-08
Invoke Aspire commands directly with System.CommandLine

haileymck:
Currently, there is this middle ground where `aspire` is folded into `dotnet-scaffold`, but `aspnet` has yet to be. Previously, `aspire` and `aspnet` commands were invoked through additional CLI calls.  Now, in this PR because `aspire` is folded into `dotnet-scaffold` , the aspire commands are invoked directly using the `System.CommandLine` runner. After `aspnet` is folded in as well, it will be invoked similarly (and `ValidateUserInputAsync` will be further refactored).

### DIFF
--- a/src/dotnet-scaffolding/dotnet-scaffold/Interactive/Flow/Steps/CommandExecuteFlowStep.cs
+++ b/src/dotnet-scaffolding/dotnet-scaffold/Interactive/Flow/Steps/CommandExecuteFlowStep.cs
@@ -93,28 +93,24 @@ namespace Microsoft.DotNet.Tools.Scaffold.Interactive.Flow.Steps
                 {
                     string command = dotnetToolInfo.Command;
                     string componentExecutionString = $"{command} {string.Join(" ", parameterValues)}";
-                    int? exitCode = null;
-                    AnsiConsole.Status()
+                    int exitCode = AnsiConsole.Status()
                         .Start($"Executing '{command}'", statusContext =>
                         {
                             var cliRunner = dotnetToolInfo.IsGlobalTool ?
                                 DotnetCliRunner.Create(command, parameterValues, envVars) :
                                 DotnetCliRunner.CreateDotNet(command, parameterValues, envVars);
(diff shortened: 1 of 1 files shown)


src/dotnet-scaffolding/dotnet-scaffold/Interactive/Flow/Steps/CommandExecuteFlowStep.cs
```diff
@@ -45,55 +52,72 @@ public ValueTask<FlowStepResult> RunAsync(IFlowContext context, CancellationToke
...
+                }
+                //asp.net
+                else
+                {
+                    string command = dotnetToolInfo.Command;
+                    string componentExecutionString = $"{command} {string.Join(" ", parameterValues)}";
+                    int? exitCode = null;
+                    AnsiConsole.Status()
+                        .Start($"Executing '{command}'", statusContext =>
+                        {
```
phenning: Could we use inline assignment here via https://spectreconsole.net/api/spectre.console/status/b6aa0337 to avoid the nullable?

Something like

```
var exitCode = AnsiConsole.Status()
    .Start($"Executing '{command}'", statusContext =>
    {
        var cliRunner = dotnetToolInfo.IsGlobalTool ?
            DotnetCliRunner.Create(command, parameterValues, envVars) :
            DotnetCliRunner.CreateDotNet(command, parameterValues, envVars);
        
        return cliRunner.ExecuteWithCallbacks(
            (s) => AnsiConsole.Console.MarkupLineInterpolated($"[green]{s}[/]"),
            (s) => AnsiConsole.Console.MarkupLineInterpolated($"[red]{s}[/]"));
    });
```

phenning: This would avoid the nullable exitcode

haileymck: updated this here

PR #3273, opened 2025-10-08
Fold aspnet under dotnet scaffold

haileymck:
This is a huge PR mostly because of all the `aspnet` file moves. The logic is separated out into different commits. 

fixes #3275 #3276 #3277 

commit:
- db23878b : bulk of this PR and the file moves. Also updates `dotnet-scaffold`'s Program.cs to include the aspnet commands 
- d910fb4f: remove the install aspnet and the install all scaffolding scripts, do not recognize `dotnet-scaffold-aspnet` as tool. this is referenced in PR submitted earlier today
- 82c74022: migrate from `Console.WriteLine` to `AnsiConsole.WriteLine` per warnings. Also, rename extensions to be aspire/ aspnet specfic
- 04a99eb6: remove `Microsoft.dotnet-scaffold-aspnet` so it is no longer auto installed on calling dotnet-scaffold
- eccc0c59: invoke aspnet commands directly instead of calling with CLI like done in previous PR for aspire

before:
image

after:
image

image

PR #3282, opened 2025-10-09
Simplify the dotnet scaffold option setting logic

### LINKED ISSUE #3278: Improve the dotnet scaffold aspnet options logic
via: https://github.com/dotnet/Scaffolding/pull/3273#discussion_r2415272092

haileymck:
fixes #3278 

mentioned in #3273 

Currently in the process of refactoring some of this code, this chunk of work cleans up some of the option setting logic. The options are now set in the their respective options file. Note that most of these are static, except the entra ID options where for each user the `username` `tenantid` and `appid` need to be calculated.

### DIFF
--- a/src/dotnet-scaffolding/dotnet-scaffold/AspNet/Commands/AspNetOptions.cs
+++ b/src/dotnet-scaffolding/dotnet-scaffold/AspNet/Commands/AspNetOptions.cs
@@ -1,12 +1,10 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.DotNet.Scaffolding.Core.Builder;
 using Microsoft.DotNet.Scaffolding.Core.ComponentModel;
 using Microsoft.DotNet.Tools.Scaffold.AspNet.Common;
 using Microsoft.DotNet.Tools.Scaffold.AspNet.Helpers;
-using Newtonsoft.Json;
 
 namespace Microsoft.DotNet.Tools.Scaffold.AspNet.Commands;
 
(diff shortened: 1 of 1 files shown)


src/dotnet-scaffolding/dotnet-scaffold/AspNet/Commands/AspNetOptions.cs
```diff
@@ -0,0 +1,258 @@
...
+    public static ScaffolderOption<bool> Overwrite => new()
+    {
+        DisplayName = AspnetStrings.Options.Overwrite.DisplayName,
+        CliOption = Constants.CliOptions.OverwriteOption,
+        Description = AspnetStrings.Options.Overwrite.Description,
+        Required = true,
+        PickerType = InteractivePickerType.YesNo
+    };
+
+    public bool AreAzCliCommandsSuccessful()
```
drewnoakes: This method reads like a getter (which is usually side-effect free (idempotent)) but it actually mutates the state of three fields, which then impacts other properties.

So, for example, the value of `Username` may be incorrect (at the least it will be different) depending upon whether this method was called first or not.

I'd suggest moving the call to `GetAzureInformation` into a constructor so that the fields are always assigned _before_ any other code can start using this type. Then you could make those fields `readonly`. This method could then just be a get-only property.

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Commands/AspNetOptions.cs
```diff
@@ -0,0 +1,258 @@
...
+        _usernames = usernames;
+        _tenants = tenants;
+        _appIds = appIds;
+
+        return isSuccessful;
+    }
+
+    List<string> _usernames = [];
+    List<string> _tenants = [];
+    List<string> _appIds = [];
```
drewnoakes: nit: Add an explicit `private` to these for consistency? (And possibly `readonly` per my other comment).

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Commands/AspNetOptions.cs
```diff
@@ -0,0 +1,258 @@
...
+                    Description = AspnetStrings.Options.Username.Description,
+                    Required = true,
+                    PickerType = InteractivePickerType.CustomPicker,
+                    CustomPickerValues = _usernames
+                };
+                _username = option;
+            }
+            return _username;
+        }
+    }
```
drewnoakes: If you're using C# 14 then this can be written a little more succinctly, and without the need for the `_username` field.

```suggestion
    public ScaffolderOption<string> Username => field ??= new()
    {
        DisplayName = AspnetStrings.Options.Username.DisplayName,
        CliOption = Constants.CliOptions.UsernameOption,
        Description = AspnetStrings.Options.Username.Description,
        Required = true,
        PickerType = InteractivePickerType.CustomPicker,
        CustomPickerValues = _usernames
    };
```

https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-14#the-field-keyword

You might need to add or bump the `<LangVersion>` project property somewhere. I'd suggest setting it to `preview`, as we're in the business of dogfooding.

drewnoakes: That said, is this property called more than once? Does it need to store the value, or can it compute a new one each time? `AspireOptions` doesn't cache values and constructs a new instance on each property getter call. Same for all the properties defined above the `AreAzCliCommandsSuccessful` method. Is there a reason for the different behaviours?

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Commands/AspNetOptions.cs
haileymck: yes I updated to put in the constructor and made them readonly :)

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Commands/AspNetOptions.cs
haileymck: done!

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Commands/AspNetOptions.cs
haileymck: updated to that! yeah `username` `tenantid` `appid` are called multiple times so I wanted to store and reusue the value. the `??=` compiles so I assume the langVersion is ok here... it is `<LangVersion>latest</LangVersion>`

haileymck:
fixes #3242

PR #3286, opened 2025-10-10
execute az cli commands async

haileymck:
- groups the azure usernames, tenantIds, and appIds in the `AzureInformation` record
- ensures that the `az` CLI commands run `async`

part of  work for #3249

### DIFF
--- a/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/CliHelpers/AzCliRunner.cs
+++ b/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/CliHelpers/AzCliRunner.cs
@@ -16,7 +16,7 @@ internal class AzCliRunner
         return new AzCliRunner(commandName);
     }
 
-    public async Task<(int ExitCode, string? StdOut, string? StdErr)> RunAzCliAsync(string arguments)
+    public async Task<(int ExitCode, string? StdOut, string? StdErr)> RunAzCliAsync(string arguments, CancellationToken cancellationToken)
     {
         using var outStream = new ProcessOutputStreamReader();
         using var errStream = new ProcessOutputStreamReader();
@@ -48,11 +48,24 @@ internal class AzCliRunner
         Task taskOut = outStream.BeginRead(process.StandardOutput);
         Task taskErr = errStream.BeginRead(process.StandardError);
 
-        // Wait for process to exit with a timeout (e.g., 30 seconds)
+        // Wait for process to exit with a timeout (e.g., 30 seconds) or cancellation
         const int timeoutMilliseconds = 30000;
-        bool exited = process.WaitForExit(timeoutMilliseconds);
+        Task<bool> waitForExitTask = Task.Run(() => process.WaitForExit(timeoutMilliseconds), cancellationToken);
+        try
+        {
+            await Task.WhenAny(waitForExitTask, Task.Delay(Timeout.Infinite, cancellationToken));
+        }
+        catch (OperationCanceledException)
+        {
+            try
+            {
+                process.Kill(entireProcessTree: true);
+            }
+            catch { }
+            throw;
+        }
 
-        if (!exited)
+        if (!waitForExitTask.Result)
         {
             try
             {
--- a/src/dotnet-scaffolding/dotnet-scaffold/AspNet/Commands/AspNetOptions.cs
+++ b/src/dotnet-scaffolding/dotnet-scaffold/AspNet/Commands/AspNetOptions.cs
@@ -172,9 +172,9 @@ internal class AspNetOptions
         _azureInformation = azureInformation;
     }
 
-    public static async Task<AspNetOptions> CreateAsync()
+    public static async Task<AspNetOptions> CreateAsync(CancellationToken cancellationToken)
     {
-        AzureInformation? azureInfo = await AzCliHelper.GetAzureInformationAsync();
+        AzureInformation? azureInfo = await AzCliHelper.GetAzureInformationAsync(cancellationToken);
         return new AspNetOptions(azureInfo);
     }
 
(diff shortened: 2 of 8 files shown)


src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/CliHelpers/AzCliRunner.cs
```diff
@@ -16,7 +16,7 @@ public static AzCliRunner Create(string? commandName = null)
         return new AzCliRunner(commandName);
     }
 
-    public int RunAzCli(string arguments, out string? stdOut, out string? stdErr)
+    public async Task<(int ExitCode, string? StdOut, string? StdErr)> RunAzCliAsync(string arguments)
```
LittleLittleCloud: Pass a cancellation token and link in `Task.WhenAll` below?

src/dotnet-scaffolding/dotnet-scaffold/Program.cs
```diff
@@ -23,11 +23,13 @@
 Option nonInteractiveOption = nonInteractiveScaffoldOption.ToCliOption();
 
 AspireCommandService aspireCommandService = new(builder);
+
+//aspire command adding does not need to be async
 aspireCommandService.AddScaffolderCommands();
```
LittleLittleCloud, line 34: Why are we adding ScaffolderComamnds twice. on L28 and L32

haileymck: one is `aspire` and the other is `aspnet` commands, they contain different scaffolders

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Commands/AspNetOptions.cs
```diff
@@ -186,7 +187,7 @@ public AspNetOptions()
         Description = AspnetStrings.Options.Username.Description,
         Required = true,
         PickerType = InteractivePickerType.CustomPicker,
-        CustomPickerValues = _usernames
+        CustomPickerValues = _azureInformation!.Usernames
```
drewnoakes, line 190: It doesn't appear to be safe to suppress these null checks. The previous code would have a default empty collection here if getting Azure info failed. This will have null, and this will throw. Consider something like:

```suggestion
        CustomPickerValues = _azureInformation?.Usernames ?? []
```

Ditto below two more times.

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Helpers/AzCliHelper.cs
```diff
@@ -13,60 +13,53 @@ internal class AzCliHelper
...
     {
         try
         {
-            int exitCode = runner.RunAzCli("account list --output json", out var stdOut, out var stdErr);
+            (int exitCode, string?  stdOut, string? stdErr) = await runner.RunAzCliAsync("account list --output json", cancellationToken);
 
             if (stdOut is not null)
             {
                 var result = StringUtil.ConvertStringToArray(stdOut);
                 if (result.Length is 0)
```
drewnoakes, line 52: Unrelated to this PR, but this seems to allocate an array when all it really wants to know is that there was at least one value. If `stdOut` is huge and there are many items, this might be wasteful.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Internal/CliHelpers/AzCliRunner.cs
```diff
@@ -42,19 +42,30 @@ public int RunAzCli(string arguments, out string? stdOut, out string? stdErr)
...
-        var taskOut = outStream.BeginRead(process.StandardOutput);
-        var taskErr = errStream.BeginRead(process.StandardError);
+        Task taskOut = outStream.BeginRead(process.StandardOutput);
+        Task taskErr = errStream.BeginRead(process.StandardError);
 
-        // Wait for process to exit with a timeout (e.g., 30 seconds)
+        // Wait for process to exit with a timeout (e.g., 30 seconds) or cancellation
         const int timeoutMilliseconds = 30000;
-        bool exited = process.WaitForExit(timeoutMilliseconds);
+        Task<bool> waitForExitTask = Task.Run(() => process.WaitForExit(timeoutMilliseconds), cancellationToken);
```
drewnoakes, line 53: Can this use `WaitForExitAsync` instead, and simplify the code below a little? It's .NET 5+ (not sure what targets this needs to run on).

Could make an extensions method for this. Something like:

```c#
public static async Task<bool> WaitForExitAsync(
    this Process process,
    TimeSpan timeout,
    CancellationToken cancellationToken = default)
{
    using var timeoutCts = new CancellationTokenSource(timeout);
    using var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, timeoutCts.Token);

    try
    {
        await process.WaitForExitAsync(linkedCts.Token);

        return !timeoutCts.IsCancellationRequested;
    }
    catch (OperationCanceledException) when (timeoutCts.IsCancellationRequested)
    {
        // Timed out.
        return false;
    }
}
```

Another option would be to register a callback on the cancellation token that performs the kill.

src/dotnet-scaffolding/dotnet-scaffold/Aspire/AspireCommandService.cs
```diff
@@ -74,4 +74,10 @@ public void AddScaffolderCommands()
                .WithStorageAddPackageSteps()
                .WithStorageCodeModificationSteps();
     }
+
+    public Task AddScaffolderCommandsAsync(CancellationToken cancellationToken)
+    {
+        //ignore
+        throw new NotImplementedException();
```
drewnoakes, line 81: It's not obvious to me why this is safe to ignore. Can it be described in the comment?

PR #3290, opened 2025-10-24
Blazor CRUD uses `NotFound()` method and re-execution middleware

ilonatommy:
This PR applies changes from https://github.com/dotnet/Scaffolding/pull/3162 to older version of scaffolder. See the original PR to understand what changes were requested.

Additionally it:
- Adds `ModifyRoutesRazorAsync` in `BlazorWebCRUDGenerator.cs`, that applies the replacements from `json` to the existing project files. Until now, we used to have only such updates done to Program.cs and that's not enough. NotFound requires changes in `Router.razor`.
- `NotFound.razor` is a page not directly connected to CRUD, so it should not stay in CRUD dir, see the project tree. That's why `ValidateAndGetOutputPath` makes an exception for not found page location. The project tree view with this change:
image

- `BlazorWebCRUDGeneratorTests` now checks the new not found file presence.
- Re-execution middleware is added to Program.cs.

Diff for the changes applied to:
- `dotnet new blazor` [blazor-server.diff.txt](https://github.com/user-attachments/files/23123740/blazor-server.diff.txt)
- `dotnet new blazor -int WebAssembly` [blazor-wasm.diff.txt](https://github.com/user-attachments/files/23124327/blazor-wasm.diff.txt)

# Concerns/questions:

- I embedded a fixed layout in `NotFound.razor`, otherwise it would look like this:
image 
instead of:
image

I would wish to make it more dynamic but I don't have a clue how. Can we assume that `MainLayout` exists in the project just like that? Can we save the layout name in a variable and then use it in the `NotFound.tt`? 

# Solution

According the the feedback below, we decided to add `MainLayout` conditionally to `NotFound.tt`, only if it exists in the project.

### DIFF
--- a/src/Scaffolding/VS.Web.CG.Mvc/Blazor/BlazorWebCRUDGenerator.cs
+++ b/src/Scaffolding/VS.Web.CG.Mvc/Blazor/BlazorWebCRUDGenerator.cs
@@ -285,6 +285,16 @@ namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Blazor
             var allTypes = await ModelTypesLocator.GetAllTypesAsync();
             var programType = ModelTypesLocator.GetType("<Program>$").FirstOrDefault() ?? ModelTypesLocator.GetType("Program").FirstOrDefault();
             var project = Workspace.CurrentSolution.Projects.FirstOrDefault(p => p.AssemblyName.Equals(ProjectContext.AssemblyName, StringComparison.OrdinalIgnoreCase));
+            if (!string.IsNullOrEmpty(project?.FilePath))
+            {
+                var projectDirectory = Path.GetDirectoryName(project.FilePath);
+                if (!string.IsNullOrEmpty(projectDirectory) && FileSystem.DirectoryExists(projectDirectory))
+                {
+                    blazorAppProperties.HasMainLayout = FileSystem
+                        .EnumerateFiles(projectDirectory, "MainLayout.razor", SearchOption.AllDirectories)
+                        .FirstOrDefault() != null;
+                }
+            }
             var programDocument = project.GetUpdatedDocument(FileSystem, programType);
             var programDocumentDirectory = Path.GetDirectoryName(programDocument.FilePath);
 
--- a/src/Scaffolding/VS.Web.CG.Mvc/Templates/Blazor/NotFound.cs
+++ b/src/Scaffolding/VS.Web.CG.Mvc/Templates/Blazor/NotFound.cs
@@ -25,9 +25,12 @@ namespace Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Templates.Blazor
         /// </summary>
         public virtual string TransformText()
         {
-            this.Write("\r\n\r\n@page \"/not-found\"\r\n@layout MainLayout\r\n\r\n\r\n<PageTitle>Not Found</PageTitle>\r" +
-                    "\n\r\n\r\n<h1>Not Found</h1>\r\n<p>Sorry, there\'s nothing at this address.</p>\r\n\r\n\r\n<a " +
-                    "href=\"/\">Back to Home</a>");
+            this.Write("\r\n\r\n@page \"/not-found\"\r\n");
+ if (Model.HasMainLayout) { 
+            this.Write("@layout MainLayout\r\n");
+ } 
+            this.Write("\r\n\r\n<PageTitle>Not Found</PageTitle>\r\n\r\n\r\n<h1>Not Found</h1>\r\n<p>Sorry, there\'s n" +
+                    "othing at this address.</p>\r\n\r\n\r\n<a href=\"/\">Back to Home</a>");
             return this.GenerationEnvironment.ToString();
         }
         private global::Microsoft.VisualStudio.TextTemplating.ITextTemplatingEngineHost hostValue;
--- a/src/Scaffolding/VS.Web.CG.Mvc/Templates/Blazor/NotFound.tt
+++ b/src/Scaffolding/VS.Web.CG.Mvc/Templates/Blazor/NotFound.tt
@@ -6,7 +6,9 @@
 
 
 @page "/not-found"
+<# if (Model.HasMainLayout) { #>
 @layout MainLayout
+<# } #>
 
 
 <PageTitle>Not Found</PageTitle>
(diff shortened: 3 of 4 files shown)


src/Scaffolding/VS.Web.CG.Mvc/Templates/Blazor/NotFound.cs
```diff
@@ -0,0 +1,383 @@
...
+    /// </summary>
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.VisualStudio.TextTemplating", "17.0.0.0")]
+    public partial class NotFound : NotFoundBase
+    {
+        /// <summary>
+        /// Create the template output
+        /// </summary>
+        public virtual string TransformText()
+        {
+            this.Write("@page \"/not-found\"\r\n\r\n<PageTitle>Not Found</PageTitle>\r\n\r\n<h1>Not Found</h1>\r\n<p" +
```
haileymck: would it make sense to write this async?

src/Scaffolding/VS.Web.CG.Mvc/Templates/Blazor/NotFound.cs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+/// <summary>
+/// Initialize the template
+/// </summary>
+public virtual void Initialize()
+{
+    if ((this.Errors.HasErrors == false))
+    {
+bool ModelValueAcquired = false;
```
haileymck, line 73: nit: spacing

src/Scaffolding/VS.Web.CG.Mvc/Templates/Blazor/NotFound.cs
```diff
@@ -0,0 +1,383 @@
...
+            }
+            this.currentIndentField = (this.currentIndentField + indent);
+            this.indentLengths.Add(indent.Length);
+        }
+        /// <summary>
+        /// Remove the last indent that was added with PushIndent
+        /// </summary>
+        public string PopIndent()
+        {
+            string returnValue = "";
```
haileymck, line 302: consider using `StringBuilder` for these string manipulations

src/Scaffolding/VS.Web.CG.Mvc/Blazor/BlazorWebCRUDGenerator.cs
```diff
@@ -200,6 +204,33 @@ internal async Task ModifyProgramCsAsync(BlazorWebAppProperties appProperties)
...
 
+        internal async Task ModifyRoutesRazorAsync()
+        {
+            CodeModifierConfig blazorChangesConfig = GetBlazorCodeModifierConfig();
+            if (blazorChangesConfig is null)
+            {
+                return;
+            }
+
+            var routesRazorFile = blazorChangesConfig.Files.FirstOrDefault(x => x.FileName.Equals("Routes.razor", StringComparison.OrdinalIgnoreCase));
```
haileymck, line 215: is this the full path?

src/Scaffolding/VS.Web.CG.Mvc/Templates/Blazor/NotFound.cs
ilonatommy: I think not. This generated file is following the pattern from other `src\Scaffolding\VS.Web.CG.Mvc\Templates\Blazor\.*.cs` files with `TransformText`. They never use async writing and if we wanted to introduce it, I believe we would have to apply wider refactoring.

It can be done in a separate PR if you decide you want to change your approach but I am not sure if it will be possible. From what I gathered, T4 (Text Template Transformation Toolkit) engine and the writing helpers are intentionally synchronous and T4 does not expose async version [docs](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.texttemplating.texttransformation.write?view=visualstudiosdk-2022), and the base class is built around synchronous string-building in memory.

src/Scaffolding/VS.Web.CG.Mvc/Templates/Blazor/NotFound.cs
ilonatommy: This file is automatically generated by T4 based on `.*tt` files. We should not expect it to look pretty.

src/Scaffolding/VS.Web.CG.Mvc/Templates/Blazor/NotFound.cs
ilonatommy: Same as above, we should not edit it, unless you change the way T4 works.

src/Scaffolding/VS.Web.CG.Mvc/Blazor/BlazorWebCRUDGenerator.cs
ilonatommy: No. It is the same pattern we are using in other places, see e.g. `ModifyProgramCsAsync`. Is intentionally just "Routes.razor". `GetDocumentFromName` searches the project directory recursively and picks the file whose path ends with that name, so we don’t need (and actually don’t want) to hard-code the full path. This covers server and WebAssembly projects even if the file lives in different folders.

src/Scaffolding/VS.Web.CG.Mvc/Templates/Blazor/NotFound.cs
haileymck: good to know! thanks!

src/Scaffolding/VS.Web.CG.Mvc/Templates/Blazor/NotFound.cs
haileymck: ok makes sense to leave as is. Thanks for the additional context here!

haileymck (dismissed):
Looks good to me!

src/Scaffolding/VS.Web.CG.Mvc/Templates/Blazor/NotFound.tt
```diff
@@ -0,0 +1,19 @@
+<#@  template hostSpecific="true" linePragmas="false" #>
+<#@  parameter type="Microsoft.VisualStudio.Web.CodeGenerators.Mvc.Blazor.BlazorModel" name="Model" #>
+<#@ import namespace="System.Collections.Generic" #>
+<#@ import namespace="System.Text" #>
+<#@ import namespace="System.Linq" #>
+
+
+@page "/not-found"
+@layout MainLayout
```
danroth27, line 10: Could we conditionally apply MainLayout as the layout for the NotFound.razor page depending on whether MainLayout exists? Something like:

```suggestion
<# if (Model.HasMainLayout) { #>
@layout MainLayout
<# } #>
```

PR #3304, opened 2025-11-05
Fix writing to console breaking UI

### LINKED ISSUE #3303: Sometimes the other scaffolders get swallowed when ad app list fails
1) `az login` to subscription that you don't have full permissions to where `az ad app list` fails
2) run `dotnet scaffold` 
3) see:

https://github.com/user-attachments/assets/c6dee6d1-4208-4600-ab40-fba2a08a49c9

Image

haileymck:
When testing `dotnet scaffold`, I was trying a subscription I normally don't use. With this subscription, I don't have full permissions. The `az ad app list` failed because of this lack of permissions. Previously, when any of the `az` commands an error message is written to the CLI, but these `az` commands fail during the construction of the CLI tools. This writing of errors was disrupting the visibility of the Scaffolders (even though they were available). Not writing these errors to the CLI allows the Scaffolders to appear correctly. 

To see this issue see bug.

fixes #3303 

Instead of surfacing these `az cli` errors when they occur using `AnsiConsole`, the errors will be passed and later surfaced after it is safe to render the UI. These errors are added to an error service and using DI are retrieved to be rendered with the other elements of the UI. An image of what this UI will look like is displayed below. 

update: 
shows the error late in the flow
image

### DIFF
--- a/src/dotnet-scaffolding/dotnet-scaffold/AspNet/Helpers/AzCliHelper.cs
+++ b/src/dotnet-scaffolding/dotnet-scaffold/AspNet/Helpers/AzCliHelper.cs
@@ -15,21 +15,35 @@ internal class AzCliHelper
     /// <param name="usernames">the user IDs</param>
     /// <param name="tenants">the tenant IDs</param>
     /// <param name="appIds">the app IDs</param>
+    /// <param name="azCliErrors">the errors from the Az CLI if there are any</param>
     /// <returns>if successful, return true</returns>
-    public static bool GetAzureInformation(out List<string> usernames, out List<string> tenants, out List<string> appIds)
+    public static bool GetAzureInformation(out List<string> usernames, out List<string> tenants, out List<string> appIds, out string? azCliErrors)
     {
         // Create a runner to execute the 'az account list' command with json output format
         var runner = AzCliRunner.Create();
 
-        if (EnsureUserIsLoggedIn(runner, out string? output) && !string.IsNullOrEmpty(output))
+        if (EnsureUserIsLoggedIn(runner, out string? output, out string? loginErrors) && !string.IsNullOrEmpty(output))
         {
-            if (GetAzureUsernamesAndTenatIds(runner, output, out usernames, out tenants))
+            if (GetAzureUsernamesAndTenatIds(output, out usernames, out tenants, out string? usernameTenantError))
             {
-                if (GetAzureAppIds(runner, out appIds))
+                if (GetAzureAppIds(runner, out appIds, out string? appErrors))
                 {
+                    azCliErrors = null;
                     return true;
                 }
+                else
+                {
+                    azCliErrors = appErrors;
+                }
             }
+            else
+            {
+                azCliErrors = usernameTenantError;
+            }
+        }
+        else
+        {
+            azCliErrors = loginErrors;
         }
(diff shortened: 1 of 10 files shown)


src/dotnet-scaffolding/dotnet-scaffold/AspNet/Helpers/AzCliHelper.cs
```diff
@@ -62,10 +61,9 @@ private static bool EnsureUserIsLoggedIn(AzCliRunner runner, out string? output)
             output = stdOut;
             return exitCode == 0 && string.IsNullOrEmpty(stdErr);
         }
-        catch (Exception ex)
+        catch (Exception)
         {
             output = null;
-            AnsiConsole.WriteLine($"Error checking Azure login status: {ex.Message}");
```
drewnoakes, line 68: The methods in this class seem like generic helper methods. If they're called in other circumstances, then the removal of this logging might make it harder to diagnose issues.

My understanding of the original bug here is that the error message isn't helpful in that scenario. Are there other scenarios where it might be helpful?

If so, one option would be to have an `out string? errorMessage` parameter on these methods so that the caller can decide if they want to do anything with the error message. The previous console logging wasn't very flexible, but removing all error reporting might cause a different kind of issue.

haileymck: maybe I wasn't clear enough in the bug, but what is happening here is that writing error messages when rendering the CLI experience causes a disruption in the layout. Totally agree that it is unideal to remove the logging, not sure if there is a way to still log the issues and present them to the user until after the app is spun up. Passing the error message to the caller wouldn't really help much because there are layers and layers that occur during the render process. I think ideally this work wouldn't be done on render and instead dynamically and then I could surface the errors when they occur, but that would be a major refactor... going to look into it though

haileymck:
Now these error messaged are passed around and surface later in the UI flow. (see updated description).

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Helpers/AzCliHelper.cs
```diff
@@ -165,12 +185,19 @@ private static bool GetAzureAppIds(AzCliRunner runner, out List<string> appIds)
                     return true;
                 }
             }
+
+            if (!string.IsNullOrEmpty(stdErr))
```
drewnoakes, line 189: This PR looks great! Much better to surface this information to the user.

My only feedback is that the error message looks much scarier than a simple auth issue. I wonder if we could make it clearer somehow.

i.e. instead of:

> Error configuring the Entra ID Scaffolder with az CLI error: Error executing 'az ad app list': ERROR: Insufficient privileges to complete the operation.

Could the message somehow be closer to:

> The "Entra ID" category is unavailable due to an authentication or authorization issue. Ensure the 'az ad app list' command works if you want to scaffold items from this category.

I'm putting this comment here, as I wonder if the exit code would reveal the auth issue better than just copying stdout.

haileymck: update! 
image

drewnoakes (approved):

PR #3307, opened 2025-11-06
Install msidentity for entra id if not installed

### LINKED ISSUE #3305: dotnet scaffolder needs to install msidentity when entraId is run
run: `dotnet tool install -g Microsoft.dotnet-msidentity ` for the user

need to pass the environment parameters to determine if the -g needs to be passed

Image

haileymck:
Fixes #3305 

If the `dotnet msidentity` tool is not installed, the EntraID secret setting scaffolder will fail. This code added checks to see if this tool is installed, and if not it installs it locally, allowing the Entra ID scaffolder to successfully complete. 

before:
ms_identity_before

after:
ms_identity_after


drewnoakes (approved):
Looks great.

PR #3312, opened 2025-11-11
Couple NuGet packages with package versions

### LINKED ISSUE #3310: Allow for specific NuGet Package versions to be installed
right now the latest available (either prerelease or regular) NuGet packages are installed with `dotnet scaffold` within Visual Studio, specific package versions will need to be used. Allow for this.

haileymck:
Right now, only the newest NuGet package versions are installed with `dotnet scaffold`. This will need to change when `dotnet scaffold` CLI commands are used inside of VS. This is an early step in the process to get there, allowing for specified package versions to be installed. 

fixes #3310

### DIFF
--- a/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Model/Package.cs
+++ b/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Model/Package.cs
@@ -1,0 +1,17 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.DotNet.Scaffolding.Core.Model;
+
+/// <summary>
+/// Represents a NuGet package with a specified name and optional version.
+/// </summary>
+/// <param name="name">The name of the package. Cannot be null.</param>
+/// <param name="version">The version of the package, or null to indicate the latest version.</param>
+public class Package(string name, string? version = null)
+{
+    public string Name { get; } = name;
+
+    //null version indicates latest and version specfication is not needed
+    public string? Version { get; } = version;
+}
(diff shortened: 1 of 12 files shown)


src/dotnet-scaffolding/dotnet-scaffold/AspNet/Extensions/RazorPagesScaffolderBuilderExtensions.cs
```diff
@@ -113,7 +114,7 @@ public static IScaffoldBuilder WithRazorPagesAddPackagesStep(this IScaffoldBuild
                     packageList.Add(projectPackageName);
                 }
 
-                step.PackageNames = packageList;
+                step.Packages = [.. packageList.Select(p => new Package(p))];
```
drewnoakes, line 117: nit: now that we have a type for `Package`, consider renaming `packageList` -> `packageNames`.

If you want to explore this, there are a few instances of this pattern throughout the PR.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Model/Package.cs
```diff
@@ -0,0 +1,17 @@
...
+/// </summary>
+/// <param name="name">The name of the package. Cannot be null.</param>
+/// <param name="version">The version of the package, or null to indicate the latest version.</param>
+public class Package(string name, string? version = null)
+{
+    public string Name { get; } = name;
+
+    //null version indicates latest and version specfication is not needed
+    public string? Version { get; } = version;
+}
```
drewnoakes, line 17: nit, could use a record class:

```suggestion
/// <summary>
/// Represents a NuGet package with a specified name and optional version.
/// </summary>
/// <param name="Name">The name of the package. Cannot be null.</param>
/// <param name="Version">The version of the package, or null to indicate the latest version.</param>
public sealed record class Package(string name, string? version = null);
```

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Steps/AddPackagesStep.cs
```diff
@@ -13,7 +14,7 @@ public class AddPackagesStep : ScaffoldStep
     /// <summary>
     /// Gets or sets the list of package names to add.
     /// </summary>
-    public required IList<string> PackageNames { get; set; }
+    public required IList<Package> Packages { get; set; }
```
drewnoakes, line 17: Does this have to be a mutable `IList<>`, vs say `IReadOnlyList<>` or `ImmutableArray<>`?

haileymck: I missed these comments so I opened #3316 sorry!!!

PR #3319, opened 2025-11-18
Update to net 10 GA packages

phenning:


src/dotnet-scaffolding/Directory.Packages.props
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
     <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
-    <NoWarn>$(NoWarn);NU1507</NoWarn>
+    <NoWarn>$(NoWarn);NU1507;NU1701</NoWarn>
```
haileymck: is there a reason that `NU1701` was added here?
https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1701

KUTlime:
@phenning can we move on with this PR? It blocks me from migrating to .NET 10. It has been a week since the .NET 10 release, MSFT has been full of cheer about the upgrade, but it still has missing components.

BTW: @haileymck has a good point about the NU1701 warning suppression. I know what it does, but I don't get why it has been added. The PR has no explanation.

src/dotnet-scaffolding/Directory.Packages.props
phenning: removed

PR #3326, opened 2025-11-20
Get Entra ID specific param when Entra Scaffolder is running. not on startup

### LINKED ISSUE #89: Package source caching in project.json intellisense is too aggressive
I changed my package sources, closed the project.json and reopened it, and it didn't update.

haileymck:
Previously, there was no way to dynamically calculate parameters needed for Entra ID. Some of these parameters take a while because they require a few `az` CLI command calls. Previously, to start up `dotnet scaffold` tool it would always run these commands, but that is not needed. These commands are now only called when Entra ID is selected

When the commands fail the current experience is below. Note we do not need the start up error service anymore (this was initially added because the az commands were failing on start up and it allowed the errors to be surfaced). Now because the parameters are calculated dynamically they fail in the flow. Like below.
error

Future work:
Add spinner when the az cli commands are running. Doing this later because it will be a little more involved and is not essential in this perf fix. 

fixes #89 

before:
before_perffix

after:

perffix


src/dotnet-scaffolding/dotnet-scaffold/AspNet/Helpers/AzCliHelper.cs
```diff
@@ -4,11 +4,79 @@
 using System.Text.Json;
 using Microsoft.DotNet.Scaffolding.Internal;
 using Microsoft.DotNet.Scaffolding.Internal.CliHelpers;
+using Spectre.Console.Flow;
 
 namespace Microsoft.DotNet.Tools.Scaffold.AspNet.Helpers;
 
 internal class AzCliHelper
```
drewnoakes: Can this class be static?

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Helpers/AzCliHelper.cs
```diff
@@ -17,7 +85,7 @@ internal class AzCliHelper
     /// <param name="appIds">the app IDs</param>
     /// <param name="azCliErrors">the errors from the Az CLI if there are any</param>
     /// <returns>if successful, return true</returns>
-    public static bool GetAzureInformation(out List<string> usernames, out List<string> tenants, out List<string> appIds, out string? azCliErrors)
+    private static void GetAzureInformation(out List<string> usernames, out List<string> tenants, out List<string> appIds, out string? azCliErrors)
```
drewnoakes: This method is now only called in one place. Consider inlining it directly into `SetAzureProperties`, to simplify the APIs a bit.

drewnoakes (approved):

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Helpers/AzCliHelper.cs
haileymck: updated!

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Helpers/AzCliHelper.cs
haileymck: done!

PR #3327, opened 2025-11-21
Merge changes from release/10.0 to main

phenning:


src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffoldRunner.cs
```diff
@@ -37,21 +37,22 @@ public async Task RunAsync(string[] args)
...
-    public void AddHandler(Func<InvocationContext, Task> handle)
+    public void AddHandler(Func<ParseResult, CancellationToken, Task> handle)
     {
         if (RootCommand is null)
         {
             throw new InvalidOperationException("RootCommand is not set.");
         }
 
-        RootCommand.SetHandler(handle);
+        RootCommand.SetAction(handle);
```
haileymck, line 56: if set action is ultimately called, it may make sense to rename all these add handlers to `setAction`

PR #3353, opened 2025-11-25
add script to install the nuget packages published from pipeline on main

haileymck:
Adding that can be used to download the most recent pipeline generated NuGet packages for those that are internal to Microsoft. To be able to download them, there are some requirements outlined in the `Getting-Started.md` file. This includes being signed in to `az cli` and having an AzDo `PAT`. The script uninstalls `dotnet-aspnet-codegenerator` `microsoft.dotnet-msidentity` `microsoft.dotnet-scaffold ` if they are already on the machine and installs the version published from the pipeline.

### DIFF
--- a/Getting-Started.md
+++ b/Getting-Started.md
@@ -56,3 +56,20 @@ I recommend cloning it under your user profile directory.
 - `.config` folders in the root of either the scaffolding project or the project you're scaffolding.
 - **Missing or overlapping SDK versions**:
   - I encountered issues with two .NET 10.0 preview versions and had to delete the newer one.
+
+# How to Install the most recent packages
+
+0) generate a personal access token (PAT).
+   a. go to Azure DevOps
+   b. click the User settings next to your picture in the top right
+   c. click on Personal Access Token toward the bottom of the menu
+   d. click "+ New Token" and copy it. You will need this token to run the following scripts
+
+1) Log in to az CLI with your Microsoft account with `az login`
+
+2) set your PAT environment variable with running `$env:PAT = [your PAT]`
+
+2) run `scripts\download-artifacts.ps1`
+optionally, specify the branch for example: `powershell.exe -File c:\Scaffolding\scripts\download-artifacts.ps1 -Branch release/10.0`, branch defaults to main
+
+3) Ensure that the correct versions are installed. The installed packages should be Microsoft.dotnet-msidenity, dotnet-aspnet-codegenerator, and Microsoft.dotnet-scaffold. 
--- a/scripts/download-artifacts.ps1
+++ b/scripts/download-artifacts.ps1
@@ -1,0 +1,85 @@
+# Download NuGet packages from Azure DevOps pipeline and install them locally
+# Requires: Azure CLI, .NET SDK
+
+param(
+    [string]$Organization = "dnceng",
+    [string]$Project = "internal",
+    [string]$DefinitionId = "345",
+    [string]$ArtifactName = "PackageArtifacts",
+    [string]$Branch = "main"
+)
+
+# Ensure Azure CLI is installed
+if (-not (Get-Command az -ErrorAction SilentlyContinue)) {
+    Write-Error "Azure CLI (az) is not installed. Please install it first."
+    exit 1
+}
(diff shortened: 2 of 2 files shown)


phenning (dismissed):

Getting-Started.md
```diff
@@ -56,3 +56,17 @@ I recommend cloning it under your user profile directory.
...
+
+0) generate a personal access token (PAT).
+   a. go to Azure DevOps
+   b. click the User settings next to your picture in the top right
+   c. click on Personal Access Token toward the bottom of the menu
+   d. click "+ New Token" and copy it. You will need this token to run the following scripts
+
+1) Log in to az CLI with your Microsoft account with `az login`
+
+2) run `scripts\download-artifacts.ps1 -PatToken [your PAT here]`
```
JakeRadMSFT: You can get pat token with az command if needed.

haileymck: oh interesting

scripts/download-artifacts.ps1
```diff
@@ -0,0 +1,89 @@
+# Download NuGet packages from Azure DevOps pipeline and install them locally
+# Requires: Personal Access Token (PAT) with access to build artifacts
```
JakeRadMSFT: If you have this script generate the link and post it as a comment ... the token can come from the CI pipeline.

JakeRadMSFT: I'm fine either way :)

Getting-Started.md
haileymck: can you link the command @JakeRadMSFT I looked for it but I thought it didn't exist...

scripts/download-artifacts.ps1
```diff
@@ -0,0 +1,89 @@
...
+$response = Invoke-RestMethod -Uri $artifactUrl -Headers @{Authorization = "Basic $base64AuthInfo"}
+$downloadUrl = $response.resource.downloadUrl
+
+if (-not $downloadUrl) {
+    Write-Host "Could not find download URL for artifact $ArtifactName."
+    exit 1
+}
+
+Write-Host "Downloading artifact zip..."
+Invoke-WebRequest -Uri $downloadUrl -OutFile "$Destination.zip" -Headers @{Authorization = "Basic $base64AuthInfo"}
```
JakeRadMSFT: I would have it use az commands as much as possible instead of direct web requests.

haileymck: 🫡

scripts/download-artifacts.ps1
JakeRadMSFT: Actually I like this approach of a file ... then you could have a github action that posts a comment and just always downloads this file and runs it.

JakeRadMSFT: That's what aspire did?

https://github.com/dotnet/aspire/pull/12950#issuecomment-3525706413

scripts/download-artifacts.ps1
haileymck: I can't use `az pipelines runs artifact download ` because the way the pipeline is set up the artifacts are "BuildArtifacts" and not "pipeline artifacts". there isnt an az command to download build artifacts.

haileymck: also because this is an arcade repo can't modify the infra that much

Getting-Started.md
haileymck: ok so i dont think there is a way to get the PAT with az, the best practice is setting it as an environment variable. I outlined this in the getting started doc

scripts/download-artifacts.ps1
haileymck: yes, it would require the users setting their PAT as a github actions secret

JakeRadMSFT: Yeah it's the odd mix of open source and private pipelines that makes this difficult.

Getting-Started.md
JakeRadMSFT: https://learn.microsoft.com/en-us/cli/azure/account?view=azure-cli-latest#az-account-get-access-token

PR #3367, opened 2025-12-01
Update Version.Details.xml with correct versions and commit shas

phenning:
This commit updates the Version.Details.xml with the updated versions to enable the darc subscriptions to flow properly.

Prior to this change, we had made some manual updates directly to Versions.props, which broke the automatic dependency flow.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -7,81 +7,81 @@
     </Dependency>
   </ToolsetDependencies>
   <ProductDependencies>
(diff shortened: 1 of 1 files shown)


phenning:
> @phenning There are still no updated NuGet versions. Why?
> 
> image

Because we have a release process, which includes configuration and vendor testing for both Visual Studio 2026, 2022 and the CLI tool.

Expect a release this week.  Keep in mind that this Release also includes updates for 10 GA, 9.0.11 and 8.0.22.

PR #3374, opened 2025-12-01
Add --framework option for dotnet scaffolding

### LINKED ISSUE #3311: add --framework option that can be used to get nuget package versions
Add the ability to specify certain .NET target frameworks (net 8, 9, 10) and install the NuGet packages that are compatible with them. This involves adding a `--target-framework` optional command option, parsing to determine it is formatted correctly. Then, find the corresponding NuGet package for the installation.

haileymck:
Fixes #3311 

In efforts to support using `dotnet scaffold` tool in VS, specific target frameworks will need to be recognized in order to install the corresponding NuGet package. This PR adds the optional `--framework` option to aspnet commands so that users are able to specify the desired framework. Eventually, VS will use this option as well when invoke the CLI took. `dotnet scaffold aspire` does not have a `--framework` option right now because its packages right now are all net8.0-net10.0 compatible, but it can be added in the future if required

The package versions are obtained for NuGet feeds. Currently intended to support net8.0, net9.0, net10.0. These frameworks were selected based on the overall long-term and standard term support. Also note, that many of the aspire NuGet packages that `dotnet scaffold` installs does not have a version available for net7.0. 

This PR changes the interactive flow a bit! The target framework selection is before the prerelease selection. Prerelease only applies to the most recent .NET major version, so if the user selects Net 8 or net 9, prerelease does not show. 

interactive flow: 
update
updateagain

logs version: 
image

help:
image

### DIFF
--- a/src/dotnet-scaffolding/Directory.Packages.props
+++ b/src/dotnet-scaffolding/Directory.Packages.props
@@ -26,6 +26,7 @@
     <PackageVersion Include="Mono.TextTemplating" Version="$(MonoTextTemplatingVersion)" />
     <PackageVersion Include="Newtonsoft.Json" Version="$(NewtonsoftJsonPackageVersion)" />
     <PackageVersion Include="NuGet.Configuration" Version="$(NugetProtocolVersion)" />
+    <PackageVersion Include="NuGet.Credentials" Version="$(NugetProtocolVersion)" />
     <PackageVersion Include="NuGet.Protocol" Version="$(NugetProtocolVersion)" />
     <PackageVersion Include="NuGet.Versioning" Version="$(NugetProtocolVersion)" />
     <PackageVersion Include="Spectre.Console.Flow" Version="$(SpectreConsoleFlowVersion)" />
--- a/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/NuGetVersionService.cs
+++ b/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/NuGetVersionService.cs
@@ -5,6 +5,7 @@ using NuGet.Protocol.Core.Types;
 using NuGet.Versioning;
 using NuGet.Configuration;
 using NuGet.Protocol;
+using NuGet.Credentials;
 
 namespace Microsoft.DotNet.Scaffolding.Core.Model;
 
@@ -19,6 +20,10 @@ internal class NuGetVersionService
     public NuGetVersionService(IEnvironmentService environmentService)
     {
         _environmentService = environmentService;
+
+        DefaultCredentialServiceUtility.SetupDefaultCredentialService(
+            NuGet.Common.NullLogger.Instance,
+            nonInteractive: false);
 
         string settingsPath = _environmentService.CurrentDirectory;
         _settings = Settings.LoadDefaultSettings(settingsPath);
--- a/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Microsoft.DotNet.Scaffolding.Core.csproj
+++ b/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Microsoft.DotNet.Scaffolding.Core.csproj
@@ -9,6 +9,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Hosting" />
     <PackageReference Include="NuGet.Configuration" />
+    <PackageReference Include="NuGet.Credentials" />
     <PackageReference Include="NuGet.Protocol" />
     <PackageReference Include="NuGet.Versioning" />
     <PackageReference Include="System.CommandLine" />


src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffolderOption.cs
```diff
@@ -15,31 +15,42 @@ public abstract class ScaffolderOption
...
+
     /// <summary>
     /// Custom picker values for interactive UI.
     /// </summary>
     public IEnumerable<string>? CustomPickerValues { get; init; } = null;
 
+    /// <summary>
+    /// Gets or sets the message displayed to the user during an interactive prompt.
+    /// </summary>
+    public string? InteractivePromptMessage { get; init; } = null;
```
haileymck: allows for specifying more robust descriptions in interactive mode

src/dotnet-scaffolding/dotnet-scaffold/ScaffoldingSteps/ValidateTargetFrameworkStep.cs
```diff
@@ -0,0 +1,104 @@
...
+        "net46",
+        "net452",
+        "net451",
+        "net45",
+        "net403",
+        "net40",
+        "net35",
+        "net20",
+        "net11"
+    ];
```
drewnoakes: The `TargetFramework` string is difficult (impossible?) to validate against in the general case. Maybe there are some constraints here that would make that easier, but I am a bit concerned about having a list of these values here for a few reasons.

1. There's some normalisation of these values. For example, `net4.7.2` and `net472` are the same thing. Similarly `net5` and `net5.0` (and maybe `net50`) are all "valid".
2. These values should be treated as aliases. That is, a customer can define their own `TargetFramework` values and have them map to the "real" values under the covers (like `TargetFrameworkIdentifier`, `TargetFrameworkVersion` etc.). See also the guideline [here](https://github.com/dotnet/project-system/blob/main/docs/repo/coding-conventions.md#data).
3. What if the user is targeting a particular platform, like `net8.0-windows`?
4. This data goes out of date on a regular basis. When `net11.0` is released, users of this tool will file bugs.

I'm still reading through this PR, but I would hope we can find a way to populate a list of values that we let the user choose between, and that we source the data for that list programmatically, such as from the packages themselves.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/CommandLine/DotnetCommands.cs
```diff
@@ -31,12 +30,13 @@ public static bool AddPackage(string packageName, ILogger logger, string? projec
...
 
             arguments.AddRange(["package", packageName]);
+
             if (!string.IsNullOrEmpty(packageVersion))
             {
                 arguments.AddRange(["-v", packageVersion]);
             }
-
-            if (includePrerelease)
+            // cannot specify both version and prerelease
```
drewnoakes: Should this validate the arguments and throw if this assumption/requirement is violated?

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Common/PackageConstants.cs
```diff
@@ -16,31 +19,32 @@ public static class EfConstants
...
-        public const string SqlServerPackageName = "Microsoft.EntityFrameworkCore.SqlServer";
-        public const string SqlitePackageName = "Microsoft.EntityFrameworkCore.Sqlite";
-        public const string CosmosPakcageName = "Microsoft.EntityFrameworkCore.Cosmos";
-        public const string PostgresPackageName = "Npgsql.EntityFrameworkCore.PostgreSQL";
+        public static readonly Package EfCorePackage = new("Microsoft.EntityFrameworkCore", true);
+        public static readonly Package EfCoreToolsPackage = new("Microsoft.EntityFrameworkCore.Tools", true);
+        public static readonly Package SqlServerPackage = new("Microsoft.EntityFrameworkCore.SqlServer", true);
+        public static readonly Package SqlitePackage = new("Microsoft.EntityFrameworkCore.Sqlite", true);
+        public static readonly Package CosmosPackage = new("Microsoft.EntityFrameworkCore.Cosmos", true);
+        public static readonly Package PostgresPackage = new("Npgsql.EntityFrameworkCore.PostgreSQL", true);
```
drewnoakes: nit: I'd find this a tad clearer if the `true` was named to explain what it means.

e.g.

```suggestion
        public static readonly Package EfCorePackage = new("Microsoft.EntityFrameworkCore", IsVersionRequired: true);
        public static readonly Package EfCoreToolsPackage = new("Microsoft.EntityFrameworkCore.Tools", IsVersionRequired: true);
        public static readonly Package SqlServerPackage = new("Microsoft.EntityFrameworkCore.SqlServer", IsVersionRequired: true);
        public static readonly Package SqlitePackage = new("Microsoft.EntityFrameworkCore.Sqlite", IsVersionRequired: true);
        public static readonly Package CosmosPackage = new("Microsoft.EntityFrameworkCore.Cosmos", IsVersionRequired: true);
        public static readonly Package PostgresPackage = new("Npgsql.EntityFrameworkCore.PostgreSQL", IsVersionRequired: true);
```

I try to do that whenever an argument is a literal value (like `true`, `null`, `0`) and the meaning isn't immediately obvious from context (e.g. `names.Take(3)` and `names.Add(null)` are pretty obvious).

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Scaffolders/ScaffolderContextExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+public static class ScaffolderContextExtensions
+{
+    /// <summary>
+    /// Gets the specified target framework from the scaffolder context.
+    /// </summary>
+    /// <param name="context">The scaffolder context.</param>
+    /// <returns>The target framework name if present; otherwise, null.</returns>
+    public static string? GetSpecifiedTargetFramework(this ScaffolderContext context)
+    {
+        if (context.GetOptionResult<string>("--target-framework") is string tfm)
```
drewnoakes: There's a constant for this string in `CliStrings`.

src/dotnet-scaffolding/dotnet-scaffold/ScaffoldingSteps/ValidateTargetFrameworkStep.cs
```diff
@@ -0,0 +1,104 @@
...
+        if (UnsupportedLegacyFrameworks.Contains(normalizedValue))
+        {
+            logger.LogError($"Invalid --target-framework option: '{value}'. .NET 8.0 is the lowest supported version. Please use net8.0, net9.0, or net10.0.");
+            return false;
+        }
+
+        // Check if it's a valid supported framework
+        if (!ValidTargetFrameworks.Contains(normalizedValue))
+        {
+            logger.LogError($"Invalid --target-framework option: '{value}'. Must be a valid .NET SDK version (e.g., net8.0, net9.0, net10.0).");
```
drewnoakes: Ditto, use constant for `--target-framework` from `CliStrings`, so that it's only defined in one place.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Model/NuGetVersionHelper.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// Gets the latest major version 10 of a NuGet package (e.g., for .NET 10).
+        /// </summary>
+        public static async Task<NuGetVersion?> GetLatestNet10VersionAsync(string packageId)
+        {
+            IEnumerable<NuGetVersion> versions = await GetVersionsForPackageAsync(packageId);
+
+            // Filter for .NET 10 compatible versions (major version 10)
+            var net10Versions = versions.Where(v => v.Major == 10).OrderByDescending(v => v);
+            return net10Versions.FirstOrDefault();
+        }
```
drewnoakes: These methods are all pretty similar, so the logic is ripe for a refactor.

By baking the version number into the method name, it requires an API change whenever a new version comes out.

So, I wonder if we can merge these into a single method and have the caller pass a major version number.

I think this could be part of a potential change that provides the user with a list of target frameworks to select from, rather than giving them freeform text. With an approach like that:

- We wouldn't have to parse user input
- We bypass all the complexity associated with trying to interpret `TargetFramework` strings
- We provide a more user-friendly experience, by guiding them to values that we know to be valid options

What if the overall logic called `GetVersionForPackageAsync` _before_ we asked the user what target framework they wanted, and used that information to drive the UI? We may still want to perform some filtering on the `NuGetVersion` values we produce.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Model/NuGetVersionHelper.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        private static async Task<IEnumerable<NuGetVersion>> GetVersionsForPackageAsync(string packageId)
+        {
+            IEnumerable<Lazy<INuGetResourceProvider>> providers = Repository.Provider.GetCoreV3();
+            PackageSource source = new("https://api.nuget.org/v3/index.json");
```
drewnoakes: If this is running in the context of an existing project, we may need to take any `NuGet.config` into account. For example, in many of the `dotnet` org's repos, we use AzDO feeds for packages. This is quite common in corporate settings. These feeds may have different package versions than nuget.org. I'm not sure how best to do this though. @zivkan, do you have some guidance on this?

src/dotnet-scaffolding/Directory.Packages.props
```diff
@@ -26,15 +25,17 @@
     <PackageVersion Include="Microsoft.Extensions.Logging.Console" Version="$(MicrosoftExtensionsLoggingConsolePackageVersion)" />
     <PackageVersion Include="Mono.TextTemplating" Version="$(MonoTextTemplatingVersion)" />
     <PackageVersion Include="Newtonsoft.Json" Version="$(NewtonsoftJsonPackageVersion)" />
+    <PackageVersion Include="NuGet.Configuration" Version="7.0.1" />
+    <PackageVersion Include="NuGet.Protocol" Version="7.0.1" />
+    <PackageVersion Include="NuGet.Versioning" Version="7.0.1" />
```
drewnoakes: Looks like (almost) all the other packages use an MSBuild property to control the version number. Perhaps do the same for consistency.

zivkan: NuGet.Versioning, and probably NuGet.Configuration, are dependencies of NuGet.Protocol, so you could just reference protocol and get the other packages transitively.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Model/NuGetVersionHelper.cs
zivkan: NuGet.Configuration has an API, `Settings.LoadDefaults(path)` which will return a NuGet `ISettings` that represents the aggregated settings for a solution in the provided directory.  You can then use `PackageSourceProvider`'s `LoadPackageSources`. But I'm pretty sure it returns disabled sources, so be sure to filter it based on the `IsEnabled` property. You should use `CachingSourceProvider` to convert `PackageSource` instances into `SourceRepository`, because as the name suggests it caches the `SourceRepository` instances, and if a package source needs interactive login for authentication, you don't want to prompt the user multiple times. Even if the package source is unauthenticated, new instances will have to re-parse the service index (index.json) every time, but reusing the same instance it'll be cached.

`FindPackageByIdResource` is poorly named, but it's something designed for restore, not for general use. Getting the version list will return unlisted packages, and there's no way to tell which are unlisted. You should use `PackageMetadataResource` instead.

To tell NuGet to use credential providers, you need to call [`DefaultCredentialServiceUtility.SetupDefaultCredentialService`](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Credentials/DefaultCredentialServiceUtility.cs#L28) once per process (it's a static). I suspect it'll be ok if you call it more than once, but it's designed to be called exactly once.

When more than one package source is defined in the nuget.config, we don't have helper methods to fan out for all sources and merge the results into a single version list. So, you'll need to handle that yourself, including parallelism and then choosing one source to download from.

If the nuget.config has package source mapping enabled, unfortunately you'll also need to handle that yourself to avoid supply chain attacks. `PackageSourceMapping.GetPackageSourceMapping(ISettings).` can give you an instance that you can then call `IsEnabled` to see if it's enabled or not, and then `GetConfiguredPackageSources(packageName)` to get the list of configured source(s) that package is allowed to be downloaded from.

It must be pretty obvious that all these APIs were written for internal use and just made public, rather than having a public API design review process 😆

haileymck:
> wondering if it makes sense to put target framework before prerelease in the interactive flow...

maybe I only show "prerelease" option if they select net10?:?

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/CommandLine/DotnetCommands.cs
haileymck: ok! I updated so that the interactive flow cannot specify both and if the user specifies in noninteractive the `dotnet package add` will capture that error in logging

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Builder/ScaffolderOption.cs
haileymck: removed this because changed the net framework to selection picking

src/dotnet-scaffolding/Directory.Packages.props
haileymck: I put the version number in Version.props and it seems that Nuget.Versioning and Nuget.configuration are not getting picked up when directly referenced, so leaving the reference here

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Model/NuGetVersionHelper.cs
haileymck: > What if the overall logic called `GetVersionForPackageAsync` _before_ we asked the user what target framework they wanted, and used that information to drive the UI? We may still want to perform some filtering on the `NuGetVersion` values we produce.

Trying to understand what you mean here, are you saying to calculate the package versions directly before prompting for the target framework. This PR only prompts for target framework for those aspnet scaffolders that install NuGet packages (mentioned above but package version doesn't differ for aspire scaffolders). Some of these Aspnet scaffolders install a few NuGet packages. I worry that putting these packages in the UI interactive selection flow would really lengthen the number of prompts the users interact with. 

 Maybe I could surface the version added in the output messages. For example,  this identity scaffolders installs the following packages it could say something `Adding package 'Microsoft.AspNetCore3.Identity.UI" version ...` or something like that.
image

src/dotnet-scaffolding/dotnet-scaffold/ScaffoldingSteps/ValidateTargetFrameworkStep.cs
haileymck: updated

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Model/NuGetVersionHelper.cs
haileymck: thanks for the help @zivkan !! I updated the way that it is the NuGet package versions with your suggestions. I did not end up calling the `SetupDefaultCredentialService` because it didn't seem needed, but happy to add it if it makes sense here

src/dotnet-scaffolding/dotnet-scaffold/ScaffoldingSteps/ValidateTargetFrameworkStep.cs
haileymck: I originally  added these target framework strings just to provide a different error message if the users specifies one (via --framework) that isn't net8, 9 or 10. I just updated the message if they don't specify net 8 9 or 10, `Invalid --framework option: {value}. Must be a valid .NET SDK version, net8.0, net9.0 or net10.0.` I don't think I need a more customized error message when they use a target framework but not one we want.  It wasn't necessary to do this validation!

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Model/NuGetVersionHelper.cs
haileymck: I updated!

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/Package.cs
```diff
@@ -0,0 +1,87 @@
...
+    /// <exception cref="NotSupportedException">Thrown if <paramref name="targetFramework"/> is not one of the supported frameworks ("net8.0", "net9.0", or
+    /// "net10.0").</exception>
+    private static Task<NuGetVersion?> GetVersionForTargetFrameworkAsync(this Package package, string targetFramework, NuGetVersionService nugetVersionHelper)
+    {
+        if (!package.IsVersionRequired)
+        {
+            return Task.FromResult<NuGetVersion?>(null);
+        }
+
+        //if target framework is net8.0 return what GetLatestNet8VersionAsync calls for net8.0
```
phenning: nit, out of date comment?

phenning (dismissed):

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/Package.cs
haileymck: removed!

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/NuGetVersionService.cs
```diff
@@ -0,0 +1,106 @@
...
+        {
+            IReadOnlyList<string> configuredSources = sourceMapping.GetConfiguredPackageSources(packageId);
+            if (configuredSources.Any())
+            {
+                packageSources = packageSources.Where(s => configuredSources.Contains(s.Name));
+            }
+        }
+
+        // Use CachingSourceProvider to cache SourceRepository instances
+        CachingSourceProvider cachingProvider = new(sourceProvider);
```
zivkan: Since you're creating this CachingSourceProvider inside the `GetVersionsForPackageAsync(string)` method, if you check two different packages, you're not going to re-use the cache.  You need to reuse the same `CachingSourceProvider` instance each time.

Similarly, unless you expect for nuget.config to change, you don't need to load it every time you check a different package id.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Model/NuGetVersionHelper.cs
zivkan: Make sure you test with a feed that requires authentication. If you only test with public feeds then you won't need authentication.  But most MSFT internal teams, and many external enterprise customers, do not allow nuget.org to be used directly. Instead, they must use private feeds that mirror nuget.org packages.

For example, have a look at dev.azure.com/devdiv's Artifacts feeds, and try using one of those with your manual test.

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Extensions/BlazorCrudScaffolderBuilderExtensions.cs
```diff
@@ -55,11 +55,11 @@ public static IScaffoldBuilder WithBlazorCrudAddPackagesStep(this IScaffoldBuild
         {
             var step = config.Step;
             var context = config.Context;
-            var packageNames = new List<string>()
+            var packages = new List<Package>()
```
drewnoakes, line 58: Unrelated to this PR, but the `packages` list is only used if an `if` check passes. It might be good to allocate it only when it's actually needed. Could be a follow up. There are other instances of this in other extensions.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/NuGetVersionService.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// </summary>
+    /// <param name="packageId">the ID of the NuGet package</param>
+    /// <param name="MajorVersion">the .NET major version</param>
+    /// <returns></returns>
+    public async Task<NuGetVersion?> GetLatestPackageForNetVersionAsync(string packageId, int MajorVersion)
+    {
+        IEnumerable<NuGetVersion> versions = await GetVersionsForPackageAsync(packageId);
+
+        // Filter for the specified .NET major version
+        var compatibleVersions = versions.Where(v => v.Major == MajorVersion).OrderByDescending(v => v);
```
drewnoakes: Should the `==` be `>=`?

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/NuGetVersionService.cs
```diff
@@ -0,0 +1,109 @@
...
+            foreach (var metadata in metadataList)
+            {
+                if (metadata.Identity?.Version != null)
+                {
+                    allVersions.Add(metadata.Identity.Version);
+                }
+            }
+        }
+
+        return allVersions.OrderByDescending(v => v);
```
drewnoakes, line 117: @zivkan if two feeds have the same package, but at different versions, does NuGet only consider the first configured feed, even if other feeds have higher/better versions? I have the sense that feed order specifies some kind of precedence, but that's not reflected here. If that's the case, then it's possible this code might suggest a version which then cannot be resolved by the selected feed.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/Package.cs
```diff
@@ -0,0 +1,85 @@
...
+
+/// <summary>
+/// Represents a NuGet package with a specified name and optional version.
+/// </summary>
+internal sealed record Package(string Name, bool IsVersionRequired = false)
+{
+    /// <summary>
+    /// Gets or sets the package version.
+    /// </summary>
+    public string? PackageVersion { get; set; } = null;
```
drewnoakes: Can this be an `init` property?


```suggestion
    public string? PackageVersion { get; init; }
```

phenning (dismissed):

src/dotnet-scaffolding/dotnet-scaffold/ScaffoldingSteps/ValidateTargetFrameworkStep.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// Gets or sets the target .NET framework for the project or component.
+    /// </summary>
+    /// <remarks>Specify the framework using a valid framework moniker, such as "net6.0" or
+    /// "netstandard2.1". This property is typically used to determine compatibility and runtime behavior.</remarks>
+    public string? TargetFramework { get; set; }
+
+    private static readonly HashSet<string> ValidTargetFrameworks =
+    [
+        "net8.0", "net9.0", "net10.0"
+    ];
```
drewnoakes: Seems like a duplicate of `TargetFrameworkConstants.SupportedTargetFrameworks`, although a slightly different collection type.

Would be good to have this data in one place, as it's easier to keep consistent over time.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/Package.cs
```diff
@@ -0,0 +1,85 @@
...
+            return nugetVersionHelper.GetLatestPackageForNetVersionAsync(package.Name, 8);
+        }
+        else if (targetFramework.Equals(TargetFrameworkConstants.Net9, StringComparison.OrdinalIgnoreCase))
+        {
+            return nugetVersionHelper.GetLatestPackageForNetVersionAsync(package.Name, 9);
+        }
+        else if (targetFramework.Equals(TargetFrameworkConstants.Net10, StringComparison.OrdinalIgnoreCase))
+        {
+            return nugetVersionHelper.GetLatestPackageForNetVersionAsync(package.Name, 10);
+        }
```
drewnoakes, line 79: What if the user targets .NET 10 and there's a .NET 9 version of the package? I would think that should work, but it would fail these checks and the user would see an error.

My preference would be to use NuGet APIs to determine whether a given package is compatible with the user's selected target framework. @zivkan do you know if such an API exists?

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/TargetFrameworkConstants.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public const string TargetFrameworkCliOption = "--framework";
+    public const string TargetFrameworkDisplayName = "Target Framework";
+    public const string TargetFrameworkDescription = "Specifies the target framework for the scaffolded project.";
+
+    public const string Net8 = "net8.0";
+    public const string Net9 = "net9.0";
+    public const string Net10 = "net10.0";
+
+    public static readonly List<string> SupportedTargetFrameworks = [ Net8, Net9, Net10];
```
drewnoakes: Consider making this immutable.


```suggestion
    public static readonly ImmutableArray<string> SupportedTargetFrameworks = [Net8, Net9, Net10];
```

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/TargetFrameworkConstants.cs
```diff
@@ -0,0 +1,17 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.DotNet.Scaffolding.Core.Model;
+
+internal class TargetFrameworkConstants
```
drewnoakes: Can be a `static class`.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/NuGetVersionService.cs
haileymck: Ok updated to reuse more things here

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/TargetFrameworkConstants.cs
haileymck: updated!

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/TargetFrameworkConstants.cs
haileymck: updated!

src/dotnet-scaffolding/dotnet-scaffold/ScaffoldingSteps/ValidateTargetFrameworkStep.cs
haileymck: updated!

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/Package.cs
haileymck: yeah good point. This is relying on the user to specify the correct target framework. It could be good to get it directly instead of prompting for it

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Extensions/BlazorCrudScaffolderBuilderExtensions.cs
haileymck: I will create another work item to do this in future

haileymck: tracked in #3398

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/NuGetVersionService.cs
zivkan: No, Nuget doesn't have any source precedence. It asks all sources (using package source mapping when configured), and then downloads the package from a source with the requested version. I only noticed code in this PR selecting package versions, but not downloading nupkgs. If this repo downloads nupkgs, then you need to implement that yourself. But if you write to a project file and then do a restore, the code I've seen for selecting the version looks good to me.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Model/NuGetVersionHelper.cs
haileymck: ok added and tested with a private feed :)

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/Package.cs
zivkan: It's rather complicated, unfortunately.

We have a class `FrameworkReducer`, where you give it a TFM and a list of TFMs, and it will tell you which one from the list is "most closely compatible" with the first TFM. Normally the first TFM is a project's "inner build" TFM, and the list of TFMs is what a package supports. It returns null if none from the list are compatible with the first.

But if the project multi-targets, then you need to call this API once for each project TFM, and then you need to make a decision what happens when a package is compatible with some, but not all, project TFMs.

The next problem is you need to know what TFMs the package supports.  NuGet's server-client protocol doesn't have this info(1), so you have to download the nupkg to figure it out. I'm not entirely sure how to get it with APIs. We have a class `ManagedCodeConventions` (in NuGet.Packaging, which I think is a dependency of NuGet.Protocol), but I only know about APIs to check one asset type (compile, runtime, build, buildTransitive, etc).  You'll need enough knolwedge about NuGet to know how NuGet determines package compatibility. For example, if the package contains ONLY ContentFiles, but none of the ContentFiles are for a TFM that's compatible with the project, NuGet considers the package compatible for restore anyway.

And finally, if you're trying to go through a list of package versions to find the highest version that is compatible with a project, then you'll need to loop over all the versions until you find one that is, downloading each nupkg in turn.

(1) The client-server protocol does list the package dependencies, but restore doesn't use package dependencies for compatibility checks. While most packages have the same compat and dependency group TFM lists, nothing enforces that (there is a pack warning as of 2019), so there are plenty of packages that don't.


But it's not clear to me if this Scaffolding repo only cares about packages created from the dotnet/dotnet VMR.  If so, then you might be able to assume that the package dependencies are the same set of TFMs as the package compatbility, and use it as a cheat that can't be used in the general case.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/NuGetVersionService.cs
drewnoakes: Thanks for confirming. My understanding of the code here is that it's just tyring to find the highest version with a given major version number, from any of the configured feeds. From your explanation I think that'll work correctly at restore time.

haileymck: yeah it is relying on "dotnet package add..." command and just installing the specified version, not downloading the packages directly.

haileymck:
fixes #3418

PR #3389, opened 2025-12-03
Color console output based on LogLevel

phenning:
Uses the same approach as .NET's built-in console formatters by writing ANSI escape codes to the TextWriter

Screenshot 2025-12-03 162921


haileymck:
I think that some Errors are logged directly too, so they won't always be red with this change

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/Logging/CleanConsoleFormatter.cs
```diff
@@ -31,10 +35,33 @@ public override void Write<TState>(in LogEntry<TState> logEntry, IExternalScopeP
...
+            {
+                LogLevel.Error or LogLevel.Critical => BrightRedColor,
+                LogLevel.Warning => BrightYellowColor,
+                _ => null
+            }
+            : null;
+
+        if (colorCode is not null)
+        {
+            textWriter.Write(colorCode);
```
haileymck, line 57: I think it makes sense to maintain consistency with AnsiConsole

haileymck (approved):

Issue #3293, opened 2025-10-31
Issues using the Entra ID authentication scaffolder


danroth27:
If I try to use the Entra ID authentication scaffolder when I'm logged in using my Microsoft organization account, the `dotnet scaffold` tool will take a long time during the "Discovering scaffolders" step, about 5min. The discovery step eventually seems to time out and then only the Aspire scaffolder is displayed. None of the other ASP.NET Core scaffolders including the Entra ID scaffolder show up. I believe this is a manifestation of https://github.com/dotnet/Scaffolding/issues/3279.

<img width="434" height="359" alt="Image" src="https://github.com/user-attachments/assets/3804185a-38cc-4356-ab59-1e392a512296" />

If I instead login with my personal Azure subscription, it seems to work around this problem and the ASP.NET Core & Entra ID scaffolders do show up.

But now if I select the Entra ID scaffolder, I'm asked to make from a very long list of usernames that all appear to be the same:

<img width="1132" height="628" alt="Image" src="https://github.com/user-attachments/assets/01771b27-29f4-4934-be41-41ca28d9f5bb" />

But none of these users are from my personal Azure subscription, so I'm not sure why they show up here.

If I just pick the first username and select that I want to create a new app registration I get asked to pick from a similarly long list of indistinguishable tenants. If I just pick the first tenant, I get stuck waiting on this Create or Select Application screen:

<img width="1123" height="256" alt="Image" src="https://github.com/user-attachments/assets/4c2e02a0-02e7-4e32-8cdd-4733a98c7c71" />

After waiting a long time, I get this connection aborted error from the Azure CLI:

```
Error executing 'az ad app list': ERROR: The command failed with an unexpected error. Here is the traceback:
ERROR: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
Traceback (most recent call last):
  File "D:\a\_work\1\s\build_scripts\windows\artifacts\cli\Lib\site-packages\urllib3/connectionpool.py", line 787, in
urlopenting 'dotnet-scaffold-aspnet'
  File "D:\a\_work\1\s\build_scripts\windows\artifacts\cli\Lib\site-packages\urllib3/connectionpool.py", line 534, in
_make_requestdotnet-scaffold-aspnet'
  File "D:\a\_work\1\s\build_scripts\windows\artifacts\cli\Lib\site-packages\urllib3/connection.py", line 565, in
getresponse 'dotnet-scaffold-aspnet'
[...]

PR #3402, opened 2025-12-05
use az rest --method GET instead of az ad app list

haileymck:
`az ad app list` is slow. It can spin for at least 10 minutes for large subscriptions. Instead of making the user wait for `az ad app list` to finish when running the `Entra Id` scaffolder, instead invoke `az rest --method GET`. Also in the process of testing, I found that if the app names contain `[` and `]`, the UI will throw because Spectre.Console expects known colors in that format

### DIFF
--- a/src/dotnet-scaffolding/dotnet-scaffold/AspNet/Helpers/AzCliHelper.cs
+++ b/src/dotnet-scaffolding/dotnet-scaffold/AspNet/Helpers/AzCliHelper.cs
@@ -222,30 +222,28 @@ internal static class AzCliHelper
         {
             failingCommand = null;
             appIds = [];
-            var exitCode = runner.RunAzCli("ad app list --output json", out string? stdOut, out string? stdErr);
+            // Use Microsoft Graph API via az rest
+            var exitCode = runner.RunAzCli("rest --method GET --url https://graph.microsoft.com/v1.0/applications?$select=appId,displayName", out string? stdOut, out string? stdErr);
 
             if (exitCode == 0 && !string.IsNullOrEmpty(stdOut))
             {
-                // Parse the JSON output
                 using JsonDocument doc = JsonDocument.Parse(stdOut);
                 JsonElement root = doc.RootElement;
-
-                if (root.ValueKind == JsonValueKind.Array)
+                if (root.TryGetProperty("value", out JsonElement valueArray) && valueArray.ValueKind == JsonValueKind.Array)
                 {
-
-                    foreach (JsonElement app in root.EnumerateArray())
+                    foreach (JsonElement app in valueArray.EnumerateArray())
                     {
-                        if (app.TryGetProperty("appId", out JsonElement appId))
+                        string? id = app.TryGetProperty("appId", out JsonElement appId) ? appId.GetString() : null;
+                        string? displayName = app.TryGetProperty("displayName", out JsonElement name) ? name.GetString() : "Unknown App";
+                        if (!string.IsNullOrEmpty(id) && !string.IsNullOrEmpty(displayName))
                         {
-                            string? id = appId.GetString();
-                            string? displayName = app.TryGetProperty("displayName", out JsonElement name) ?
-                                                 name.GetString() : "Unknown App";
-
-                            if (!string.IsNullOrEmpty(id) && !string.IsNullOrEmpty(displayName))
+                            // the tool will throw if displayName contains []
+                            if (displayName.Contains('[') || displayName.Contains(']'))
                             {
-                                // Format as "DisplayName (AppId)" for better user experience
-                                appIds.Add($"{displayName} {id}");
+                                displayName = displayName.Replace("[", "(").Replace("]", ")");
                             }
+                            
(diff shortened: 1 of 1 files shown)


src/dotnet-scaffolding/dotnet-scaffold/AspNet/Helpers/AzCliHelper.cs
```diff
@@ -222,30 +222,28 @@ private static bool GetAzureAppIds(AzCliRunner runner, out List<string> appIds,
...
+                        string? displayName = app.TryGetProperty("displayName", out JsonElement name) ? name.GetString() : "Unknown App";
+                        if (!string.IsNullOrEmpty(id) && !string.IsNullOrEmpty(displayName))
                         {
-                            string? id = appId.GetString();
-                            string? displayName = app.TryGetProperty("displayName", out JsonElement name) ?
-                                                 name.GetString() : "Unknown App";
-
-                            if (!string.IsNullOrEmpty(id) && !string.IsNullOrEmpty(displayName))
+                            // the tool will throw if displayName contains []
+                            if (displayName.Contains("[") || displayName.Contains("]"))
```
phenning: Don’t have context here. Why is this necessary?

phenning: Oh I see, never mind.

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Helpers/AzCliHelper.cs
```diff
@@ -222,30 +222,28 @@ private static bool GetAzureAppIds(AzCliRunner runner, out List<string> appIds,
...
                             {
-                                // Format as "DisplayName (AppId)" for better user experience
-                                appIds.Add($"{displayName} {id}");
```
phenning, line 247: Did the old display name already have () around it?

phenning: Or did id?

Do we just need to strip the brackets rather than replace them?

haileymck: I think it was still a bug in the old display. I just found it now because one of my MSFT apps have [MVP] in the title

haileymck: I just selected to put () because they are valid here, but could just remove [] and not replace with anything

phenning: Yea I think that’s probably what the other behavior is. I wonder if the other command just returns the same data without anything around it.

phenning: If so we should probably just strip. Maybe we should format the string to put () around the id like the original comment claimed it was doing :)

haileymck: the other command returns the same thing, with the "[MVP]" just the same. I just never encountered the exception because I was not running it with my microsoft subscription.

haileymck: added the parens around app id back!

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Helpers/AzCliHelper.cs
```diff
@@ -222,30 +222,29 @@ private static bool GetAzureAppIds(AzCliRunner runner, out List<string> appIds,
...
+                                displayName = displayName.Replace("[", "(").Replace("]", ")");
                             }
+                            
+                            // Format as "DisplayName (AppId)" for better user experience
+                            appIds.Add($"{displayName} ({id})");
```
phenning: This breaks the split logic here:

https://github.com/dotnet/Scaffolding/blob/11bf37d78c6daedace2d2804afe71b7649d80ffa/src/dotnet-scaffolding/dotnet-scaffold/AspNet/Extensions/BlazorEntraScaffolderBuilderExtensions.cs#L90

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Helpers/AzCliHelper.cs
haileymck: There should not be parenthesis here. keeping removed

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Helpers/AzCliHelper.cs
haileymck: yeah I dont think it makes sense to have the parenthesis here. keeping them out

haileymck:
fixes #3293

PR #3410, opened 2025-12-08
[Release 9.0] Port --targetframework and Nuget packages changes to Release9.0

haileymck:
Porting changes from #3374

### DIFF
--- a/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/Package.cs
+++ b/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/Package.cs
@@ -73,10 +73,6 @@ internal static class PackageExtensions
         {
             return nugetVersionHelper.GetLatestPackageForNetVersionAsync(package.Name, 9);
         }
-        else if (targetFramework.Equals(TargetFrameworkConstants.Net10, StringComparison.OrdinalIgnoreCase))
-        {
-            return nugetVersionHelper.GetLatestPackageForNetVersionAsync(package.Name, 10);
-        }
         else
         {
             throw new NotSupportedException($"Target framework '{targetFramework}' is not supported.");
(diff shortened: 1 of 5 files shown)


src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/Package.cs
```diff
@@ -0,0 +1,85 @@
...
+            return nugetVersionHelper.GetLatestPackageForNetVersionAsync(package.Name, 8);
+        }
+        else if (targetFramework.Equals(TargetFrameworkConstants.Net9, StringComparison.OrdinalIgnoreCase))
+        {
+            return nugetVersionHelper.GetLatestPackageForNetVersionAsync(package.Name, 9);
+        }
+        else if (targetFramework.Equals(TargetFrameworkConstants.Net10, StringComparison.OrdinalIgnoreCase))
+        {
+            return nugetVersionHelper.GetLatestPackageForNetVersionAsync(package.Name, 10);
+        }
```
phenning: Do we want dotnet-scaffold 9 to be able to specify net10 here?

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/TargetFrameworkConstants.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public const string TargetFrameworkCliOption = "--framework";
+    public const string TargetFrameworkDisplayName = "Target Framework";
+    public const string TargetFrameworkDescription = "Specifies the target framework for the scaffolded project.";
+
+    public const string Net8 = "net8.0";
+    public const string Net9 = "net9.0";
+    public const string Net10 = "net10.0";
+
+    public static readonly ImmutableArray<string> SupportedTargetFrameworks = [ Net8, Net9, Net10];
```
phenning: Maybe it shouldn't be supported.

Just removing it here might be good enough - i don't think we need to remove it in the prerelease logic or anything.  Leaving the const above makes it so we only need to do the change here and in the prior if else logic

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/Package.cs
haileymck: I can remove all references to net10 if that makes sense. Also it might make sense to remove the prerelease option as well. what do you think?

haileymck:
fixes #3418

PR #3433, opened 2025-12-10
Get project target framework from msbuild

haileymck:
first commit:
- removes the target framework option
- calls `dotnet msbuild` to get targetframework(s)
- returns the lowest target framework (here net 10 is higher than net 8)

second commit:
- makes it so that the "prerelease" option only shows if net 10 is the target framework

third coimmit:
- logs instead of throw when unsupported framework is in the project

when project has target framework not supported (supports net 8, 9, 10):
image

### DIFF
--- a/All.sln
+++ b/All.sln
@@ -2,7 +2,7 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 18
 VisualStudioVersion = 18.0.11022.115
-MinimumVisualStudioVersion = 15.0.26730.03
+MinimumVisualStudioVersion = 15.0.26730.3
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{A4057ACF-27F0-4724-963B-44548B6BC4E9}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "test", "test", "{81B64EBF-613D-43AE-82DA-B375FB751921}"
@@ -128,6 +128,8 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "dotnet-scaffolding", "dotnet-scaffolding", "{0E27E3F4-C93F-4DF1-8F4D-1E62A6641C75}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.DotNet.Scaffolding.Roslyn.Tests", "test\dotnet-scaffolding\Microsoft.DotNet.Scaffolding.Roslyn.Tests\Microsoft.DotNet.Scaffolding.Roslyn.Tests.csproj", "{83AFCAC3-4AED-49A4-A7C1-61EC051562E4}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "dotnet-scaffold.Tests", "test\dotnet-scaffolding\dotnet-scaffold.Tests\dotnet-scaffold.Tests.csproj", "{992AB1BF-23A3-40DB-A3A5-06C80C760973}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
@@ -808,6 +810,24 @@ Global
 		{83AFCAC3-4AED-49A4-A7C1-61EC051562E4}.Release|x64.Build.0 = Release|Any CPU
 		{83AFCAC3-4AED-49A4-A7C1-61EC051562E4}.Release|x86.ActiveCfg = Release|Any CPU
 		{83AFCAC3-4AED-49A4-A7C1-61EC051562E4}.Release|x86.Build.0 = Release|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.debug_x86|Any CPU.ActiveCfg = Debug|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.debug_x86|Any CPU.Build.0 = Debug|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.debug_x86|x64.ActiveCfg = Debug|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.debug_x86|x64.Build.0 = Debug|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.debug_x86|x86.ActiveCfg = Debug|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.debug_x86|x86.Build.0 = Debug|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.Debug|x64.Build.0 = Debug|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.Debug|x86.ActiveCfg = Debug|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.Debug|x86.Build.0 = Debug|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.Release|Any CPU.Build.0 = Release|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.Release|x64.ActiveCfg = Release|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.Release|x64.Build.0 = Release|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.Release|x86.ActiveCfg = Release|Any CPU
+		{992AB1BF-23A3-40DB-A3A5-06C80C760973}.Release|x86.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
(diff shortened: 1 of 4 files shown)


src/dotnet-scaffolding/dotnet-scaffold/AspNet/Common/ProjectInfo.cs
```diff
@@ -24,9 +32,89 @@ internal class ProjectInfo
...
+                return msbuildOutput.Properties.TargetFramework;
+            }
+
+            // If multiple TargetFrameworks are set, find the lowest version
+            if (!string.IsNullOrEmpty(msbuildOutput.Properties.TargetFrameworks))
+            {
+                var frameworks = msbuildOutput.Properties.TargetFrameworks
+                    .Split(';')
+                    .Where(x => x.StartsWith("net") || x.StartsWith("netstandard"))
+                    .OrderBy(ParseFrameworkVersion)
```
drewnoakes: If the parse fails, it returns 0.0, which will come before any valid version. Is that intentional?

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Common/ProjectInfo.cs
```diff
@@ -24,9 +32,89 @@ internal class ProjectInfo
...
+            {
+                return msbuildOutput.Properties.TargetFramework;
+            }
+
+            // If multiple TargetFrameworks are set, find the lowest version
+            if (!string.IsNullOrEmpty(msbuildOutput.Properties.TargetFrameworks))
+            {
+                var frameworks = msbuildOutput.Properties.TargetFrameworks
+                    .Split(';')
+                    .Where(x => x.StartsWith("net") || x.StartsWith("netstandard"))
```
drewnoakes: Parsing the `TargetFramework` is not robust. For example, these are all valid values:

- `net9.0`
- `net90` (would be considered version 90.0)
- `net9000` (would be considered version 9000.0)
- `net9.0.0.0` (does the version class here support four parts)
- `NET9.0` (would be filtered out by case-sensitive checks earlier)

In this case, it only impacts projects that multi-target.

To do this correctly involves more steps. We need to take each of the values from `TargetFrameworks`, then call MSBuild again for each. The properties that should be used here are `TargetFrameworkIdentifier` and `TargetFrameworkVersion`. These are the "real" values that define what framework the project uses. The `TargetFramework` is just an alias and can be anything.

For example, given project:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFrameworks>net9.0;netstandard2.1</TargetFrameworks>
  </PropertyGroup>

</Project>
```

The following command will give you a reliable version:

```bash
$ dotnet msbuild -p:TargetFramework="netstandard2.1" -getProperty:"TargetFrameworkIdentifier;TargetFrameworkVersion" MyProject.csproj
{
  "Properties": {
    "TargetFrameworkIdentifier": ".NETStandard",
    "TargetFrameworkVersion": "v2.1"
  }
}

$ dotnet msbuild -p:TargetFramework="net9.0" -getProperty:"TargetFrameworkIdentifier;TargetFrameworkVersion" MyProject.csproj
{
  "Properties": {
    "TargetFrameworkIdentifier": ".NETCoreApp",
    "TargetFrameworkVersion": "v9.0"
  }
}

```

So, I'd recommend keeping the first call as:

```
$ dotnet msbuild -getProperty:"TargetFramework;TargetFrameworks" MyProject.csproj
{
  "Properties": {
    "TargetFramework": "",
    "TargetFrameworks": "net9.0;netstandard2.1"
  }
}
```

Then you have the following decisions:

- If `TargetFramework` has a value, the project only has one target and you can return it directly.
- If `TargetFrameworks ` has a single value (which can happen) you can return it directly.
- Otherwise, split `TargetFrameworks` and for each, invoke the command passing `-p:TargetFramework="<value>"` and `-getProperty:"TargetFrameworkIdentifier;TargetFrameworkVersion"`. This will give you pairs of identifier/version values to process. If you want to keep the same logic, you can filter out anything that's not `.NETStandard` or `.NETCoreApp`, then order by the parsed versions (removing the leading `v`).

drewnoakes:
In the case that the version is not supported, I think it'd be nice to not show the user a stack trace. Stack traces are usually shown when something internal goes wrong. In this case, the problem is the user input data (their project), and the tooling should handle all kinds of invalid input in as graceful and helpful way as possible.

haileymck:
> In the case that the version is not supported, I think it'd be nice to not show the user a stack trace. Stack traces are usually shown when something internal goes wrong. In this case, the problem is the user input data (their project), and the tooling should handle all kinds of invalid input in as graceful and helpful way as possible.

Ok! I will update to a log and exit instead of a throw here.

haileymck:
updated so log instead of throw when there is a unsupported target framework in the project. See the description for the updated image of what this will look like

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/Package.cs
```diff
@@ -79,7 +80,8 @@ public static async Task<Package> WithResolvedVersionAsync(this Package package,
         }
         else
         {
-            throw new NotSupportedException($"Target framework '{targetFramework}' is not supported.");
+            logger?.LogError("Target framework '{TargetFramework}' is not supported. Installing recent release of '{PackageName}'. Consider upgrading your target framework to install a compatible package version.", targetFramework, package.Name);
```
drewnoakes: Apologies if I've asked this before, but does this need to be localized?

What does "Installing recent release of ..." mean in this context? I'm not 100% clear on this error message.

haileymck: Here if the target framework is not net 8, net 9 or net 10, it just runs `dotnet add package [packageName]` without the package version specified, so it installs the latest public non pre-release version. Let me know if you have suggestions to make the error message more clear here. Or alternatively, it could not install the package if there is a non supported target framework

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Common/ProjectInfo.cs
haileymck: Thank you for writing this!! it is super helpful. I updated the code to reflect this. Let me know if it looks ok!

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Common/ProjectInfo.cs
haileymck: removed!

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/Package.cs
haileymck: Updated to "Installing latest stable version of '{PackageName}'". hopefully that is clearer.

src/dotnet-scaffolding/dotnet-scaffold/Interactive/Flow/Steps/ParameterBasedFlowStep.cs
```diff
@@ -165,5 +168,33 @@ private void SelectCodeService(IFlowContext context, string projectPath)
...
+        {
+            //TODO update with each major release of .NET
+
+            string projectParameterKey = Parameter.GetParameterName(Constants.CliOptions.ProjectCliOption, AspnetStrings.Options.Project.DisplayName);
+
+            if (context.Properties.Get(projectParameterKey) is FlowProperty projectFileProperty &&
+                projectFileProperty.Value is string projectFilePath && !string.IsNullOrEmpty(projectFilePath))
+            {
+                string? targetFramework = ProjectInfo.GetLowestTargetFrameworkFromCli(projectFilePath);
+                if (!string.IsNullOrEmpty(targetFramework) && !targetFramework.Equals(TargetFrameworkConstants.Net10, StringComparison.OrdinalIgnoreCase))
```
phenning: I'm wondering if this should be Net11 now that net10 is GA'ed.   Also, lets consider adding a TargetFrameworkConstants.LatestPreview which points to the Net11 const, so that we can just update in one place in case we have multiple checks.  I'm thinking then when we ship net11, we proactively update this to net12 on that release.

haileymck: is Net 11 in preview?

phenning: No, but technically we have no previews right at this moment.  What does this option even mean anymore if we are auto detecting. For net10, net9 detected, there are no prerelease version anymore.  

When this ships and net11 goes into preview, you’d always want to be preview no?   

Thinking out loud, what should the behavior be on the 10.x version of the tool when the first net 11 sdk ships.  Should older versions ever support versions newer than the running version of the tool?

Should we just always allow prerelease packages? What’s the downside?

haileymck: I will look at what the `dotnet package add --prerelease` does right now because we should just copy what it does

haileymck: right now all the prereleases seem to be net 10. I think we should leave as net 10 for right now. When net 11 becomes available, I can update it then

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Common/ProjectInfo.cs
```diff
@@ -24,9 +32,118 @@ internal class ProjectInfo
...
+
+        static string? GetLowestFromFrameworks(string[] frameworks, string projectPath)
+        {
+            List<(string tfm, Version version)> targetFrameworks = [];
+            foreach (string tfm in frameworks)
+            {
+                try
+                {
+                    MsBuildFrameworkOutput? frameworkOutput = MsBuildCliRunner.RunMSBuildCommandAndDeserialize<MsBuildFrameworkOutput>([$"-p:TargetFramework=\"{tfm}\"",
+                        "-getProperty:TargetFrameworkIdentifier;TargetFrameworkVersion"],
```
drewnoakes: This isn't currently using `TargetFrameworkIdentifier`. The idea behind using it is to protect against frameworks that aren't supported here. Do you have to support `.NETFramework` or `.NETStandard`, or is it enough to limit this to `.NETCoreApp` (i.e. projects targeting `net5.0` and later)?

haileymck: this is limited to .net core , net 8, net 9, net 10. Not sure I need the `FrameworkIdentifier ` here because I think the version is sufficient in this case. Will remove `FrameworkIdentifier` as a property here!

haileymck: updated to not use `FrameworkIdentifier` !

drewnoakes: The version-only check works today only as an accident of history, that the version numbers of .NET Framework and .NET Standard are all lower. But by not checking the TFI then we're comparing apples and oranges.

For example, a project with `TargetFramework=monoandroid13.0` has `TargetFrameworkVersion=v13.0`. That version is high enough to be considered here, but it's not a compatible framework. The `TargetFrameworkIdentifier=MonoAndroid`, and checking that lets you see that it's unsupported.

haileymck: I updated the logic a bit! now it returns null if there is any incompatible framework and if they are all compatible, returns the lowest one. also added some unit tests :)

PR #3543, opened 2026-02-06
add net 11 end to end tests

### LINKED ISSUE #3536: Create End to End tests of running dotnet scaffold on a net 11 project

haileymck:
fixes #3536 beginning of end to end tests for `dotnet scaffold` for net 11 scenarios.

### DIFF
--- a/test/Scaffolding/E2E_Test/Net11ScaffoldingE2ETests.cs
+++ b/test/Scaffolding/E2E_Test/Net11ScaffoldingE2ETests.cs
@@ -1,0 +1,446 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.IO;
+using System.Linq;
+using Microsoft.Extensions.Internal;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.VisualStudio.Web.CodeGeneration.E2E_Test
+{
+    /// <summary>
+    /// End-to-end tests for scaffolding operations on .NET 11 projects.
+    /// These tests verify that scaffolding commands work correctly with the net11.0 target framework.
+    /// Uses the new dotnet scaffold tool instead of the legacy aspnet-codegenerator.
+    /// 
+    /// NOTE: Many of these tests are currently skipped as the dotnet scaffold tool and underlying
+    /// dotnet new templates may not yet create files in the expected locations or with the expected
+    /// naming conventions. These tests will need to be updated once the tool behavior is finalized.
+    /// </summary>
+    [Collection("E2E_Tests")]
+    public class Net11ScaffoldingE2ETests : E2ETestBase
+    {
+        public Net11ScaffoldingE2ETests(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        /// <summary>
+        /// Override Scaffold method to use dotnet scaffold instead of aspnet-codegenerator for .NET 11 tests
+        /// </summary>
+        protected new void Scaffold(string[] args, string testProjectPath)
+        {
+            var muxerPath = DotNetMuxer.MuxerPathOrDefault();
+
+            var invocationArgs = new[]
+            {
+                "scaffold"
+            }.Concat(args);
+
+            Output.WriteLine($"Executing {muxerPath} {string.Join(" ", invocationArgs)}");
+
+            var exitCode = Command.Create(muxerPath, invocationArgs)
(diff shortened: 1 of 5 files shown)


test/Scaffolding/E2E_Test/Net11ScaffoldingE2ETests.cs
```diff
@@ -0,0 +1,446 @@
...
+                    "--name",
+                    "ProductsController"
+                };
+
+                Scaffold(args, TestProjectPath);
+
+                var controllerFile = Path.Combine(TestProjectPath, "ProductsController.cs");
+                Assert.True(File.Exists(controllerFile), $"Controller file not found: {controllerFile}");
+                
+                var controllerContent = File.ReadAllText(controllerFile);
```
drewnoakes, line 80: I wonder if using async IO here would help the test suite to complete more quickly, as the test runner thread wouldn't be blocked here, and could do other work. I don't know enough about xUnit and collections to know if it'd be a factor here, but something to consider and should be an easy change.

Ditto other tests in this class.

test/Scaffolding/E2E_Test/Net11ScaffoldingE2ETests.cs
```diff
@@ -0,0 +1,446 @@
...
+        }
+
+        #region Controller Scaffolding Tests
+
+        [SkippableFact]
+        public void TestNet11_MvcController_EmptyScaffold()
+        {
+            string runSkippableTests = Environment.GetEnvironmentVariable("SCAFFOLDING_RunSkippableTests");
+            Skip.If(string.IsNullOrEmpty(runSkippableTests));
+            Skip.If(true, "dotnet scaffold aspnet mvccontroller does not yet create files in expected locations. Tool uses dotnet new which creates files in project root with template naming.");
```
drewnoakes, line 58: These first three tests don't do anything, as they always skip themselves. Is that intended?

drewnoakes: Actually, eight of the tests here are always skipped.

test/Scaffolding/E2E_Test/Net11ScaffoldingE2ETests.cs
```diff
@@ -0,0 +1,446 @@
...
+                .ExitCode;
+        }
+
+        #region Controller Scaffolding Tests
+
+        [SkippableFact]
+        public void TestNet11_MvcController_EmptyScaffold()
+        {
+            string runSkippableTests = Environment.GetEnvironmentVariable("SCAFFOLDING_RunSkippableTests");
+            Skip.If(string.IsNullOrEmpty(runSkippableTests));
```
drewnoakes, line 57: Can you explain this logic? These tests will only run if called from `run-net11-e2e-tests.ps1`, which sets this environment variable. What is this for? Does anything call that ps1 file, or will these tests never run during CI or IDE runs? Seems surprising.

test/Scaffolding/E2E_Test/Net11ScaffoldingE2ETests.cs
```diff
@@ -0,0 +1,446 @@
...
+
+                var controllerFile = Path.Combine(TestProjectPath, "ProductsController.cs");
+                Assert.True(File.Exists(controllerFile), $"Controller file not found: {controllerFile}");
+                
+                var controllerContent = File.ReadAllText(controllerFile);
+                Assert.False(string.IsNullOrEmpty(controllerContent), "Controller file is empty");
+                
+                // Verify basic controller structure for .NET 11
+                Assert.Contains("class ProductsController", controllerContent);
+                Assert.Contains("Controller", controllerContent); // Should inherit from Controller base class
```
drewnoakes, line 85: Is there a reason we don't test the full content of the output?

On the one hand, checking the full content would find more regressions, but on the other hand it would make the tests more susceptible to breaks. Interested to understand the philosophy here.

test/Scaffolding/Shared/MSBuildProjectStrings.cs
```diff
@@ -1269,5 +1269,266 @@ public class Blog
   </Target>
 </Project>
 ";
+
+        #region .NET 11 Project Templates
```
drewnoakes, line 1273: nit: consider raw strings (i.e. `"""`) for these multiline source code strings. They're able to be indented in a way that doesn't break the indentation of the containing type's members.

e.g.

```c#
        public const string Net11ProjectTxt = """
            <Project ...>
            """;
```

PR #3567, opened 2026-02-14
Do not surface aspire scaffolding for net 8 projects

### LINKED ISSUE #3566: Do not display aspire scaffolding for net 8

haileymck:
fixes #3566

### DIFF
--- a/test/dotnet-scaffolding/dotnet-scaffold.Tests/Interactive/Flow/FlowContextExtensionsTests.cs
+++ b/test/dotnet-scaffolding/dotnet-scaffold.Tests/Interactive/Flow/FlowContextExtensionsTests.cs
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections.Generic;
+using System.Linq;
+using Microsoft.DotNet.Scaffolding.Core.ComponentModel;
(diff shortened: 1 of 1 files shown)


src/dotnet-scaffolding/dotnet-scaffold/Interactive/Flow/FlowExtensions.cs
```diff
@@ -102,6 +102,19 @@ internal static class FlowContextExtensions
             return context.GetValueOrThrow<IRemainingArguments>(FlowContextProperties.RemainingArgs, throwIfEmpty);
         }
 
+        /// <summary>
+        /// Gets whether Aspire scaffolders should be available (based on detected project TFM).
+        /// Returns true if Aspire is available or if the property hasn't been set.
+        /// </summary>
+        public static bool GetIsAspireAvailable(this IFlowContext context)
```
drewnoakes: super nit, but `GetIs*` would be more commonly just `Is*`. Getters for bools often start with `Is` rather than `Get` in .NET APIs.

haileymck: updated the logic here so the method isn't here anymore

LittleLittleCloud (approved):

PR #3731, opened 2026-04-23
Resolve ArgumentIsEmpty Error after ef migrations

haileymck:
fixes [Azdo bug 2960694](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2960694)
The issue that sometimes users can have a version of the Entity Framework packages specified in their csproj file, but when adding a new entity framework version we would just add the most recent non prerelease version of the package. Now, match the package version to that existing in the csproj file.

### DIFF
--- a/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/Package.cs
+++ b/src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/Package.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.Extensions.Logging;
 using NuGet.Versioning;
+using System.Xml.Linq;
 
 namespace Microsoft.DotNet.Scaffolding.Core.Model;
 
@@ -19,6 +20,18 @@ internal sealed record Package(string Name, bool IsVersionRequired = false)
 
 internal static class PackageExtensions
 {
+    private static readonly StringComparer PackageNameComparer = StringComparer.OrdinalIgnoreCase;
+    private static readonly string[] AspNetEfVersionAnchorPackages =
+    [
+        PackageNameConstants.AspNetCorePackages.IdentityEntityFrameworkCorePackageName,
+        PackageNameConstants.AspNetCorePackages.DiagnosticsEntityFrameworkCorePackageName
+    ];
+
+    public static Task<Package> WithResolvedVersionAsync(this Package package, TargetFramework? targetFramework, NuGetVersionService nugetVersionHelper, ILogger? logger)
+    {
+        return package.WithResolvedVersionAsync(targetFramework, nugetVersionHelper, projectPath: null, logger: logger);
+    }
+
     /// <summary>
     /// Returns a copy of the specified package with its version property resolved for the given target framework, if
     /// the version is not already set.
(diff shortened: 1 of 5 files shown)


src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/Package.cs
```diff
@@ -72,6 +87,11 @@ public static async Task<Package> WithResolvedVersionAsync(this Package package,
             return Task.FromResult<NuGetVersion?>(null);
         }
 
+        if (targetFramework is TargetFramework.Net10 && TryResolveAspNetAlignedEfVersion(package.Name, projectPath, out NuGetVersion? alignedVersion))
```
emaf: Should we do this regardless of the TFM version? In the description you mentioned the issues is users could already have an EF version set in the csproj, but that could happen for any TFM version, right?

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Common/PackageConstants.cs
```diff
@@ -65,9 +64,9 @@ public static class EfConstants
     public static class AspNetCorePackages
     {
         public static readonly Package QuickGridEfAdapterPackage = new("Microsoft.AspNetCore.Components.QuickGrid.EntityFrameworkAdapter", IsVersionRequired: true);
-        public static readonly Package AspNetCoreDiagnosticsEfCorePackage = new("Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore", IsVersionRequired: true);
+        public static readonly Package AspNetCoreDiagnosticsEfCorePackage = new(PackageNameConstants.AspNetCorePackages.DiagnosticsEntityFrameworkCorePackageName , IsVersionRequired: true);
```
emaf:
```suggestion
        public static readonly Package AspNetCoreDiagnosticsEfCorePackage = new(PackageNameConstants.AspNetCorePackages.DiagnosticsEntityFrameworkCorePackageName, IsVersionRequired: true);
```

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Templates/net10.0/EfController/MvcEfController.cs
```diff
@@ -32,7 +32,7 @@ public virtual string TransformText()
     string modelNameLowerInv = modelName.ToLowerInvariant();
     string dbContextNamespace = string.IsNullOrEmpty(Model.DbContextInfo.DbContextNamespace) ? string.Empty : Model.DbContextInfo.DbContextNamespace;
     string dbContextName = Model.DbContextInfo.DbContextClassName;
-    string entitySetName = string.IsNullOrWhiteSpace(Model.DbContextInfo.EntitySetVariableName) ? modelName : Model.DbContextInfo.EntitySetVariableName;
+    string entitySetName = Model.DbContextInfo.EntitySetVariableName ?? modelName;
```
emaf: Is this intentional? `Model.DbContextInfo.EntitySetVariableName` could be `string.empty`, and with this change `entitySetName` will be too in that case.

src/dotnet-scaffolding/Microsoft.DotNet.Scaffolding.Core/NuGet/Package.cs
haileymck: ok yes! thanks for point this out!

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Templates/net10.0/EfController/MvcEfController.cs
haileymck: nope! this got pulled in from another branch!!

src/dotnet-scaffolding/dotnet-scaffold/AspNet/Common/PackageConstants.cs
haileymck: fixed

emaf (approved):

drewnoakes (approved):

PR #3737, opened 2026-04-28
Add aspnet scaffolding testing to CI to ensure that post scaffolding build

### LINKED ISSUE #3582: Add ASP NET integration tests

haileymck:
Adds integration tests for the aspnet scaffolding. They are run through creating new projects and then running non-interactive commands on this project and ensures that the project builds after completing these tests.


test/dotnet-scaffolding/dotnet-scaffold.Tests/AspNet/ScaffoldSteps/AddClientSecretStepTests.cs
```diff
@@ -297,32 +297,4 @@ public void SecretName_CanBeCustomized()
         // Assert
         Assert.Equal(customSecretName, step.SecretName);
     }
-
-    [Fact]
```
haileymck, line 301: this test isn't valid because of the credentials needed for the az cli work

eng/pipelines/scaffold-integration-matrix.yml
```diff
@@ -0,0 +1,162 @@
...
+  - task: UseDotNet@2
+    displayName: Install .NET 8 SDK
+    inputs:
+      packageType: sdk
+      version: 8.x
+      installationPath: $(Build.SourcesDirectory)/.dotnet
+  - task: UseDotNet@2
+    displayName: Install .NET 9 SDK
+    inputs:
+      packageType: sdk
```
phenning, line 43: I’m wondering if we need the actual sdks here or if we can specify runtime instead. 

It would improve agent setup time, but if there are global jsons anywhere pointing to older sdks it wouldn’t work

Could always update later after the initial commit as well in order to improve time

phenning (approved):

eng/pipelines/scaffold-integration-matrix.yml
haileymck: it was super finiky with the runtimes and this is the only way anything actually worked

PR #3801, opened 2026-08-14
Avoid invalid framework fallback for latest-version packages

haileymck:
## Summary

- prevent packages marked `UseLatestVersion` from receiving a target-framework-major fallback when NuGet metadata resolution returns no version
- allow packages with independent versioning, such as `SQLitePCLRaw.bundle_e_sqlite3`, to use the latest compatible stable version
- address recurring CI failures where net9 tests attempted to restore the nonexistent `SQLitePCLRaw.bundle_e_sqlite3` version `9.*-*`

## Validation

- built `dotnet-scaffold.Tests` targeting net11.0
- ran all 14 package-resolution tests successfully on net11.0
- ran the full 3,425-test suite locally; remaining integration failures occur before scaffolding because the machine has .NET 11 preview 5 while the repository requires preview 6


danroth27 (approved):
LGTM

